PR #47, opened 2018-06-22
Fix failing test_styles tests.

tlovett:
Thanks for contributing this library.  We're exploring using it for some work on the Open Network Automation Platform (ONAP). 

While attempting to run the test suite on my machine using tox, the first test_styles test case would pass, but then all subsequent test cases would fail.   I discovered the the css files associated with the StandaloneHTMLBuilder were not being reset between tests.  

I made the following tweak as a workaround and they are now passing.

### DIFF
--- a/tests/test_styles/test_style_blank.py
+++ b/tests/test_styles/test_style_blank.py
@@ -1,8 +1,13 @@
 import sys
+
+from sphinx.builders.html import StandaloneHTMLBuilder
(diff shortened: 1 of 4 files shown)


danwos:
Ohh great, I have struggled with this problem also in the past but wasn't able to solve it.
So big thanks for your pull request and the spent time.

PR #48, opened 2018-06-22
Allow Custom Options for Needs

tlovett:
One use case we encountered was the desire to specify our own set of metadata/options associated with a set of requirements.  We looked at using tags, but we wanted the metadata to be able to be part of columns, filters, and the JSON output as discrete fields.

This change allows these options to be specified via a new configuration option called `needs_extra_options`.

These options simply extend the default options, and can be used in tables, filters, and in the json output.

Internally the main change was to allow the `need_info` dict to be passed in its entirety to places where previously only selected items were passed.  Where needed, the items we don't want passed are explicitly filtered out rather than being specifically selected.

One other change was to change the internal key for tracking the target node from target to target_node.  I made this change because one of the custom attributes we wanted to use was target, and this created a conflict.

Once again, thanks for making this available.  We're excited to start using it, and we hope others might find this contribution useful.

Thanks,
Trevor


danwos:
Wow, that's an awesome pull request and a great idea.
Will need some time to make a deeper review, but I try to finish it on the beginning of next week.

I'm not sure if new options need to be / should be registered inside the option_spec dict of the NeedDirective class. But tests are passing, so looks like everything is okay.
Thanks also for updating docs and tests.
If my review is fine (what I think), I will also release a new version with your changes.

Thanks for diving so deep into sphinx-needs and sharing this improvement.

tlovett:
Sounds good. Thanks for taking a look.  Definitely open to any suggestions
or further tweaks.

danwos:
Merged it.
Added some extra content to docs.
Added also your name to the AUTHORS list. I hope your a fine with that.

Found a small bug, where needs can not be filtered, if custom option was not set.
Reason: the new option must be always part of need_info, even if it was not set by the user.
So all needs get now the custom options with an empty value, if not set.

tlovett:
Fantastic.  Thanks again for the review and collaborating on this.  It will be a big help to us.

PR #52, opened 2018-06-25
Adding sections to needs metadata

tlovett:
Sorry, me again :).  We encountered another use case where we wanted to create a set of appendices that had a `needtable` that included all the requirements from a specific section in the document.  This feature adds a processor to update all the needs with the section titles after the doctree is resolved.  Then the section is available in the filters and its also exported the needs.json.

Any interest in this capability?  If not, I could probably just add it via a setup method in our conf.py.

I didn't provide documentation on this one yet as I was still thinking through where it might go.  Before I went that far, I thought I'd see if you'd want to consider this as a feature at all.

If so, I don't mind taking a crack at some documentation for it, and curious if you have any thoughts on where it might make sense to document it.

I also added some tests for the needs.json creation while I was already in there validating that section was in there.

Thanks,
Trevor


danwos:
I like the idea and also the implementation looks great again (as usual).
However, could you please squash your commits to a single commit?
That makes it easier for me to follow the changes, especially if a later commit removes lines, which where introduced by a prior commit.
Maybe we could also get rid of the merge and pull requests, as they are already part of our master branch.

For documentation I would simply mention it on the filter-page:
https://sphinxcontrib-needs.readthedocs.io/en/latest/filter.html#option-filter

And also the needs-builder json example file should be updated.
That's it. So not much to document for this feature.

danwos:
Just checked the code and the functions.
Nothing to complain about.

I'm just thinking, if we could extend the feature a little bit.
Currently only the "lowest" section gets stored.
Maybe it makes sense to get also all the other above sections and store them as list.

Example:
```rst
headline
=======

section A
-------------

section B
-------------

subsection B.1
~~~~~~~~~~~
```
Result:
`section = ['headline', 'section B', 'subsection B.1' ]`

May it's a benefit to filter needs on higher section names and still be able to use subsections to structure the internal layout.

But currently I personal do not need such a feature inside my projects, so it's totally up to you to decide, if this makes sense.

tlovett:
Good idea. Let me mess with that today. I’ll also do the other items you
mentioned around docs and squashing the commits.

PR #53, opened 2018-06-26
Add Hierarchy of Section Titles to Needs Metadata

tlovett:
@danwos Can you cancel #52? I've re-worked this feature per our discussion and squashed the commits.   It was just easier to rebuild off a new branch to lose the merge commits.

This feature adds the section title hierarchy to each need as part of its internal metadata.  This can be used in two key ways.

1. The sections are now exported as part of `needs.json` as a list.  The first item is the needs current section, and the remaining items (if any) represent the chain of parent sections.

2. `sections` is now available in `:filter:` specifications.

One note on the implementation, when Sphinx creates a section title with auto-section numbering (`.. sectnum::`), then it inserts 3 non-breaking spaces between the section number and the section title.  I opted to automatically replace those with a single normal space.  I think that makes it easier to work with, but it's not *precisely* what the section title is.


danwos:
And again, Thanks! :)

Merged it and created also a new version for this and your last awesome changes on pypi: 0.2.2 https://pypi.org/project/sphinxcontrib-needs/

Issue #49, opened 2018-06-24
Optional Need Titles?


tlovett:
We have a situation where we have a large number of existing guidelines for vendors written as requirements.  These all have IDs and descriptions, but they don't have unique titles.  We're in the process of migrating to this framework, but at this time the title is a required element as its a required argument of the directive.

It would be nice to be able to configure this behavior differently.   

Some ideas I've considered...
1. Simply making title optional via something like `needs_title_optional = True`
    Then you'd just write the need like this.
    ```
     .. req::
          :id: R-12345
          This is my requirement
    ```
    I think this is pretty clean to implement, but when the ID isn't provided you'd need to hash the 
    content instead of the title if one isn't provided. 

2.  Allowing the title to be treated as the ID.   Something like `needs_id_as_title = True`
    ```
     .. req:: R-12345
          This is my requirement
    ```
    I like the way this looks (at least for our use case).  However, similarly to above if 
   `needs_id_required = False`, then you have to hash the content.

Any other thoughts on this is a use case that could be supported or other ideas on how to handle it?  I'm more than happy to take a crack on the implementation if so...

danwos:
Not having a title will be hard to handle internally, because the title is needed for the hash and it is also an import part of the needs templates.
Without a title the generated html side would look strange and maybe not work as expected (collapse function).
So the user should also change the templates, if not title is available. 
That's a huge task, which may lead to many, hard understandable problems for the user.

However, I understand your use case and see the need to find a solution for this.
So here is my proposal, let me know what you think:

1. There will be a global option "needs_title_required" and an need option as flag ":no-title:".
If one of this is set, a need does not need a title but some content!
2. A title gets automatically generated from the content, if no title is set by user.
Therefore we do not need to touch templates, hashing or other functions.
3. The title is based on the first sentence of the content.
4. We add a global option "needs_max_title_length", which is by default 0 and therefore deactivated.
5. If a title is longer as "needs_max_title_length", it gets cut after the last word that fits and some "..." are added.

Note: "needs_max_title_length" will be checked independently from "needs_title_required".
So it's a new functions, which has in general nothing to do with the "no title" problem, but is needed to not have too long, auto-generated titles from content.

What do you thing? Will this support your use case the right way?

tlovett:
I think it can.  I can think of a few things that would need to be considered, and you may have already done that.

1. We may need to consider how the rendered content would read as there would be some duplication. 

[...]

PR #55, opened 2018-06-27
Adding support for optional and auto-generating titles per #49

tlovett:
As discussed, here's a pass at implementing a solution for #49 and having optional titles.  Here are a few notes on implementation that might warrant further discussion.

- **`needs_max_title_length`:** By default, there is no max length.  I felt that kept things backwards compatible, and people can opt into this feature if required.

- **Handling `:title_from_content:` or `needs_title_from_content = True` when a title is provided**: The behavior I went for was to allow the title provided as the directive argument to take precedence.  If a title was provided and `:title_from_content:` was also present on the need, then I issued a warning to the log. However, I did not issue a warning if the setting was simply from the `conf.py` file as I could see legitimate cases where one might want to override the generated title.

- **Handling of a requirement with a blank title, no content, and generating a requirement ID:**  I opted to raise a `InvalidNeedException` as I could see no reasonable way of handling it.

- **Tests for warnings:** I didn't create tests to validate the warnings were being created.  pytest has the caplog feature for this, but I didn't see an easy way to inject that feature using `@with_app`

- The ellipsis on the elided requirements was difficult to test because the smartquotes feature converts them into a unicode character, but how this is done isn't consistent across Sphinx versions.  I opted to just disable the conversion using the `smartquotes_action` setting.

- When generating a title for a requirement, I left the title in the heading of the requirement. This leads to some duplication of content in the rendered document.  This was brought up in the thread on #49, but I don't think we decided anything.   Do you think we should suppress the title in the rendered need if it was auto-generated (we'd still use it in other things such as needtable)?  I it would introduce some extra complexity in the template, but could be done.

- If a generated title is used, I opted to hash on the title vs. the full content.  I figured that was a way to enforce that a duplicate title wasn't introduced, but that behavior could be changed.

- On an unrelated note, I noticed that lineno doesn't seem to provide the correct line in the source RST document.   I'll probably log a separte issue on that.


sphinxcontrib/needs/directives/need.py
```diff
@@ -138,11 +142,21 @@ def run(self):
...
+                raise NeedsDuplicatedId("A need with ID {0} already exists! "
+                                        "This is not allowed".format(id))
+            else:  # this is a generated ID
+                raise NeedsDuplicatedId(
+                    "Needs could not generate a unique ID for a need with "
+                    "the title '{}' because another need had the same title. "
+                    "Either supply IDs for the requirements or ensure the "
+                    "titles are different.  NOTE: If title is being generated "
+                    "from the content, then ensure the first sentence of the "
+                    "requirements are different.".format(' '.join(self.content)))
```
danwos: ...format(' '.join(self.content))) -->
Shouldn't this be self.trimmed_title?
Content could be quite long...

sphinxcontrib/needs/directives/need.py
```diff
@@ -31,18 +37,23 @@ class NeedDirective(Directive):
                    'hide_tags': directives.flag,
                    'hide_status': directives.flag,
                    'hide_links': directives.flag,
+                   'title_from_content': directives.flag,
```
danwos: Maybe it shouldn't be a flag.
use_case: needs_title_from_content set in conf.py.
Now there shall be a need with no title but content.
For this need we would need some kind of a boolean option: :title_from_content: false (like the collapse option).
But I'm not sure, if this is  a valid use case...

sphinxcontrib/needs/directives/need.py
tlovett: It’s wrong, but I think it should be self.full_title as that’s what gets hashed. If we just show trimmed, then there’s a small chanced it might be ambiguous which requirements caused the collision.

tlovett: It’s wrong, but I think it should be self.full_title as that’s what gets hashed. If we just show trimmed, then there’s a small chanced it might be ambiguous which requirements caused the collision.

tlovett: It’s wrong, but I think it should be self.full_title as that’s what gets hashed. If we just show trimmed, then there’s a small chanced it might be ambiguous which requirements caused the collision.

sphinxcontrib/needs/directives/need.py
tlovett: At a minimum in your use case you’d also need needs_title_optional set to True as well or this wouldn’t be allowed. 

For completeness you could do this, but I can’t think of a reason someone would selectively disable a title on an individual requirement if they were using them on all others. 

I don’t mind making the change though.

tlovett:
Also - are you OK with having the generated title show up in the rendered need?  Basically having the title and first sentence match?  It’s consistent, but some duplication for the reader

sphinxcontrib/needs/directives/need.py
danwos: You are right, this would be a realy rare case.
So no change needed here.

danwos:
Im fine with that.
And it may be necessary: if the title gets shorten, you still be able to find it in the content.

sphinxcontrib/needs/directives/need.py
tlovett: Fixed and pushed an update.

Issue #61, opened 2018-07-11
Dashes in Need IDs Not Supported by needflow


tlovett:
If a need has an ID that has a dash in it, then needflow will not be able to render the diagram.  This is because PlantUML does not support dashes in entity names.

Example:

```reStructuredText
TEST DOCUMENT
=============

.. req:: Log out (open)
    :id: A-001
    :status: open
    :tags: user;authentication

    The user should be able to log out

.. req:: Authenticate user
    :id: A-002
    :status: open
    :tags: user;login
    :links: A-001

    The user must be authenticated

.. needflow::
```
Would result in error like this:

![image](https://user-images.githubusercontent.com/6649/42581414-51cdda9a-84f2-11e8-892c-ec411cf502ac.png)

I have a fix for this, and I'll provide a Pull Request later today.  The fix is simply to transform the entity names in the diagram definition to replace dashes with underscores.   This doesn't affect how the diagram is displayed as the text used in the diagram node does not have a restriction on dashes - only the entity names in the diagram definition.

danieleades:
the test case for this is failing for me when I run it locally.

it looks like this - 
 ```python
@with_app(buildername='html', srcdir='doc_test/doc_github_issue_61')
def test_doc_github_61(app, status, warning):
    """
    Test for https://github.com/useblocks/sphinxcontrib-needs/issues/61
    """
    # PlantUML doesn't support entity names with dashes in them, and Needs uses
    # the IDs as entity names, and IDs could have dashes.  To avoid this limitation,
    # Entity names are transformed to replace the dashes with underscores in the entity
    # names.
    # Even if there's an error creating the diagram, there's no way to tell since the
    # error message is embedded in the image itself. The best we can do is make sure
    # the transformed entity names are in the alt text of the image.
    app.build()
    html = Path(app.outdir, 'index.html').read_text()
    alt_text = re.findall('<img.*?alt=(.*?)>', html, re.MULTILINE + re.DOTALL)
    assert len(alt_text) == 5
    assert "A-001" in alt_text[4]
    assert "A-002" in alt_text[4]
    assert "A_001" in alt_text[4]
    assert "A_002" in alt_text[4]
```

this line causes the issue - `assert len(alt_text) == 5`. If i debug it, I can see the `len(alt_text)` is in fact `4`.

[...]

Issue #58, opened 2018-07-05
Dynamic Values


danwos:
This feature adds the ability to ask a given function for the value of a need option.
It can be used on global_options, which are part of `conf.py`, and each option of a need, which is defined inside a rst file.

**Inside conf.py**
```python
needs_global_options = {
  'validated' : repos.is_validated,
  'validation_link': repos.get_link(),
  }
```
*Open Problem*: Not sure how to avoid python to call this function directly before sphinx-needs was able to inject current need and needlist to the call argument list (see func definition below).
Maybe we should use the same syntax as in rst files `[[ ... ]]`. Benefit: No special handling, if option is coming from global-conf.

**Inside a rst-file**
The function needs to be wrapped into ``[[ ... ]]`` to easily detect if the given string shall be treated as string or as a function call.
Inside ``[[ ... ]]`` the syntax shall be like the normal python syntax.
Arguments and keyword arguments are allowed and forwarded to the function call.
```rst
.. need:: Calculate status dynamically
   :status: [[status_by_linked_need_status("ID-1234", needed_status="Tested", new_status="Close")]]
```

**Function definition**
A function, which shall be callable from rst/global_options, must to be part of/imported to the `conf.py` file.

```python
def my_dynamic_func(need, needlist, *args, **kwargs):
    # Request arguments (see call above)
    id = args[0]
    needed_status = kwargs["needed_status"]
    new_status = kwargs["new_status"]
   
    return "always a string!"
```

@tlovett : What do you think?

tlovett:
I think the addition to global options should be handled in #57.  I think it's relatively easy to implement in conjunction in with that feature.  If the value is a `callable`, then call it using the required parameters, and if not then use the text representation of the value.

I think the functions would be more generally flexible if the environment is injected into the function as well.  For example, if you need to get access to a setting for some logic.

```
def my_function(env, need, needlist, *args, **kwargs):
   if env.config.needs_id_required:
       pass
```

danwos:
That's a good idea.
So we add `env` to the function call and do the global_options realisation in #57.

I will have some free time in 2 weeks to do some implementation work for this.
But if you need this earlier and are happy to support again, you're welcome :)

tlovett:
Sounds good.  I'm tied up in the short term as well, but I'll let you know if I free up and have time to work on it.

danwos:
Implemented and released by 0.3.0

tlovett:
Excited for this.  We'll definitely use this.

Issue #57, opened 2018-06-28
Global Options


tlovett:
Sometimes a useful attribute about a need can be derived rather than explicitly stated in the RST.

In our use case, we have a separate project that implements scripts that validate the requirement against some vendors implementation of the spec.  We'd like to keep track of which requirements have validation scripts written, but we'd also prefer not to have to keep track of this information in both the requirements project and the validation project.  The validation project already tracks traceability to the requirements in an usable data format such as YAML.   

It would be nice if we could add derived metadata to a requirement that a script exists and provide a link to that particular script.  

The idea I thought would be useful would be being able to specify dynamic attributes in the `conf.py` file, and map them to some callable that could derive the value at build time. 

Perhaps something like this in conf.py:

```Python
# Instantiate some object that can look up the information.
# In a more simple example, it could map to a single function

repos = ValidationRepository('some config to get the data')

needs_dynamic_options = {
   'validated' : repos.is_validated,
   'validation_link': repos.get_link,
}
``` 

These callables could be invoked and be passed the app environment and the need_info during the build, and then derive additional attributes.  This derivation could make use of external resources or just other attributes in need_info.

Technically a single callable could return a dict that gets merged, but I'm thinking it's better to be explicit about the values.

The downside here is this attribute only gets created at build time, but at least in our case we could have our CI builds could trigger the documentation build whenever it's updated. 

 I could also see a use case for a JavaScript method to get the value, but that would only be available to HTML builders and would probably be more complicated to implement.

tlovett:
My only concern with this approach is that I have to specify that on every need vs. just declare it once on confy.py.  In my case, I want to apply this to all needs, but my thinking was if the callable returned None then the attribute would not be added to the need.  Can you expand on why you think that would be a bad thing?

I do see the utility in being to blend it in other attributes, but I don't think that fully replaces the other use case in my opinion.

danwos:
Okay, I think this feature should be separated into 2 features:
* global_options and
* dynamic values

*Dynamic values* can be solved as mentioned above.
*Global_Options* can be defined via a variable in `conf.py` (as described above by you).

So the combination of both should fully support your case. 
But the benefit would be, that both features can be used separately and therefore supports in sum a lot of more use cases.

For *global options* I think a filter option would be great. So something like this:
 ```python
needs_global_options = [{
   'need_filter' = "'user story' == type and 'imported' in tags"
   'validated' : [[repos.is_validated]],
[...]

Issue #56, opened 2018-06-28
Dynamic Changelog


tlovett:
If someone is using a needs file for your prior version, it would be nice to be able to generate a summary of changes since the last version.

For example, a page that lists out all the needs that were 
- Added
- Deleted 
- Changed

It would also be useful if these changes were tracked as dynamic fields so they could be used in filterable directives.  For example, here's a table of new requirements.

```rst
.. needtable::
   :columns: id,title,change_type
   :filter: 'added' in change_type
```
Need to think of how this would be formatted, but it might be a new directive.  

We're actively looking at creating this on every release so we're interested in automating this.

danwos:
I'm not really sure how high the minimum effort will be to get this done or if we even should aim higher, to make sphinx-needs capable to handle historical data, which gets imported from multiple needs.json files.

Currently sphinx-needs knows nothing about old or new versions of a need. 
There is a single flat list, which contains all valid needs, which are used in the current rst docs.
Being able to get the differences of need versions means to save need-data, which is not used directly in the rst.

I see a lot of use cases, which would be realizable, if sphinx-needs can handle and provide needs based on their version (added/deleted, diff of content, highlight if status has changed, selection field on a need to dynamically load a different version, ...).

But all this means three changes/improvements for me:
**Add an additional layer to all needs**
The need-list must look something like
```python
needlist = {
    "NEED_ID_A": {
         "CURRENT": needinfo,
         "VERSION_1": needinfo,
         "VERSION_2": needinfo,
   },
   # Was not part of a specific version in the past
    "NEED_ID_B": {
         "CURRENT": needinfo,
         "VERSION_2": needinfo,
   },
    # Specified by the first time on the current docs
    "NEED_ID_C": {
         "CURRENT": needinfo,
   },
   # Deleted on the current docs 
    "NEED_ID_D": {
         "VERSION_2": needinfo,
   },
}
```
With data like this we could do a lot of awesome voodoo in future.

**Updated Import mechanism**
Currently `needimport` loads the data of a needs.json directly to the rst and makes all imported needs part of the current document.
But we would need a mechanism to load need.json data without adding this information directly to the docs. May be some option for `conf.py` or maybe even the `:hide:` option of `needimport` could be enough

**Load single need directive**
To show a specific need in a rst, we would need a directive like:

```rst
.. needload: NEED_ID : NEED_VERSION
```
[...]

PR #62, opened 2018-07-11
Resolves issue #61 by ensuring dashes are replaced with underscores

tlovett:
This should resolve the issue I reported in #61.


tlovett:
After playing with this some more, I realized there are a number of characters that exhibit the same behavior.  I'll update this up to replace all of these characters to underscores.

tlovett:
OK, pushed changed to replace all invalid characters

danwos:
Thanks for fixing this.
Do you need a single release for this?
Otherwise I would wait till some of #56, #57, #58 were realized.

tlovett:
No, no need for an immediate release.

Issue #63, opened 2018-07-20
Performance Issue - add_sections


tlovett:
I've discovered a performance issue on large documents and it's related to the add_sections in need.py.  The call to .as_text() in get_sections compounds greatly as the number of sections and documents grows.  

I'll need to find a more efficient was to get the section names.

tlovett:
During profiling, I found another hot spot in need.py's ``process_need_nodes``.  This method has n^2 behavior because it has a nested loop through every need to establish the back links.

```python
        needs = env.need_all_needs
        for key, current_need in needs.items():
            current_need["links_back"] = []
            for linked_key, linked_need in needs.items():
                if current_need["id"] in linked_need["links"]:
                    current_need["links_back"].append(linked_need["id"])
```
This can be simplified so you only need to iterate of all needs once, and then just do a dict lookup for the parent need.  On our internal doc, making this changed dropped time in this method from 4.2s to 0.014s.

danwos:
Okay looks like the pull request introduced some errors.
I get multiple links of the same target inside a need:
![image](https://user-images.githubusercontent.com/998700/43144332-275be14c-8f5d-11e8-8243-411b29e33cf7.png)

danwos:
I got also an KeyError with `needs[link]["links_back"]`, that links_back is not a valid key.
Restarting the build fixes this somehow?! Currently I was not able to reproduce this.
You can push your changes. I will integrate them locally and do some further research based on my problematic sphinx-project.

tlovett:
OK, I pushed up the changes up.  I'm really not sure how you could get KeyError since links_back is set on every need.  I wonder if the new code was run on a pickled environment and that was the problem.

PR #65, opened 2018-07-24
Prevent multiple link backs from being added.

tlovett:
This should prevent multiple link backs from being added to a single parent requirement.  I'm unable to replicate the issue that @danwos raised in #63, but I think this should prevent it.


tlovett:
Did you ever figure out how those duplicates were getting in there?  I tried a couple more things using the roles, but I couldn’t replicate it.

danwos:
Not completely but it was reproducible on other computers as well.

But I recognized that the link_back calculation is done in the event "doctree_ready", which gets called once per document. But used code sets everytime the links_back for all needs. So this calculation is done multiple times.
However this is part of the code since beginning and I think that it may be the reason but I'm unable to tell why it wasn't a problem in the past (and also your change fixed it somehow)

We may switch all event_handlers to "env_ready", which gets called only once. But I'm afraid that needed objects like the status_machine are no longer available and I don't know how to update the docs  in this case.

This affects also the dynamic_handlers, which must to be calculated after all doctrees are resolved but before filters are executed (but they listen actually on "doctree_ready" event)

So it looks like that some big code changes may be needed to perform actions in the right order on the right events.  Still investigating and looking at code from other projects...
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

Issue #80, opened 2019-01-09
needimport: hide attribute not handled properly


maglo:
Importing needs and setting attribute `:hide:` produces incorrect HTML.

Given a correctly formatted `needs.json` and:

```rst
.. needimport:: ../../requirements/business.json
   :hide:
```

Produces:

```text
Security Goal: The System must prevent unauthorized access

hide:
```

Expected that the requirement is imported and hidden.

Probably due to an error in `sphinxcontrib/needs/directives/needimport_template.rst`

PR #81, opened 2019-01-09
Bugfix: needimport fails to properly set hide attribute

maglo:
See #80

Modify import template so that whitespace after closing tags is removed.


maglo:
Please advise regarding testing.

maglo:
figured out the testing and added a test for hidden needs when importing. had to modify test for calc_sum. please review.

PR #92, opened 2019-06-11
Enable sorting on arbitrary numerical attributes

andersthune:
I was working on a project where we wanted to sort needs in a table by a `priority` attribute, but the library only supported sorting by `id` or `status` as far as we could tell. This pull request enables sorting on any alphanumeric attribute. It also removes sorting code that was duplicated between the files `filter_common.py` and `needtable.py` and friends.

### DIFF
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -112,7 +112,7 @@ For **:types:** the type itself or the human-readable type-title can be used as 
 
 sort_by
 ~~~~~~~
-Sorts the result list. Allowed values are ``id`` and ``status``
+Sorts the result list. Allowed values are ``status`` or any alphanumerical property.
 
 .. container:: toggle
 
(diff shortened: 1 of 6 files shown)


danwos:
Closed and reopen because of problems with triggering travis

danwos:
Thanks for the great feature implementation.
I have merged it, but will need some time to extend some test cases.
So it will be part of the next release, which I plan to create in the next 2-3 days.

andersthune:
Great :slightly_smiling_face: 

Thanks for the awesome package!

danieleades:
did this get released?

I have an extra option on my needs called 'priority', and i'm attempting to sort a `needextract` with

```rst
.. needextract::
    :sort_by: priority
    :status: open
```

and i'm getting `WARNING: Sorting parameter priority not valid: Error: 'priority'`

how is this intended to be used?

danwos:
Yes, it got released.
But needextract is newer, maybe a bug was introduced. 
Can you open a bug report, so that I don't miss it and can take a look? Thanks.

PR #95, opened 2019-09-18
Add new config option to control title length in :need: references

hkaimio:
This change adds a new configuration option to control length of title in :need: references. Currently it is hard coded to 30 characters; any titles longer than that are truncated. The new option needs_role_need_max_title_length can be used to change that limit. If set to -1 the tiltle truncation is disabled.


danwos:
Nice idea and even better: awesome PR.
Thanks for updating also tests and docs.
Will test and merge is shortly.

danwos:
Got released under 0.3.14: https://pypi.org/project/sphinxcontrib-needs/0.3.14/
Thanks for your commitment!

PR #120, opened 2020-01-31
Add parameter exclude parameter to meta_links_all

ubmarco:
I added a new parameter `exclude` to the layout function `meta_links_all`. Usecase is to have all links documented excluding an unwanted subset.

Documentation and release notes got also updated.


danwos:
Looks good, thanks

PR #121, opened 2020-01-31
Bugfix in meta_all for unwanted updates to INTERNALS global variable

ubmarco:
`meta_all` never had the links as part of its output. This was due to the fact that the `INTERNALS` variables was assigned to the local variable `default_excludes`. `exclude` was then assigned to  `default_excludes` and got modified later. As only references got passed here, the original `INTERNALS` variable got changed on every call to `meta_all` function which led to a couple of hundred entries.
Now the `INTERNALS` variable gets copied using  `INTERNALS.copy()` so a new list instance gets created.

I tested locally and it works now. Also the `no_links` option works now.
Changelog also got updated with the bugfix.


danwos:
Another big thank you :)

PR #171, opened 2021-02-05
fix copy-paste duplication in common.css

danieleades:
the following code snippet appears twice in this file. I've removed one copy.


danwos:
Thanks for the PR and cleaning up the code :+1:

PR #172, opened 2021-02-05
remove duplicate plantuml legend creation logic

danieleades:


danieleades:
some quick static analysis suggests that almost 40% of this code-base is duplicate lines of code. Hoping to start chipping away at that (very slowly, as I don't have a good understanding of the code yet!)

danwos:
Activated Travis builds for PRs, need to close and reopen this already existing PR to trigger it.

danieleades:
rebased on master

danieleades:
these tests are pretty slow aren't they? I've found github actions to be better than travis- certainly they spend a lot less time idle waiting for server time.
of course, you have to deal with the pain of migrating!

danwos:
Yes, Travis allows 3 parallel builds only.
Not much, but it was enough till you (thankfully) showed up :)

I have nothing against a migration, but it costs time and we won't get so much benefits.
But I stored the tasksfor some more quieter days on my side.

sphinxcontrib/needs/directives/needfilter.py
```diff
@@ -58,10 +60,10 @@ def run(self):
             'docname': env.docname,
             'lineno': self.lineno,
             'target_node': targetnode,
-            'show_tags': self.options.get("show_tags", False) is None,
-            'show_status': self.options.get("show_status", False) is None,
-            'show_filters': self.options.get("show_filters", False) is None,
-            'show_legend': self.options.get("show_legend", False) is None,
+            'show_tags': self.options.get("show_tags", False),
```
danwos, line 63: I think this is not the same logic.

The option is None, if the key exists.
Goal of the replace code was to get True or False, now it will be set to None.

I think ``show_tags in self.options.keys()`` is what you want, if a replacement is needed.

sphinxcontrib/needs/directives/needflow.py
```diff
@@ -96,10 +96,10 @@ def run(self):
             'lineno': self.lineno,
             'target_node': targetnode,
             'caption': caption,
-            'show_filters': self.options.get("show_filters", False) is None,
-            'show_legend': self.options.get("show_legend", False) is None,
-            'show_link_names': self.options.get("show_link_names", False) is None,
-            'debug': self.options.get("debug", False) is None,
+            'show_filters': self.options.get("show_filters", False),
```
danwos, line 99: same here

danwos (changes requested):
Nice refactoring, thanks!
Found one little thing

sphinxcontrib/needs/directives/needfilter.py
danwos: But maybe Sphinx as changed it's logic for boolean values in the meantime.
Not sure as the code got created during Sphinx 1.8.

danieleades: ah i see! will update.

one way to capture this would be to make the 'options' dictionary a class, instead of a plain `dict`. It would be a little verbose, because you don't yet have access to [dataclasses](https://docs.python.org/3/library/dataclasses.html) (requires Python 3.7+). Another option would be to use [Pydantic](https://pydantic-docs.helpmanual.io/), but that still requires Python 3.6+.
The advantage of this approach is if you tried to assign `None` to a field typed as `bool` (which is what i'm erroneously doing here) than your IDE and linter would complain

Anywhere that you changed a dictionary to a class, and that dictionary is exposed in the API, that would be a breaking change. so probably not something to do lightly.

PR #175, opened 2021-02-06
remove python2 compatibility code

danieleades:


danwos:
Close & reopen to trigger travis

danieleades:
the build has failed, but i don't really understand why.
It's complaining that `NEEDS_FUNCTIONS` is not defined, though that hasn't been touched in this PR

also i can see that `NEEDS_FUNCTIONS` is defined in `sphinxcontrib/needs/functions/util`, then imported in `sphinxcontrib/needs/functions/functions`, then declared GLOBAL and referenced in `sphinxcontrib/needs/functions/functions.register_func`- which is where the error is coming from.

danieleades:
fixed the missing import. should be good now

danwos:
Yeah, no Python2 code anymore.
Makes things much easier. Thanks. :+1:

PR #178, opened 2021-02-07
A collection of tiny, pedantic, and unhelpful refactors

danieleades:


sphinxcontrib/needs/directives/needgantt.py
```diff
@@ -194,12 +194,11 @@ def process_needgantt(app, doctree, fromdocname):
...
-            if current_needgantt['milestone_filter'] is None or current_needgantt['milestone_filter'] == '':
-                is_milestone = False
-            else:
+            if current_needgantt['milestone_filter']:
                 is_milestone = filter_single_need(need, current_needgantt['milestone_filter'])
-            if current_needgantt['milestone_filter'] is None or current_needgantt['milestone_filter'] == '' or \
-                    not is_milestone:
+            else:
+                is_milestone = False
+            if not (current_needgantt['milestone_filter'] and is_milestone):
```
danieleades, line 201: even now this has been dramatically simplified, this logic looks a bit fishy to me. Also the exact same logic is repeated several times in this file

sphinxcontrib/needs/directives/needgantt.py
```diff
@@ -81,14 +81,15 @@ def run(self):
...
-            if timeline not in timeline_options:
-                raise NeedGanttException('Given scale value {} is invalid. Please use: '
-                                         '{}'.format(timeline, ",".join(timeline_options)))
+        if timeline and timeline not in timeline_options:
+            raise NeedGanttException(
+                "Given scale value {} is invalid. Please use: "
+                "{}".format(timeline, ",".join(timeline_options))
+            )
         else:
             timeline = None  # Timeline/scale not set later
```
danieleades, line 90: these are not *exactly* equivalent

the original code handles one extra case - that `timeline` is 'falsey' *and* is in `timeline_options`. But since that case can never be met, we can prune that branch

danwos:
Looks good for me. But lets wait for the travis results before it gets merged.
So close & reopen....

danieleades:
Looks like more GitHub API rate limiting errors

sphinxcontrib/needs/diagrams_common.py
```diff
@@ -94,16 +94,16 @@ def collect_diagram_attributes(self):
             caption = self.arguments[0]
 
         collected_diagram_options = {
-            'show_legend': True if self.options.get("show_legend", False) is None else False,
```
danieleades, line 97: @danwos  what is this meant to be checking?

`self.options.get("show_legend", False) is None` will *always* evaluate to `False` since you're comparing a bool to `None`, right?

also i've simplified
```python
True if self.options.get("show_legend", False) is None else False
```
to
```python
self.options.get("show_legend", False) is None
```
which can be further simplified to
```python
not self.options.get("show_legend")
```
on the assumption that `self.options.get("show_legend", False) is None` is supposed to be `self.options.get("show_legend", None) is None`

but that seems an inversion of the intention? ie. this will be true if "show_legend" is not set, or is set to None, and false if "show_legend" is set to anything except `None`.

is this meant to just be


```python
self.options.get("show_legend", False)
```

???

danieleades: i've updated this to what i *think* this was meant to be checking all along...

danieleades: i guess i did break the logic here-

```
sphinxcontrib.needs.layout.SphinxNeedLayoutException: Given layout "" is unknown for need FEATURE_3. Registered layouts are: test ,clean ,clean_l ,clean_lp ,clean_r ,clean_rp ,complete ,focus ,focus_f ,focus_l ,focus_r ,debug ,example ,github
```

should these be


```python
bool(self.options.get("show_legend", False))
```

??

i still think the problem lies with `self.options` being 'stringly'-typed

sphinxcontrib/needs/layout.py
```diff
@@ -57,13 +57,13 @@ def create_need(need_id, app, layout=None, style=None, docname=None):
     node_inner.attributes['ids'].append(need_id)
 
     if layout is None:
-        if need_data['layout'] is None or len(need_data['layout']) == 0:
+        if need_data['layout']:
```
danieleades, line 60: there's a bug here, which is related to the failing CI.

I seem to have uncovered it by cleaning up the boolean logic above.

i'll add some comments to the original code to illustrate what i mean here

```python
if layout is None:
    if need_data['layout'] is None or len(need_data['layout']) == 0  # this is equivalent to `if need_data['layout']`
        layout = getattr(app.config, 'needs_default_layout', 'clean')
    else:  # need_data['layout'] is 'Falsey'
        layout = need_data['layout']  # why are we assigning this? we *just* established that it's Falsey!
```

danieleades:
i might leave this diff for now, and make a new pull request to address *only* the boolean logic. I think that might be tidier

sphinxcontrib/needs/diagrams_common.py
danwos: Yeah there is a lot of voodoo.
Is far as I remember the flag handling in sphinx is a little bit strange.

Example:
```
.. need::
   :show_options:   #Simple flag, no Value!
```
To figure out, if ``show_options`` is set, it must be part of ``self.options``, but it's value will be ``None``.
If it is not set, it's not part of ``self.options``.
So we can't just copy its value, as it is not set.
So the original ``show_legend`` from sphinx is not a bolean, my code tries to make it to one (in a really too complex way.)

I think ``True if self.options.get("show_legend", False) is None else False`` can be transformed to ``"show_legend in self.options``.

But I have already recognized that your logic is "much stronger" as mine ;)

danwos:
See my comment above.
I think your assumption what Sphinx put into these vars was wrong.
Only its existence tells you if option was set or not.
So it looks like you are on a good way of cleaning up everything the right way.
thanks for your effort :+1:

sphinxcontrib/needs/diagrams_common.py
danieleades: ah I see, thankyou. I was thinking the check was trying to see if the value was set *or* was `False`, but if it's sufficient to check whether it exists at all then `"show_legend" in self.options` is probably the way to go!

> But I have already recognized that your logic is "much stronger" as mine ;)

Dunno about that. sometimes you just need fresh eyes on something

danieleades:
> I tried to fix also the builds, by mocking all calls to external services in test cases.
> The test should be reliable again.

brilliant! i've just rebased, so we'll see how we go. If the tests are still failing because of my changes i'll split this diff up a little. I was only intending to make 'cosmetic' changes in this pull request anyway, so if i've broken any logic it's because i was getting too ambitious.

danwos:
Nice refactoring, learned some new stuff.
Thanks a lot.

PR #180, opened 2021-02-07
fix boolean logic

### LINKED ISSUE #179: 'needgantt' logic seems to be trying to compare the length of a vector with an empty string
https://github.com/useblocks/sphinxcontrib-needs/blob/a7df6d096611dc68092cbe6dc66ce7cbb75b83e6/sphinxcontrib/needs/directives/needgantt.py#L290

since `len(current_needgantt['align']) != ''` is a comparison between and integer and a string, this will always return `False`

unless i'm missing something?

danieleades:
closes #179

is this the logic that was intended here?


danwos:
Yes, logic looks good.
Thanks for this improvement.

PR #182, opened 2021-02-08
fix logic bug

### LINKED ISSUE #181: possible logic bug in `layout.py`
possible logic bug in `layout.py`

https://github.com/useblocks/sphinxcontrib-needs/blob/a7df6d096611dc68092cbe6dc66ce7cbb75b83e6/sphinxcontrib/needs/layout.py#L592

i'm pretty sure this isn't doing what it's supposed to do

```python
if not show_empty and result is None or len(str(result)) == 0 or bool(result) is False:
    continue
```

by operator precendence, is equivalent to

```python
if (not show_empty and result is None) or len(str(result)) == 0 or bool(result) is False:
    continue
```

which doesn't really make any sense.

Is this perhaps what this is *trying* to check? - 

```python
if not show_empty and not result:
    continue
```

or, equivalently - 

```python
if not (show_empty or result):
    continue
```

(by De Morgan's laws)

danieleades:
closes #181

### DIFF
--- a/sphinxcontrib/needs/layout.py
+++ b/sphinxcontrib/needs/layout.py
@@ -587,7 +587,7 @@ class LayoutHandler:
                 continue
 
             data_line = nodes.line()
-            label = f'{prefix}{data}:{postfix} '
+            label = prefix + '{}:'.format(data) + postfix + ' '
             result = self.meta(data, label, show_empty)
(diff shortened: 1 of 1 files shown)


danwos:
And as usual:  Thanks for your PR :)

PR #191, opened 2021-02-14
WIP: replace a few instances of 'os.path' with 'pathlib'

danieleades:
since i can't run the tests on my machine, i'm going to do this bit by bit and use the CI to check my work


danieleades:
i applied some formatting to `sphinxcontrib.needs.utils` using `Black` and `iSort`. Happy to revert this if it's not wanted.

@danwos do you have any feelings about formatting, and in particular line length?

danwos:
Yes, line length is set to 120 chars in my linter config.
I think the "normal" 80 are too less, fragment the code and forces you to use short variable names instead of understandable ones.
But I'm open to discuss it.

For the rest I don't care, as long as it follows the "mainstream", is testable and is supported by most common IDEs.

danieleades:
some tests are failing with python 3.5. attempted to open files using `open(path) as f` rather than with `path.read_text()` to see if that helps.

danieleades:
i think i'll need to abandon a lot of these changes until python 3.5 support is dropped. pathlib support doesn't seem to be quite there in either the standard library or possibly the sphinx_testing fixtures, which is leading to some edge cases.

will roll back some of these commits

danieleades:
closing in favour of #198

PR #192, opened 2021-02-14
remove 'install_feather_icons' no-op

danieleades:
this is dead code, i think this can be safely removed


danwos:
Thanks for this PR. :+1:

PR #198, opened 2021-02-16
refactor environment.py

danieleades:


danwos:
Gave me some headache at some places to understand the changes :)
But it was also enlightened to see how my old code can be rewritten.
So good work, thanks for all of it!

PR #200, opened 2021-02-17
Changed github service max_amount default to -1 instead of 5

### LINKED ISSUE #199: GitHub services returns maximum five contents
I observed that GitHub services returns by default a max_amount of 5 (tested with github-issues). To get all contents, I had to set max_amount to -1. Is this the expected behavior?

danwos:
Fixes #199


twodrops:
Even with "-1" there seems to be a limit set by the server to return by default only a fixed set of issues (ca. 25). As @danwos  suggested in #199, I guess it is better to leave the default as "-1". I set for for 500 in the config file for now.

PR #202, opened 2021-02-21
refactor filter_common

danieleades:
small refactors so far. I was hoping to ask a few questions about the API to hopefully refactor further-

### DIFF
--- a/sphinxcontrib/needs/directives/needfilter.py
+++ b/sphinxcontrib/needs/directives/needfilter.py
@@ -63,7 +63,7 @@ class NeedfilterDirective(FilterBase):
             'show_filters': 'show_filters' in self.options,
             'show_legend': 'show_legend' in self.options,
             'layout': self.options.get("layout", "list"),
-            'export_id': self.options.get("export_id", ""),
+            'export_id': self.options.get("export_id"),
             'env': env,
         }
         env.need_all_needfilters[targetid].update(self.collect_filter_attributes())
--- a/sphinxcontrib/needs/directives/needfilter.py
+++ b/sphinxcontrib/needs/directives/needfilter.py
@@ -63,7 +63,7 @@ class NeedfilterDirective(FilterBase):
             'show_filters': 'show_filters' in self.options,
             'show_legend': 'show_legend' in self.options,
             'layout': self.options.get("layout", "list"),
-            'export_id': self.options.get("export_id"),
+            'export_id': self.options.get("export_id", ""),
             'env': env,
         }
         env.need_all_needfilters[targetid].update(self.collect_filter_attributes())
--- a/sphinxcontrib/needs/filter_common.py
+++ b/sphinxcontrib/needs/filter_common.py
@@ -6,7 +6,7 @@ like needtable, needlist and needflow.
 import re
 import copy
 from itertools import product
-from typing import Dict, Any
+from typing import Dict, Any, List
 
 from docutils.parsers.rst import Directive
 from docutils.parsers.rst import directives
@@ -176,7 +176,7 @@ def prepare_need_list(need_list):
     return all_needs_incl_parts
 
 
-def check_need_list(list_a, list_b):
+def check_need_list(list_a, list_b) -> List[Dict[str, Any]]:
 
     def get_id(element: Dict[str, Any]) -> str:
         id = element['id']
@@ -184,12 +184,7 @@ def check_need_list(list_a, list_b):
             id = "{}.{}".format(element['id_parent'], id)
         return id
 
-    common = []
-    for a, b in product(list_a, list_b):
-        if get_id(a) == get_id(b):
-            common.append(a)
-
-    return common
+    return [a for a, b in product(list_a, list_b) if get_id(a) == get_id(b)]
 
 
 def filter_needs(needs, filter_string="", current_need=None):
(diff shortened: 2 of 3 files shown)


danieleades:
there's a lot of type-checking in `sphinxcontrib.needs.filter_common`
for example- 

```python
def collect_filter_attributes(self):
        tags = str(self.options.get("tags", ""))
        if tags:
            tags = [tag.strip() for tag in re.split(";|,", tags) if len(tag) > 0]

        status = self.options.get("status", None)
        if status:
            try:
                status = str(status)
                status = [stat.strip() for stat in re.split(";|,", status)]
            except Exception:
                # If we could not transform/use status information, we just skip this status
                pass
        else:
            status = []

        types = self.options.get("types", [])
        if isinstance(types, str):
            types = [typ.strip() for typ in re.split(";|,", types)]

        # Add the need and all needed information
        collected_filter_options = {
            'status': status,
            'tags': tags,
            'types': types,
            'filter': self.options.get("filter", None),
            'sort_by': self.options.get("sort_by", None),
            'filter_code': self.content,
            'export_id': self.options.get("export_id", ""),
        }
        return collected_filter_options
```

what possible types can `self.options["tags"]` and `self.options["status"]` be?

danieleades:
rebased on master

danieleades:
will remove the 'WIP'. there's a lot more that can be done, but this is already a useful refactor

sphinxcontrib/needs/filter_common.py
```diff
@@ -80,16 +75,12 @@ def process_filters(all_needs, current_needlist):
...
-        else:
-            try:
-                sorted_needs = sorted(all_needs, key=lambda node: node[sort_key])
-                all_needs = sorted_needs
-            except Exception as e:
-                logger.warning(
-                    "Sorting parameter {0} not valid: Error: {1}".format(sort_key, e)
-                )
+        try:
+            all_needs = sorted(all_needs, key=lambda node: node[sort_key])
```
danwos, line 79: what if a value is None?
then sorted will throw an exception as comparison between None and string does not work

danwos: Maybe we should have a test case, where certain fields are set and not set.
Then test the here changed code.

It all may look simpler now,  but the checks are needed if you can't "trust" the incoming data.
I personal don't like to trust data validation on upper levels only.
The upper level may not know, what kind of validation is needed here and it even may change.
Python version change, Sphinx version , ....
So my goal is to be able to sort here everything, not matter how it looks like. For status at least: "abc", "", " " and None.

sphinxcontrib/needs/filter_common.py
```diff
@@ -188,24 +179,15 @@ def prepare_need_list(need_list):
...
-                common_list.append(element_a)
-                break
+    def get_id(element: Dict[str, Any]) -> str:
+        id = element['id']
+        if element['is_part']:
+            id = "{}.{}".format(element['id_parent'], id)
+        return id
 
-    return common_list
+    return [a for a, b in product(list_a, list_b) if get_id(a) == get_id(b)]
```
danwos, line 190: Holy crap, just learned new stuff.
Great code optimization.

sphinxcontrib/needs/utils.py
```diff
@@ -90,18 +93,7 @@ def row_col_maker(app, fromdocname, all_needs, need_info, need_key, make_ref=Fal
     return row_col
 
 
-def status_sorter(a):
```
danwos, line 93: See comments above, not sure if we still need it

danwos (commented):
Thanks for the optimizations.

At 1-2 points I have the feeling we drop to much checks and rely too much on earlier checks from Sphinx itself.

danwos:
> what possible types can `self.options["tags"]` and `self.options["status"]` be?

`tags` is normally a comma-separated string or None
`status` is a string or None

danieleades:
rebase on master

sphinxcontrib/needs/filter_common.py
danieleades: i'll revert this change.

i have to respectfully disagree with putting the type checking at the 'lower level'. It would be far better to enforce the typing at the API boundary, so that your code-base only has to deal with sanitised data. that way any type errors are caught by a linter (like mypy), rather than with messy type-checks in every function.

the way you'd do this is by changing the 'node info' from a stringly-typed `dict` to a strongly-typed concrete type (then you would *know* the type of `node[sort_key]`, without needing to check). that's a big change, and a problem for another day!

sphinxcontrib/needs/directives/needfilter.py
```diff
@@ -65,7 +65,7 @@ def run(self):
             'show_filters': 'show_filters' in self.options,
             'show_legend': 'show_legend' in self.options,
             'layout': self.options.get("layout", "list"),
-            'export_id': self.options.get("export_id", ""),
+            'export_id': self.options.get("export_id"),
```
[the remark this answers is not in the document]
danieleades: would be better if the filter config were a class, rather than a dict. I won't touch that in this pull request, i'll just revert the change, since you're right, i don't know what the downstream code will do.

certainly `None` would make more sense than an empty string, semantically. right?

sphinxcontrib/needs/utils.py
danieleades: don't think we need it, i've changed the above to

```python
all_needs = sorted(all_needs, key=lambda node: node[sort_key] or "")
```

does that do what you're after?

danwos: Yes, thanks. Looks good.

PR #212, opened 2021-02-27
Refactor configuration

danieleades:
i'm refactored the configuration in the setup function, and a few places where that configuration is consumed.

There's a lot of checking attributes and types throughout, which is unnecessary since the sphinx application API ensures config values are present, have defaults, and are the correct type. there's no need to do this again.

i've also removed a `global` parameter ("NEEDS_FUNCTIONS_CONF"). It didn't seem to be doing anything, and global variables are almost never what you want

### DIFF
--- a/sphinxcontrib/needs/directives/need.py
+++ b/sphinxcontrib/needs/directives/need.py
@@ -81,15 +81,13 @@ class NeedDirective(Directive):
 
         # ToDo: Keep this in directive!!!
         collapse = self.options.get("collapse", None)
-        if isinstance(collapse, str) and len(collapse) > 0:
+        if isinstance(collapse, str):
             if collapse.upper() in ["TRUE", 1, "YES"]:
                 collapse = True
             elif collapse.upper() in ["FALSE", 0, "NO"]:
                 collapse = False
             else:
                 raise Exception("collapse attribute must be true or false")
-        else:
-            collapse = getattr(env.app.config, "needs_collapse_details", None)
 
         hide = True if "hide" in self.options.keys() else False
 
@@ -296,9 +294,7 @@ def process_need_nodes(app, doctree, fromdocname):
         for index, attribute in enumerate(node_need.attributes['classes']):
             node_need.attributes['classes'][index] = check_and_get_content(attribute, need_data, env)
 
-        layout = need_data['layout']
-        if layout is None or len(layout) == 0:
-            layout = getattr(app.config, 'needs_default_layout', 'clean')
+        layout = need_data['layout'] or app.config.needs_default_layout
 
         build_need(layout, node_need, app)
 
--- a/sphinxcontrib/needs/needs.py
+++ b/sphinxcontrib/needs/needs.py
@@ -248,25 +248,23 @@ def setup(app):
             'parallel_write_safe': True}
 
 
-def load_config(app, *args):
+def load_config(app: Sphinx, *args):
     """
     Register extra options and directive based on config from con.py
     """
     types = app.config.needs_types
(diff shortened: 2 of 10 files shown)


sphinxcontrib/needs/directives/need.py
```diff
@@ -81,15 +81,13 @@ def run(self):
...
-        if isinstance(collapse, str) and len(collapse) > 0:
+        if isinstance(collapse, str):
             if collapse.upper() in ["TRUE", 1, "YES"]:
                 collapse = True
             elif collapse.upper() in ["FALSE", 0, "NO"]:
                 collapse = False
             else:
                 raise Exception("collapse attribute must be true or false")
-        else:
-            collapse = getattr(env.app.config, "needs_collapse_details", None)
```
danwos, line 92: Why can this be removed?
It's load the value from conf.py if it is not set via directive option?

sphinxcontrib/needs/needs.py
```diff
@@ -1,148 +1,243 @@
...
+    app.add_config_value("needs_duration_option", "duration", "html")
+    app.add_config_value("needs_completion_option", "completion", "html")
 
     # If given, only the defined status are allowed.
     # Values needed for each status:
     # * name
     # * description
     # Example: [{"name": "open", "description": "open status"}, {...}, {...}]
-    app.add_config_value('needs_statuses', [], 'html')
+    app.add_config_value("needs_statuses", [], "html")
```
danwos, line 187: Is there any common rule why to prefer " ?
I normally use ' as it is easier to reach on german keyboard layout.
And " when I need to use ' in the string itself:
- ``"Hi, my name is 'foo' "``
or ``f"Hi, name name is {config['name']}"``

Just want to be sure that we have a reason why this gets changed.
If there is no reason, I personal would like to stay with ', as it makes thing easier.

danwos:
Thanks for the refactoring.
Yes ``NEEDS_FUNCTIONS_CONF`` seems to do nothing.
I'm not sure why I have used it or if it some leftover from a refactoring.

sphinxcontrib/needs/needs.py
danieleades: i'm using [Black](https://github.com/psf/black) to format these.

Black is a deliberately opinionated and unconfigurable formatter. It offers no way to configure the formatting of single quotes vs double quotes

see here - https://github.com/psf/black/blob/master/docs/the_black_code_style.md
and in particular here - https://github.com/psf/black/blob/master/docs/the_black_code_style.md#strings

> On certain keyboard layouts like US English, typing single quotes is a bit easier than double quotes. The latter requires use of the Shift key. My recommendation here is to keep using whatever is faster to type and let Black handle the transformation.

that said, i don't have a strong opinion, and Black is far from the only option.

I think regardless of the style that you choose @danwos, this project should absolutely recommend a formatting tool and provide a configuration (unless it's Black and there is no configuration!)

sphinxcontrib/needs/directives/need.py
danieleades: it states in the comments and documentation that "needs_collapse_details" is deprecated, in favour of setting in `layout`. It's also not set anywhere as a configuration option in the `setup` function, so `collapse = getattr(env.app.config, "needs_collapse_details", None)` will *always* return None

sphinxcontrib/needs/needs.py
danieleades: maybe it's best to drop the formatting commit from this pull request and open a separate issue?

unfortunately that commit is pretty low in the stack, so i've created a pretty horrible rebase for myself, i suspect ;)

danwos:
And again, thanks for this huge clean up.

PR #217, opened 2021-03-02
Formatting

### LINKED ISSUE #213: Consistent Formatting
I think this project should choose a formatting style and enforce it.

That is, it should-

- stipulate a particular formatter
- provide a configuration file for that formatter (if applicable)
- check that it has been applied in CI

I *strongly* recommend [Black](https://github.com/psf/black/), precisely because it's almost completely unconfigurable (except for line length, default is 88).
I would also throw in `iSort` for PEP8-compliant import formatting

flake8 has plugins that can enforce both of these - `flake8-black` and `flake8-isort`

danieleades:
this is what it would look like if sphinx-needs enforced a formatting style.

note that while i've used Black for this pull request, any formatting style could be used.

This pull request is stacked on top of #216

i've added

- [x] formatting dev dependencies (black + isort)
- [x] Makefile targets for applying and checking formatting
- [x] CI job for checking the formatting

you can see the result of the jobs here - https://github.com/danieleades/sphinxcontrib-needs/actions

(I think 'actions' need to be enabled for this repository before you will see them *here*)

closes #213


danieleades:
@danwos i think you mentioned you prefer a line width of 120?

that's the *one thing* you *can* configure with Black!

danieleades:
i've bumped the max line length from 88 to 120 to match the previous settings.

which do you prefer?

danwos:
> i've bumped the max line length from 88 to 120 to match the previous settings.
> 
> which do you prefer?
120 chars is fine :)

danieleades:
rebased on #216

PR #218, opened 2021-03-07
tidy 'urlparse' imports

danieleades:
I noticed most of the urlib imports are unused. Also there are two different 'flavours' of `urlparse` used in different places.


danwos:
Thanks for the PR.
I'm looking forward to implement new stuff after you cleaned-up the code so much :+1:

PR #221, opened 2021-03-14
Simplify platform dependent code

swaldhoer:
- simplify code by using ``os.path.join()``
- fix wrong path to plantuml jar in documentation

### DIFF
--- a/tests/doc_test/add_sections/conf.py
+++ b/tests/doc_test/add_sections/conf.py
@@ -41,15 +41,7 @@ needs_extra_options = {
     "impacts": directives.unchanged
 }
 
-cwd = os.getcwd()
-plantuml = 'java -jar %s' % os.path.join(cwd, "../utils/plantuml_beta.jar")
-
-# If we are running on windows, we need to manipulate the path,
-# otherwise plantuml will have problems.
-if os.name == "nt":
-    plantuml = plantuml.replace("/", "\\")
-    plantuml = plantuml.replace("\\", "\\\\")
-
+plantuml = 'java -jar %s' % os.path.join(os.getcwd(), "..", "utils", "plantuml_beta.jar")
 plantuml_output_format = 'svg'
 
 # Add any paths that contain templates here, relative to this directory.
--- a/tests/doc_test/add_sections/conf.py
+++ b/tests/doc_test/add_sections/conf.py
@@ -41,7 +41,7 @@ needs_extra_options = {
     "impacts": directives.unchanged
 }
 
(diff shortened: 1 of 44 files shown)


danwos:
Thanks for the PR. :+1: 
Makes totally sense to refactor this.

I just noticed that some configs use ``os.getcwd()`` and some use ``os.path.dirname(__file__)``.
Technically being based on ``cwd`` makes not really a sense and ``os.path.dirname(__file__)`` is the better choice.

So, because you are currently changing this part, would it be okay to use ``os.path.dirname(__file__)`` everywhere instead of ``os.getcwd()``

swaldhoer:
Yes, makes sense, I will update the PR tomorrow 👍

danieleades:
Should this use `pathlib` instead of `os.path`?

tests/doc_test/add_sections/conf.py
```diff
@@ -41,15 +41,7 @@
...
-cwd = os.getcwd()
-plantuml = 'java -jar %s' % os.path.join(cwd, "../utils/plantuml_beta.jar")
-
-# If we are running on windows, we need to manipulate the path,
-# otherwise plantuml will have problems.
-if os.name == "nt":
-    plantuml = plantuml.replace("/", "\\")
-    plantuml = plantuml.replace("\\", "\\\\")
-
+plantuml = 'java -jar %s' % os.path.join(os.path.dirname(__file__), "..", "utils", "plantuml_beta.jar")
```
danieleades, line 44: rather than define this in every file, do you think it might be better to define it *once* and then import it into each file?
that way if it changes, you only need to change it one place.

admittedly you *could* do it with a find + replace instead

swaldhoer:
@danwos  I have removed the ``os.getcwd()`` calls in favor of ``os.path.dirname(__file__)``.

tests/doc_test/add_sections/conf.py
swaldhoer: This can only bee achived when  messing around with ``sys.path`` to have have such a helper file importable everywhere you need it. I don't like this, but if @danwos likes this solution, we can do it that way. Personally I would not change it. The current implementation is simple, it works and it is unlikely that this will change.

danieleades: I'm not sure I understand. Why does it involve messing with sys.path?

What happens if you calculate the absolute path *once* and then pass it around as a constant?

danieleades: Surely you would just stick an __init__.py file in the tests/ root and then use a relative import?

danwos:
Thanks for the PR and updating it. 
The suggested import mechanism may be nice, but if really needed we can have another PR.
Works also without it :)

PR #224, opened 2021-04-11
refactor string formatting

danieleades:
# changes

- use lazy formatting for logging strings

(this diff is stacked on #217)


danieleades:
there's a couple of suspect logging calls-

`sphinxcontrib.needs.directives.needflow.split_link_types`:

```python
def split_link_types(link_types: str) -> Iterable[str]:
    def is_valid(link_type) -> bool:
        if len(link_type) == 0 or link_type.isspace():
            logger.warning(
                "Scruffy link_type definition found in needflow {}. "
                "Defined link_type contains spaces only.".format(id)
            )
            return False
        return True

    return filter(
        is_valid,
        (x.strip() for x in re.split(";|,", link_types)),
    )
```

the format string uses the parameter `id`, but this variable doesn't appear to be in scope. an IDE won't complain, because there *is* a built-in function with the same name that *is* in scope. Is this a bug?

it's also present in 2 other places - `
- `sphinxcontrib.needs.directives.need.NeedDirective.read_in_links`
- `sphinxcontrib.needs.diagrams_common.DiagramBase.collect_diagram_attributes`

danieleades:
closing. not sure the performance benefits are worth it. might revisit one day

PR #234, opened 2021-04-17
needimport supports extra options and links

### LINKED ISSUE #227: needimport ignores all links between needs
I have a proper `needs.json` containing some valid needs that are linked to each other. Now when I "needimport" this file, all needs are imported, but without links. I couldn't find a hint in the documentation, that specifies this behavior: https://sphinxcontrib-needs.readthedocs.io/en/latest/directives/needimport.html

Probably this is a bug? Or are there any other way to achieve this?

danwos:
* extra options and links are imported
* pre/post/template can be defined by needimport
* same for style and layout

Fixes #227

### DIFF
--- a/sphinxcontrib/needs/api/need.py
+++ b/sphinxcontrib/needs/api/need.py
@@ -6,6 +6,7 @@ from docutils import nodes
 from docutils.statemachine import ViewList
 from jinja2 import Template
 from sphinx.util.nodes import nested_parse_with_titles
+from typing import List, Union
 
 import sphinxcontrib.needs.directives.need
 from sphinxcontrib.needs.api.exceptions import (
@@ -369,7 +370,7 @@ def _render_template(content, docname, lineno, state):
     return node_need_content
 
 
-def _read_in_links(links_string):
+def _read_in_links(links_string: Union[str, List[str]]):
     # Get links
     links = []
     if links_string:
--- a/sphinxcontrib/needs/api/need.py
+++ b/sphinxcontrib/needs/api/need.py
@@ -144,7 +144,10 @@ def add_need(app, state, docname, lineno, need_type, title, id=None, content="",
     if tags is None:
         tags = []
     if len(tags) > 0:
-        tags = [tag.strip() for tag in re.split(";|,", tags)]
+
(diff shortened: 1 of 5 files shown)


sphinxcontrib/needs/api/need.py
```diff
@@ -370,7 +373,13 @@ def _read_in_links(links_string):
...
     links = []
     if links_string:
-        for link in re.split(";|,", links_string):
+        # Check if links_string is really a string, otherwise it will be a list, which can be used
+        # without modifications
+        if isinstance(links_string, str):
+            link_list = re.split(";|,", links_string)
+        else:
+            link_list = links_string
+        for link in link_list:
```
danieleades, line 382: do you want to add a type annotation to this function?

```python
def _read_in_links(links_string: Union[str, List[str]]):
    ...
```
?

sphinxcontrib/needs/directives/needimport.py
```diff
@@ -120,15 +119,49 @@ def run(self):
...
-        content = template.render(needs_list=needs_list, hide=hide, id_prefix=id_prefix)
-        self.state_machine.insert_input(content.split('\n'),
-                                        self.state_machine.document.attributes['source'])
+        # template_location = os.path.join(os.path.dirname(os.path.abspath(__file__)), "needimport_template.rst")
+        # with open(template_location, "r") as template_file:
+        #     template_content = template_file.read()
+        # template = Template(template_content)
+        # content = template.render(needs_list=needs_list, hide=hide, id_prefix=id_prefix)
+        # self.state_machine.insert_input(content.split('\n'),
+        #                                 self.state_machine.document.attributes['source'])
```
danieleades, line 128: if this is no longer used, should it just be deleted?

sphinxcontrib/needs/api/need.py
danwos: Yes, will add it.
Sorry I'm not so familiar with type annotations (till now) :)

sphinxcontrib/needs/directives/needimport.py
danwos: Ohh yes.
Will add you next time as reviewer.
Thanks.

PR #262, opened 2021-05-04
lint using pre-commit

danieleades:
here's another experiment in automatic linting, this time using [pre-commit](https://pre-commit.com/)

i've also added

- flake8-bugbear: opinionated lints
- pyupgrade: checks for, and automatically updates deprecated syntax

interested to hear what you think about this approach


danieleades:
i've not enabled it, but the pre-commit github action can also be configured to automatically fix and commit to pull requests. Depends whether you prefer to *fix* pull requests that don't meet the standards, or simply *reject* them.

danwos:
Wow, that's really helpful little tool.
Like it much.

But we should not fix PRs on our own.
Just check it and report the problems.
Will force committers to understand our checks and tool chain. 

Can you also update the ``contribute`` page?
I think the old ``make lint`` is mentioned there and we should write a few words about how to set up pre-commit.

The rest looks all good.
Thanks for this great enhancement for our tool chain. :+1:

danieleades:
closing in favour of #266

PR #266, opened 2021-05-05
lint using pre-commit

danieleades:
here's another experiment in automatic linting, this time using [pre-commit](https://pre-commit.com/)

i've also added

- flake8-bugbear: opinionated lints
- pyupgrade: checks for, and automatically updates deprecated syntax


pyproject.toml
```diff
@@ -29,7 +29,7 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.6"
+python = "^3.6.1"
```
danieleades, line 32: this is a pre-commit requirement. it's possible to avoid bumping this version, by instead specifying that pre-commit is installed as a system dependency, instead of using Poetry. do you have a preference?

docs/contributing.rst
```diff
@@ -83,6 +83,16 @@ Running Tests
 Linting
 -------
 
+This project uses [pre-commit](https://pre-commit.com/) for linting.
```
danwos, line 86: This is markdown syntax and does not get rendered. :)
Please stay with rst.

docs/contributing.rst
```diff
@@ -83,6 +83,16 @@ Running Tests
...
+
+1. install pre-commit hooks (this only needs to be done once)
+
+::
+
+   poetry run pre-commit install
+
+2. run/apply lints
+
```
danwos, line 95: Not so happy with the docs.
I think pre-commit gets not introduced to support linting.
It's more or less a general quality gate which needs to be passed before a commit can be pushed.

So we need paragraph how to create and send PRs to this project.
If you like, you can already add it.
Or I will do it with a new PR (after yours got merged).

pyproject.toml
danwos: It's fine to have it as project dependency, as the version diff is not really soo huge :)

danwos (changes requested):

docs/contributing.rst
danieleades: my bad! I live in a world of markdown usually.

I *think* i've fixed this properly, but please give this a sanity-check for me!

docs/contributing.rst
danieleades: can you suggest a wording that you prefer?

or i can leave it to you to update in a following commit? whichever you prefer

danwos (approved):

danieleades:
this PR was a little ambitious anyway- too much in one go.

closing in favour of #415

Issue #137, opened 2020-02-24
Is it possible to reference needs with intersphinx? / Add intersphinx support


lassebn:
As the title says I'd like to know if it is possible to cross reference using intersphinx.

As a try out I tried adding the following to `conf.py`

```python
extensions = [
  "sphinx.ext.intersphinx",
]

intersphinx_mapping = {
  'need': (
    'https://sphinxcontrib-needs.readthedocs.io/en/latest/',
     None
  )
}
```

and referencing using
```
:need:`:need:SPEC-1`
```
or
```
:need:`SPEC-1`
```

(`SPEC-1` is defined on the front page of the documentation), but no such luck. Has anybody tried this?

danwos:
No, I'm afraid this is not possible and I haven't worked with intersphinx up to now.

Okay, did just a fast training on intersphinx and how it works.
And Sphinx-Needs is definitely doing nothing to get its data into the ``object.inv``.
Here is a link to the interpreted content of the ``object.inv`` file of Sphinx-Needs docs:
https://gist.github.com/danwos/8ac053396f449c52e569de2edcbe7712
No need-ids in it :(

I'm just not sure how to get own data into the ``object.inv`` file.
It looks like that Sphinx does this automatically for all registered Domains.

So one solution would be to register a domain for need. 
But then we have to add automatically all needs to this domain.
Not sure if there will be any conflicts, as Sphinx-Needs is not using domains and there may be some default handling for domains, which we don't want... (module entry, needed representation in docs, ...)

Another solution would be to add our needs to the ``std:label`` domain.
But here the same as above, how and may there be any conflicts?

I'm willing to support this great feature, but I have to do some tests, to figure out how to do it.
So it will need some time...

If there is anybody who can help, maybe with a PR, you are welcome :)

lassebn:
Thank you for your fast reply and willingness to support intersphinx in the future 🙌

I've never done any work on any extensions for Sphinx yet. At the moment I'm just using your excellent extension to write some documentation.

Having the ability to link needs across projects might become important enough for us at my work for me to get time allocated to work on it. Realistically that will be months out in the future though :(

danwos:
The last days I have thought a lot about this feature and how to implement it.
My proposal is to **not support intersphinx**  but to provide another mechanism, which allows the same behavior: Getting links to external needs into your document.

It would be great, if you could give me your opinion, if it covers your case.

### Reason why not using intersphinx
[...]

PR #270, opened 2021-05-11
External needs

danwos:
see #137


sphinxcontrib/needs/api/need.py
```diff
@@ -158,7 +168,12 @@ def run():
         )
 
     # Calculate target id, to be able to set a link back
-    target_node = nodes.target("", "", ids=[need_id], refid=need_id)
+    if not is_external:
+        target_node = nodes.target("", "", ids=[need_id], refid=need_id)
+        external_url = None
+    else:
+        target_node = None
+        external_url = external_url
```
danieleades, line 176:
```suggestion
    if is_external:
        target_node = None
        external_url = external_url
    else:
        target_node = nodes.target("", "", ids=[need_id], refid=need_id)
        external_url = None
```

sphinxcontrib/needs/utils.py
```diff
@@ -145,6 +156,22 @@ def rstjinja(app: Sphinx, docname: str, source):
     source[0] = rendered
 
 
+def import_prefix_link_edit(env, needs, id_prefix):
```
danieleades, line 159: instead of passing in the `env` object, are you able to just pass in the object you actually *need*? namely the `needs_extra_links`. as far as i can see that's the only `env` attribute you use in this method.

so something like

```python
def import_prefix_link_edit(needs: List[??], id_prefix: str, needs_extra_links: List[??])
   ...
```

sphinxcontrib/needs/utils.py
```diff
@@ -145,6 +156,22 @@ def rstjinja(app: Sphinx, docname: str, source):
...
+    needs_ids = needs.keys()
+
+    for key, need in needs.items():
```
danieleades, line 162:
```suggestion
    for need in needs.values():
```

sphinxcontrib/needs/utils.py
```diff
@@ -145,6 +156,22 @@ def rstjinja(app: Sphinx, docname: str, source):
...
+        for id in needs_ids:
+            # Manipulate links in all link types
+            for extra_link in env.config.needs_extra_links:
+                if extra_link["option"] in need.keys() and id in need[extra_link["option"]]:
+                    for n, link in enumerate(need[extra_link["option"]]):
+                        if id == link:
+                            need[extra_link["option"]][n] = "".join([id_prefix, id])
+            # Manipulate descriptions
+            # ToDo: Use regex for better matches.
+            need["description"] = need["description"].replace(id, "".join([id_prefix, id]))
```
danieleades, line 172:
```suggestion
            need["description"] = need["description"].replace(id, f"{id_prefix}{id})
```

sphinxcontrib/needs/utils.py
```diff
@@ -145,6 +156,22 @@ def rstjinja(app: Sphinx, docname: str, source):
...
+    needs_ids = needs.keys()
+
+    for key, need in needs.items():
+        for id in needs_ids:
+            # Manipulate links in all link types
+            for extra_link in env.config.needs_extra_links:
+                if extra_link["option"] in need.keys() and id in need[extra_link["option"]]:
+                    for n, link in enumerate(need[extra_link["option"]]):
+                        if id == link:
+                            need[extra_link["option"]][n] = "".join([id_prefix, id])
```
danieleades, line 169: i think this might be better done with a list comprehension. This is a very confusing bit of logic

sphinxcontrib/needs/api/need.py
danwos: thanks, changed it.

sphinxcontrib/needs/utils.py
danwos: Changed it, I hope I got the type annotation correct.

sphinxcontrib/needs/utils.py
danwos: done

sphinxcontrib/needs/utils.py
danwos: Maybe, but it works and is tested (copy & past from needimport code).
So I don't touch it with this PR.

sphinxcontrib/needs/api/need.py
```diff
@@ -371,6 +390,62 @@ def run():
...
+    kwargs["external_url"] = external_url
+    kwargs["external_css"] = external_css
+
+    return add_need(app=app, **kwargs)
+    # need_type=need_type, title=title, id=id,
+    # content=content, status=status, tags=tags, links_string=links_string,
+    # is_external=True, external_url=external_url, external_css=external_css)
+
+
 def _prepare_template(app, needs_info, template_key):
```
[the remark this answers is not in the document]
danwos: Yes, but this is not part of this PR :)

sphinxcontrib/needs/utils.py
danwos: done

sphinxcontrib/needs/api/need.py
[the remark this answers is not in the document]
danieleades: > Yes, but this is not part of this PR :)

this is true...

sphinxcontrib/needs/utils.py
danieleades: if it's copy-pasted, can it be extracted into a function (to avoid duplicating logic!)?

Issue #271, opened 2021-05-11
needtable: horizontal scrollbar


danwos:
If a needtable has many columns and data, it may be bigger has the "content" area.
So it has some overflows on the side.
![image](https://user-images.githubusercontent.com/998700/117874709-9c7f4f80-b2a1-11eb-9883-49979b1f6120.png)

Maybe some css could be added to activate horizontal scrollbars.

Based on discussion in #164

twodrops:
@danwos For me the tables appear even now as in the screenshot above. Any clue why?

Does it make sense to have an additional hosting of sphinx-needs with RTD Theme somewhere to easily check such layout issues with RTD? The default can still be Alabaster.

For example,
```https://sphinxcontrib-needs.readthedocs.io/rtd/en/latest/```

**Comment by twodrops** [comment]
Is this related to this Sphinx RTD problem? https://github.com/readthedocs/sphinx_rtd_theme/issues/117
Is your fix addressing needs tables or all rst tables?

**Comment by danwos** [comment]
Honestly I have checked it with our default theme (alabaster) only :(

I will check again and also work on a local solution to build and view our official docs with different theme in the browser.
Something like ``make theme-test`` and it opens 3 browser tabs with same docs but different config.

**Comment by danwos** [comment]
A quick check with  RTD theme and our docs shows me a good result:
![image](https://user-images.githubusercontent.com/998700/118938027-723b2b00-b94e-11eb-97a1-63e0179e47ee.png)

As you can see, there is no content overflow and a horizontal scrollbar.

Mhh, but I see other problems:
- The head columns don't have the same width as the content columns.
- When scrolling head and content is not align.
- There is some space before and after the content rows.

Okay the last point I got fixed.
But the alignment is another problems and shows up when using RTD theme only.

**Comment by danwos** [comment]
@twodrops: I should read your comment more carefully.
Missed the part that is not working for "normal" style only. Sorry.

For "normal" tables the solution should be quite easy css modification:
```
  table {
      display: block;
      overflow-x: auto;
      white-space: nowrap;
  }
```

But I'm concerned that this is not handled by RTD automatically, as Sphinx-Needs produces normal, docutil-based tables here.
so no additional voodoo. Same as defining tables in rst directly...

twodrops:
Sorry for my confusing comments :)

I meant two points.

[...]

PR #277, opened 2021-05-20
RTD Tables: Removes space around head and content

danwos:
Deals with #271


twodrops:
I tested the for style as "table"in this branch and the table still flows out of the margin. Could it be because the contents are longer?
2021-05-20_14h12_12

Issue #351, opened 2021-08-04
needimport path is not consistent to other directives


twodrops:
Other directives like `image `or `include `use relative path starting from rSt file. Looks like this is the standard.
`needimport` uses relative path starting from folder where `conf.py` is. This behavior is not consistent with the other directives.

Hint: Change to this behavior will break existing builds. Possibly it shall be changed asap or never.

danwos:
Good point.
Absolute paths are starting from ``conf.py`` folder for e.g. ``image`` directive.
Relative paths from rst file.

I'm totally open to change this behavior for ``needimport`` to the sphinx default way.
Will do it already with the upcoming release.

danwos:
This change will break the current implementation.

Maybe we can check if the "old" way is used to define the location (checking if file can be found) .
Then we throw an deprecation warning but go on without any error.

If old way does not lead to an existing file, use the new way (No warning).

Chriner:
Wouldn't it be better to use the new way first? 
- If file is found at new position, then OK. 
- If not found at new position, then use old way
  - if file at old position found throw an deprecation warning and go on
  - if no file found throw error

danwos:
@Chriner : You are right, that makes more sense.

For clarification:
``conf.py`` under ``/home/me/project/docs``.
Example rst under ``/home/me/project/docs/subfolder/page.rst``

```rst
.. needimport:: /my.json   
    --> absolute from conf.py because of starting /
   --> /home/me/project/docs/my.json

.. needimport:: my.json  
   --> relative from current rst-file because of missing /
   --> /home/me/project/docs/subfolder/my.json

   * if not found, check /home/me/project/docs/my.json
     * if found there, use it but throw  DEPRECATION warning
     * if not found, raise already implemented exception/warning
```

PR #356, opened 2021-08-19
Fixed needimport relative path consistency

haiyangToAI:
Fix for #351


docs/directives/needimport.rst
```diff
@@ -22,7 +22,7 @@ The directive **.. needimport::** can be used in all rst-documents. Simply write
       :post_template: post_template.rst
 
 The directive needs an absolute or relative path as argument.
-If the path is relative, an absolute path gets calculated with the folder of the **conf.py** as basedir.
+If the path is relative, an absolute path gets calculated like other directives, e.g. image, in Sphinx.
```
danwos, line 25: ... gets calculated based on the current document location.

sphinxcontrib/needs/directives/needimport.py
```diff
@@ -52,12 +52,30 @@ def run(self):
...
+            new_need_import_path = os.path.join(env.app.confdir, curr_dir, need_import_path)
+
+            correct_need_import_path = new_need_import_path
+            if not os.path.exists(new_need_import_path):
+                # Check the old way that calculates relative path starting from conf.py directory
+                old_need_import_path = os.path.join(env.app.confdir, need_import_path)
+                if os.path.exists(old_need_import_path):
+                    correct_need_import_path = old_need_import_path
+                    logger.warning(
+                        "Deprecated needimport relative path calculatation based on directory of conf.py used. "
```
danwos, line 66: Deprecation warning: Relative path must be relative to the current document in future, not to the conf.py location.
Use a starting "/", like "/needs.json", to make the path relative to conf.py.

tests/test_import.py
```diff
@@ -14,7 +14,7 @@ def test_import_json(app, status, warning):
     assert "new_tag" in html
 
     # Check filters
-    filter_html = Path(app.outdir, "filter.html").read_text()
+    filter_html = Path(app.outdir, "subdoc/filter.html").read_text()
     assert "TEST_01" not in filter_html
     assert "TEST_02" in filter_html
```
danwos, line 19: Missing some checks here, which checks if it the import has worked for relative and absolute path.

danwos (changes requested):

tests/test_import.py
danwos: Negative mssing, not existing file is detected.

danwos (changes requested):

tests/test_import.py
```diff
@@ -28,6 +28,38 @@ def test_import_json(app, status, warning):
...
+
+@with_app(buildername="html", srcdir="doc_test/non_exists_file_import")  # , warningiserror=True)
+def test_import_non_exists_json(app, status, warning):
+    # Check non exists file import
+    try:
+        app.build()
+    except ReferenceError as err:
+        assert (
+            err.args[0]
+            == "Could not load needs import file /home/haiyang/work/useblocks/github/sphinxcontrib-needs/tests/"
```
danwos, line 59: ``` /home/haiyang/work/useblocks/github/sphinxcontrib-needs/tests/```
This will not work on all machines ;)

Maybe you could check twice.
First for "Could not load needs import file", then for "doc_test/...".

danwos: I wonder, why this test is green on CI.
Please check, if the this re really working as expected. E.g. by changing the assert-check slightly.

tests/test_import.py
```diff
@@ -28,6 +28,38 @@ def test_import_json(app, status, warning):
...
+    rel_path_import_html = Path(app.outdir, "subdoc/rel_path_import.html").read_text()
+    assert "small_rel_path_TEST_01" in rel_path_import_html
+
+    # Check deprecated relative path import based on conf.py
+    deprec_rel_path_import_html = Path(app.outdir, "subdoc/deprecated_rel_path_import.html").read_text()
+    assert "small_depr_rel_path_TEST_01" in deprec_rel_path_import_html
+
+    warnings = warning.getvalue()
+    assert (
+        "WARNING: Deprecation warning: Relative path must be relative to the current document in future, "
```
danwos, line 45: No need to check for the complete warning.
" Deprecation warning:" should be enough.

So we can change the message without changing the tests

danwos (changes requested):

tests/test_import.py
danwos: Ohh no, seems the check have failed....

haiyangToAI: done.

tests/test_import.py
haiyangToAI: done.

danwos (approved):

Issue #348, opened 2021-08-03
feature request: caption for needtable


gregegg:
For `needflow `and several other figure directives, there exists the ability to define the figure caption in the directive, as shown in the example for [needflow](https://sphinxcontrib-needs.readthedocs.io/en/latest/directives/needflow.html).

```
.. needflow:: My first needflow
   :filter: is_need
```

Using the latex builder, the caption "My first needflow" shows up in the list of figures in the latex output when inserting the below directive: 

```
.. raw:: latex

   \listoftables
   \listoffigures
```

I request a new feature, where similar caption capability is added to the `needtable` directive.  Ideally, one would define the following, and have "my table caption" appear in the List of Tables, as created with the latex directive shown above. 

```
.. needtable::  my table caption
   :tags: main_example
   :style: table
```

danwos:
Thanks for sharing your idea.

I like the idea and the implementation should not take so much time.
But I will need some time as a lot of other stuff needs to be done as well.

gregegg:
Thank you!

danwos:
Hi @gregegg,

a PR from @haiyangToAI just got merged, which implements your feature request.
May I ask you to check, if it fulfills your needs?

Just install the current version from master and the feature is available.

gregegg:
Hi @danwos and @haiyangToAI ,

Thank you for including this feature request!  Unfortunately, the approach implemented doesn't work -- it results in a latex error that prevents the table from being created in the PDF.

I have uploaded a sample repository with test cases, and an illustration of how to make the table render correctly (basically, just copying the approach that sphinx uses to create latex tables).    

Please see https://github.com/gregegg/needtablecaption_348

danwos:
Fixed by #377

PR #366, opened 2021-08-27
Added caption for needtable

haiyangToAI:
#348

### DIFF
--- a/docs/directives/needtable.rst
+++ b/docs/directives/needtable.rst
@@ -9,14 +9,16 @@ needtable
 
 .. code-block:: rst
 
-   .. needtable::
+   .. needtable:: Example table
       :tags: main_example
       :style: table
 
-.. needtable::
+.. needtable:: Example table
    :tags: main_example
    :style: table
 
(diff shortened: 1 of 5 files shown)


docs/directives/needtable.rst
```diff
@@ -9,14 +9,16 @@ needtable
...
+   .. needtable:: Example table
       :tags: main_example
       :style: table
 
-.. needtable::
+.. needtable:: Example table
    :tags: main_example
    :style: table
 
+The argument of the ``needtable`` will be used as caption for the table. If argument is used, then
```
danwos, line 20: The argument of  ``needtable`` will be used as caption for the table. If argument is used, then 
the caption and table content will be wrapped together into a figure.

Have added/removed from above
```
- the
+ a
```

danwos (changes requested):
Great work.
Only 1 minor change in the docs, please :)

docs/directives/needtable.rst
haiyangToAI: done.

danwos (approved):

Issue #350, opened 2021-08-04
needextend does not add options as a list


twodrops:
When ```:+option:``` is used in ``needextend``, it does a string append and not a list append.

```
.. need::
:id: NEED_ID
:tag: tag1
```
```
.. needextend:: NEED_ID
   :+tag: tag2
```

**Resulting need**
```
.. need::
:tag: tag1 tag2
```

**Expected need**
```
.. need::
:tag: tag1, tag2
```

danwos:
I think this is the supposed behavior for extra_options, which contain internally a string.

Only tags, links and extra_links are handled as lists and therefor get a "," between the values.

I also see no way to change it, as a user maybe wants to connect two strings via a ``needextend`` update.

So you could add something like ``, tag2``, but for sure this does not work/look good, if the original need does not contain already a "tag 1"

danwos:
This may get possible with #378, which would introduce extra_option field types.
And then we can decide, if a comma is needed (list type) or  nor.

PR #369, opened 2021-09-06
Added test case for needextend option

haiyangToAI:
#350

### DIFF
--- a/tests/doc_test/doc_needextend/index.rst
+++ b/tests/doc_test/doc_needextend/index.rst
@@ -50,3 +50,10 @@ Need extend
 
 .. needextend:: extend_test_007
    :+links: extend_test_004, extend_test_005
+
+.. test:: needextend Example 8
+   :id: extend_test_008
+   :checks: extend_test_003
+
+.. needextend:: extend_test_008
+   :+checks: extend_test_004
--- a/tests/doc_test/doc_needextend/conf.py
+++ b/tests/doc_test/doc_needextend/conf.py
@@ -41,6 +41,8 @@ needs_types = [
     dict(directive="test", title="Test Case", prefix="TC_", color="#DCB239", style="node"),
 ]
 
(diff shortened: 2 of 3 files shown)


tests/doc_test/doc_needextend/index.rst
```diff
@@ -50,3 +50,10 @@ Need extend
 
 .. needextend:: extend_test_007
    :+links: extend_test_004, extend_test_005
+
+.. test:: needextend Example 8
+   :id: extend_test_008
+   :author: Foo_001
+
+.. needextend:: extend_test_008
+   :+author: Foo_002, Foo_003
```
danwos, line 59: The issue #350 is having its problems with "tag" option, which is a predefined field and stores the values in  a list.
I'm not sure if your solution works with tags.
Can you please update the test case and use "tag" as well?

sphinxcontrib/needs/directives/needextend.py
```diff
@@ -121,9 +121,9 @@ def process_needextend(app, doctree, fromdocname):
 
                         # else it must be a normal string
                         else:
-                            # If content is already stored, we need to add some whitespace
+                            # If content is already stored, we need to add some whitespace and comma
                             if need[option_name]:
-                                need[option_name] += " "
+                                need[option_name] += ", "
```
danwos, line 126: But what if I just want to add a string?
"," gets added now everywhere, for each field.

We only need it for comma-separated fields, which are tag and all links.
Everything else is a String-based field.

So this needs to work:
```

.. req:: My req
   :id: TEST_01
   :author: Me

.. needextend:: TEST_01
   +author: and some friends
```

So "author" should be "me and some friends", whithout any comma.

danwos (changes requested):

danwos:
On hold as long as discussion not finished in #350

Issue #317, opened 2021-07-05
New config: needs_filter_data  -> External values for filter strings


danwos:
Filter strings are not used only to filter existing data, but there are often used to decide, which needs shall somehow get manipulated, e.g. via `needextend`, `global_options` or also `needextract`.

Right now the filterable data is based on the internally collected need-data only.
It would be good to allow the injection of external data via configuration, so that a filter-string may behave different.

Example config:
```python
needs_filter_data = {
  "external_variant": "project_x",
  "env-example": os.env.get('ENV_VAR'),
  "another_var": a_custom_python_function(),
  "sphinx_tag": tags.has('my_tag') 
}
``` 
As you see the additional data can based on a bunch of external triggers(env vars, sphinx tags, files, ...) 

Hint: For Sphinx-Needs all values must be finally Strings. 
So a function like``a_custom_python_function()`` must get executed during config-read and need to return a string-object.
So this is **not allowed**:
```python
needs_filter_data = {
  "another_var": a_custom_python_function,   # () missing to execute the func
}
```

Then finaly used in rst like:
```rst
.. needextend:: type == "req" and external_variant == "project_x"
   :variant: project_x
   :customer: Mr X
   :secret: True 

.. needextend:: type == "req" and external_variant != "project_x"
   :variant: unknown
   :customer: unknown
   :secret: False

```

This can get very powerful for variant management, e.g. together with the planned `delete` option. See #316 .

PR #371, opened 2021-09-09
Added new config needs_filter_data

haiyangToAI:
#317


sphinxcontrib/needs/needs.py
```diff
@@ -514,6 +518,21 @@ def check_configuration(_app: Sphinx, config: Config):
     extra_options = config["needs_extra_options"]
     link_types = [x["option"] for x in config["needs_extra_links"]]
 
+    external_filter = config["needs_filter_data"]
```
danwos, line 521: what if "needs_filter_data" does not exist?
use get() to get values from dicts safely

sphinxcontrib/needs/needs.py
```diff
@@ -514,6 +518,21 @@ def check_configuration(_app: Sphinx, config: Config):
...
+    # Check if external filter values is really a string
+    for extern_filter, value in external_filter.items():
+        if not isinstance(value, str):
+            logger.warning(
+                "External filter value: {0} from needs_filter_data {1} is not a string.".format(value, external_filter)
+            )
+
+    # Check if needs external filter and extra option are using the same name
+    for ext_filter in external_filter:
```
danwos, line 530: can you add .keys() here.
I know Python 3.x send back keys for dicts automatically.
But the rest of the code is always using dict.keys() to handle keys

sphinxcontrib/needs/filter_common.py
```diff
@@ -58,6 +58,14 @@ def collect_filter_attributes(self):
         return collected_filter_options
 
 
+def load_needs_extra_filter_data(app, doctree, fromdocname):
+    """
+    Load needs external filter data from conf.py and set as global variable.
+    """
+    global NEEDS_EXTERNAL_FILTER_DATA
```
danwos, line 65: NEEDS_EXTERNAL_FILTER_DATA should be defined on module level, if global

sphinxcontrib/needs/filter_common.py
```diff
@@ -209,6 +218,11 @@ def filter_needs(needs, filter_string="", current_need=None):
 
     found_needs = []
 
+    # needs external filter data
+    if NEEDS_EXTERNAL_FILTER_DATA:
+        for extern_filter, value in NEEDS_EXTERNAL_FILTER_DATA.items():
+            filter_string = filter_string.replace(extern_filter, '"{}"'.format(value))
```
danwos, line 224: That kind of handling is not needed.

Just add NEEDS_EXTERNAL_FILTER_DATA to the context of the eval()-func

sphinxcontrib/needs/needs.py
```diff
@@ -89,7 +90,7 @@
 from sphinxcontrib.needs.roles.need_ref import NeedRef, process_need_ref
 from sphinxcontrib.needs.services.github import GithubService
 from sphinxcontrib.needs.services.manager import ServiceManager
-from sphinxcontrib.needs.utils import INTERNALS, NEEDS_FUNCTIONS
+from sphinxcontrib.needs.utils import INTERNALS, NEEDS_FUNCTIONS, logger
```
danwos, line 93: There gets already a logger imported in this file.
Use this one.

sphinxcontrib/needs/needs.py
```diff
@@ -281,6 +284,7 @@ def setup(app):
     # doctree-read. So manipulating the doctree may result in conflicts, as e.g. images get not
     # registered for sphinx. So some sphinx-internal tasks/functions may be called by hand again...
     # See also https://github.com/sphinx-doc/sphinx/issues/7054#issuecomment-578019701 for an example
+    app.connect("doctree-resolved", load_needs_extra_filter_data)
```
danwos, line 287: "doctree-resolved" gets executed per rst file.
So you load the data multiple times.

I would also use the function "load_config" to load the need_filter_data.
And then not by calling another function, but by simply:
```python

from sphinxcontrib/needs/filter_common.py  import  NEEDS_EXTERNAL_FILTER_DATA

def load_config():
...

global NEEDS_EXTERNAL_FILTER_DATA
NEEDS_EXTERNAL_FILTER_DATA = app.config['needs_filter_data']

```

danwos (changes requested):
I would do some parts a little different.
But I like the validation of the input data.

sphinxcontrib/needs/needs.py
haiyangToAI: Done. This won't work, adapted to the approach as discussed.

sphinxcontrib/needs/needs.py
haiyangToAI: Done.

sphinxcontrib/needs/filter_common.py
haiyangToAI: Done.

sphinxcontrib/needs/filter_common.py
haiyangToAI: Done.

sphinxcontrib/needs/needs.py
haiyangToAI: Done.

sphinxcontrib/needs/needs.py
haiyangToAI: It's fine here, config["needs_filter_data"] is a dictionary.

danwos: yes I know, but ``config.get('needs_filter_data', {})`` is more save, as it always set a value.
In your case ``config["needs_filter_data"]`` can be None, then ``external_filter`` is None and ``eval()`` will raise an exception.

I know, Sphinx should have set ``{}`` as default value for ``needs_filter_data``, but i would not trust this.

docs/configuration.rst
```diff
@@ -405,6 +405,61 @@ Use ``style_start`` and ``style_end`` like this::
       }
    ]
 
+.. _needs_filter_data:
+
+needs_filter_data
+~~~~~~~~~~~~~~~~~
+
+This option allows to custom define extra filter.
```
danwos, line 413: This option allows to use custom data inside a filter-string. (filter-string should be linked to the related chapter)

docs/configuration.rst
```diff
@@ -405,6 +405,61 @@ Use ``style_start`` and ``style_end`` like this::
...
+   needs_filter_data = {
+       "current_variant": "project_x",
+       "sphinx_tag": custom_defined_func(),
+   }
+
+
+The defined ``needs_filter_data`` must be a dictionary. Its values can be a string variable or a custom defined 
+function. The function get execued during config loading and must return a string. 
+
+Hence, this is **not allowed**::
```
danwos, line 429: Not needed to show, what is not allowed in docs.
Users are lazy and they may not read the first sentence and just copy your example.

danwos: So remove sentence and example

docs/configuration.rst
```diff
@@ -405,6 +405,61 @@ Use ``style_start`` and ``style_end`` like this::
...
+
+   needs_filter_data = {
+       "sphinx_tag": custom_defined_func,  # () missing to execute the function
+   }
+
+
+The value of ``needs_filter_data`` will be used as filter string and can be very poweful togehter with
```
danwos, line 436: No, the values of needs_filter_data will be available as data inside filter_strings.

They do not filter anything or even contain filters. Its just data for comparisons.

docs/configuration.rst
```diff
@@ -405,6 +405,61 @@ Use ``style_start`` and ``style_end`` like this::
...
+
+   needs_extra_options = {
+       "variant": directives.unchanged,
+   }
+
+
+the defined extra filter can also be used like::
+
+   .. needextend:: type == "test" and variant == current_variant
+      :variant: my_external_variant
```
danwos, line 455: Don't set variant here.
Better set ``:hide: True``

And for the filter ``variant != current_variant``

So all need objects get hidden, if the y are not defined for the current_variant

sphinxcontrib/needs/needs.py
```diff
@@ -514,6 +516,22 @@ def check_configuration(_app: Sphinx, config: Config):
...
+    # Check if external filter values is really a string
+    for extern_filter, value in external_filter.items():
+        if not isinstance(value, str):
+            log.warning(
+                "External filter value: {0} from needs_filter_data {1} is not a string.".format(value, external_filter)
+            )
+
+    # Check if needs external filter and extra option are using the same name
+    for ext_filter in external_filter.keys():
+        if ext_filter in extra_options:
```
danwos, line 530: I think this can be done in the first loop as well.
Saves some time

danwos (changes requested):
Overall good work.
Found 2-3 points.

But I don't like the test structure.
Needextend and needextract have nothing to do with needs_filter_data.
Sure they use it, but they should be tested on there own for their provided functions.
needs_filter_data is a common solution for all filter-related directives, so it should be tested independently.

Please remove the needs_filter_data from test for needextract and needextend.
Then create a new test-doc "doc_needs_filter_data". This documentation can make use of needextend, needextract, needfilter or whatever is needed to test the correct usage of needs_filter_data.

docs/configuration.rst
```diff
@@ -405,6 +405,54 @@ Use ``style_start`` and ``style_end`` like this::
...
+   needs_filter_data = {
+       "current_variant": "project_x",
+       "sphinx_tag": custom_defined_func(),
+   }
+
+
+The defined ``needs_filter_data`` must be a dictionary. Its values can be a string variable or a custom defined 
+function. The function get execued during config loading and must return a string.
+
+The value of ``needs_filter_data`` will be available as data inside :ref:`filter_string` and can be very poweful togehter with
```
danwos, line 429: typo: together

docs/configuration.rst
```diff
@@ -405,6 +405,54 @@ Use ``style_start`` and ``style_end`` like this::
...
+   }
+
+
+The defined ``needs_filter_data`` must be a dictionary. Its values can be a string variable or a custom defined 
+function. The function get execued during config loading and must return a string.
+
+The value of ``needs_filter_data`` will be available as data inside :ref:`filter_string` and can be very poweful togehter with 
+internal needs info to filter needs.
+
+The defined extra filter can be used like this::
```
danwos, line 432: "the defined extra filter data...."
It's not a filter :)

docs/configuration.rst
```diff
@@ -405,6 +405,54 @@ Use ``style_start`` and ``style_end`` like this::
...
+
+   .. needextend:: type == "req" and sphinx_tag in tags
+      :+tags: my_external_tag
+
+
+or if need has :ref:`needs_extra_options` defined like::
```
danwos, line 438: "or if project has :ref..."

docs/configuration.rst
```diff
@@ -405,6 +405,54 @@ Use ``style_start`` and ``style_end`` like this::
...
+
+   needs_extra_options = {
+       "variant": directives.unchanged,
+   }
+
+
+the defined extra filter can also be used like::
```
danwos, line 445: ...extra filter DATA

sphinxcontrib/needs/needs.py
```diff
@@ -514,6 +516,20 @@ def check_configuration(_app: Sphinx, config: Config):
     extra_options = config["needs_extra_options"]
     link_types = [x["option"] for x in config["needs_extra_links"]]
 
+    log = get_logger(__name__)
+    external_filter = getattr(config, "needs_filter_data", {})
+    for extern_filter, value in external_filter.items():
+        # Check if external filter values is really a string
+        if not isinstance(value, str):
+            log.warning(
```
danwos, line 524: I think this should raise an exception as well.
Not having string-values in our data may be a huge problem, as our code does not check this everywhere.

danwos (commented):
Thanks for the changes.
Found 1-2 small new points.

danwos (approved):

PR #377, opened 2021-09-23
Fix needtable caption

haiyangToAI:
Fix for #348


sphinxcontrib/needs/directives/needtable.py
```diff
@@ -64,16 +64,20 @@ def run(self):
 
         sort = self.options.get("sort", "id_complete")
 
-        caption = None
         if self.arguments:
-            caption = self.arguments[0]
+            title_text = self.arguments[0]
+            text_nodes = self.state.inline_text(title_text, self.lineno)[0]
+            title = nodes.title(title_text, "", *text_nodes)
```
danwos, line 70: I would try to keep the creating of nodes out of run() and only collect data here.

The creation then is done in process().
So need_all_needtables mostly contain only str and not more

danwos: So please move the node creation to process()

tests/test_needtable.py
```diff
@@ -16,27 +16,28 @@ def test_doc_build_html(app, status, warning):
...
-# """
-#     assert figure in html
-#
-#     # negative test to check table without caption
-#     if sphinx.version_info[0] >= 4:
-#         assert '<table class="NEEDS_TABLE rtd-exclude-wy-table docutils align-default" id="needtable-index-1">'
-#         in html
-#     else:
-#         assert '<table class="NEEDS_TABLE rtd-exclude-wy-table docutils" id="needtable-index-1">' in html
+@with_app(buildername="xml", srcdir="doc_test/doc_needtable")
```
danwos, line 19: you can/should keep html here.
But use ElementTree to analyse it.

danwos: Or maybe leave it in the existing test-case (as it was before), but us elementree-checks instead of string-compares.

danwos (changes requested):

sphinxcontrib/needs/directives/needtable.py
haiyangToAI: Done.

tests/test_needtable.py
haiyangToAI: Done.

pyproject.toml
```diff
@@ -46,6 +46,7 @@ requests = "^2.25.1"
 [tool.poetry.dev-dependencies]
 
 # test dependencies
+lxml = ">=4.6.3"
```
danwos, line 49: Please use version selection as described here:
https://python-poetry.org/docs/dependency-specification/

In this case ``^4.6.3`` would be good

danwos (changes requested):

pyproject.toml
haiyangToAI: done

danwos (approved):

PR #414, opened 2021-11-25
add 'flake8-comprehensions'

danieleades:


sphinxcontrib/needs/directives/needfilter.py
```diff
@@ -114,7 +114,7 @@ def process_needfilters(app, doctree, fromdocname):
                 continue
 
             plantuml_block_text = ".. plantuml::\n" "\n" "   @startuml" "   @enduml"
-            puml_node = plantuml(plantuml_block_text, **dict())
+            puml_node = plantuml(plantuml_block_text, **{})
```
danieleades, line 117: what is the `**dict()` arg doing anyway? can this just be removed entirely?

danieleades: i've removed them. let's see if the CI is happy

danieleades:
hmm. unsure if this CI error is related to this PR. Would you kindly retrigger the github action @danwos?

sphinxcontrib/needs/directives/needfilter.py
danwos: not sure,  but often a node gets 2 arguments. So I think the plantuml-node needs it.

danwos:
Tests retriggered

sphinxcontrib/needs/directives/needfilter.py
danieleades: i'm just trying to figure out what this expands to. the `**` operator turns a dictionary into keyword arguments, but since it's an empty dictionary, surely that results in 0 keyword arguments, which is equivalent to calling the method with only a single arg.

right?

i've not seen this construction before, so it's got me second-guessing myself

danwos: ohh yes, for sure you are right. So it makes no sense.

danieleades: the 'linkcheck' job has passed this time, so it appears to have been some transient error

PR #415, opened 2021-11-25
migrate to pre-commit for linting+formatting

danieleades:
using pre-commit has some advantages

- you *can't* commit code to the repo doesn't meet your quality gates (unless you use git's `--no-verify` flag)
- there are numerous hooks available for improving quality
- dependencies are installed in an isolated environment

that last point is especially important for this project, as its dependency resolution seems quite brittle


danwos:
Thanks for the PR.
As I don't  have any experiences with pre-commit hook, I will test it a little bit today.
But I like the idea and the PR looks valid for me.

PR #418, opened 2021-11-25
lint and fix using pyupgrade

danieleades:
stacked on #415 

`pyupgrade` is a hook for automatically identifying and fixing outdated syntax


danwos:
Ohh wow, that's an amazing tool. Glad to have it now :) :+1:

PR #419, opened 2021-11-26
remove lockfile from VCS

danieleades:
I can't remember if i've raised this before or not...

Does the lockfile need to be committed to VCS? usual practice is to *not* include the lockfile in VCS, since for a library it is ignored anyway when imported as a dependency. Thus testing without the lockfile better matches the deployment environment

this will require reworking the CI.

This has the additional benefit of making dependabot a lot quieter, since it will only create PRs that update `pyproject.toml`, and not `poetry.lock`


danwos:
Ci does not work without ``poetry.lock``.
I don't know why, but I have currently no time to investigate this topic any further.

I close the PR to keep the PR list clean.

PR #423, opened 2021-11-26
Adds colwidths option for needtable

danwos:


sphinxcontrib/needs/directives/needtable.py
```diff
@@ -58,6 +60,15 @@ def run(self):
...
 
+        colwidths = str(self.options.get("colwidths", ""))
+        if colwidths:
+            colwidths = [int(width.strip()) for width in re.split(";|,", colwidths)]
+            if len(columns) != len(colwidths):
+                raise NeedsInvalidException(
+                    f"Amount of elements in colwidths and columns do not match: "
+                    f"colwidths: {len(colwidths)} and columns: {len(columns)}"
+                )
+
```
danieleades, line 71: can you do this without shadowing the `colwidths` variable with multiple types?

sphinxcontrib/needs/directives/needtable.py
```diff
@@ -154,13 +166,19 @@ def process_needtables(app, doctree, fromdocname):
...
+            if colwidths:  # Get values from given colwidths option
+                tgroup += nodes.colspec(colwidth=int(colwidths[colwidths_counter]))
+            elif option == "TITLE":  # if nothing in colwidths...
                 tgroup += nodes.colspec(colwidth=15)
             else:
                 tgroup += nodes.colspec(colwidth=5)
 
+            colwidths_counter += 1
+
```
danieleades, line 181: `option` and `title` don't appear to be used in the loop. why not iterate over the colwidths directly?

danieleades (changes requested):

sphinxcontrib/needs/directives/needtable.py
```diff
@@ -154,13 +166,19 @@ def process_needtables(app, doctree, fromdocname):
...
 
         # Define Table column width
-        # ToDo: Find a way to chosen to perfect width automatically.
+        colwidths_counter = 0
+        colwidths = current_needtable["colwidths"]
         for option, title in current_needtable["columns"]:
-            if option == "TITLE":
+
+            if colwidths:  # Get values from given colwidths option
+                tgroup += nodes.colspec(colwidth=int(colwidths[colwidths_counter]))
```
danieleades, line 174: you already coerced the elements of `colwidths` to `int` above (line 65)

sphinxcontrib/needs/directives/needtable.py
danwos: Done

sphinxcontrib/needs/directives/needtable.py
danwos: Option is used in the if-clause.
But yes, this could overall be restructured. 
But it's working, so I keep it :)

sphinxcontrib/needs/directives/needtable.py
danwos: You are right, removed.

sphinxcontrib/needs/directives/needtable.py
```diff
@@ -154,9 +167,13 @@ def process_needtables(app, doctree, fromdocname):
         tgroup = nodes.tgroup()
 
         # Define Table column width
-        # ToDo: Find a way to chosen to perfect width automatically.
-        for option, _title in current_needtable["columns"]:
-            if option == "TITLE":
+        colwidths = current_needtable["colwidths"]
+        for index, value in enumerate(current_needtable["columns"]):
+            option, _title = value
```
danieleades, line 172: does this work?

```suggestion
        colwidths = current_needtable["colwidths"]
        for index, option in enumerate(current_needtable["columns"].keys()):
```

danwos: Just tested and no, as it is a list and does not know anything about ".keys()"

PR #424, opened 2021-11-26
add dependabot config

danieleades:
this config will automatically create pull requests to update major versions in github actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


danwos:
Nice idea. Lets see how "annoying" it can be, as I have read no so good feedback about it.
It may create a lot of noise. But lets give it a try.

Thanks again for all of your PRs. They really enhance the project infrastructure a lot.

danieleades:
> Nice idea. Lets see how "annoying" it can be, as I have read no so good feedback about it. It may create a lot of noise. But lets give it a try.

yeah i'm not sure about this. i think there's potential for it to create a lot of noise since the lockfile is committed. If the lockfile was not committed, then it would only create PRs for major changes. So there are two options in that case-

1. remove dependabot. or clamp its updates to 'monthly' or something
2. rework the CI to not rely on the `poetry.lock` file, and then remove it from VCS

PR #437, opened 2021-12-03
Fixed dependency for needflow

haiyangToAI:
* Added sphinxcontrib-plantuml back, needed by needflow
* Updated poetry.lock

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,74 +1,71 @@
 [[package]]
-category = "main"
-description = "A configurable sidebar-enabled Sphinx theme"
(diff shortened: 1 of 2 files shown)


danwos:
Up to now installing sphinxcontrib-plantuml was always a manual task.
See https://sphinxcontrib-needs.readthedocs.io/en/latest/installation.html#id1

Mostly because java and the plantuml.jar must be provided/installed by the user itself.

Are there any error in the build itself, if this dependency is missing.
I think only in the docs is a "red box", if sphinxcontrib-plantuml is missing

haiyangToAI:
ah okay, that's why. Thanks for the explaination. I will close this PR.

Issue #450, opened 2021-12-14
needpie: figures are not closed


PhilipPartsch:
In needpie the created figures are not closed. This leads to the warning on Windows 10:
`RuntimeWarning from matplotlib: More than 20 figures have been opened. Figures created through the pyplot interface ('matplotlib.pyplot.figure') are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam 'figure.max_open_warning')`

PhilipPartsch:
I reset the matplotlib style to the values of the start of the script. "This is nice."

danwos:
Never have seen this warning before, but I also never had any use cases with more than 20 figure / needpies.

Great finding and thanks for the fix. :+1:

PR #451, opened 2021-12-14
close the figures, to free memory

PhilipPartsch:
fix issue #450


danwos:
ahh some linter problems.
Do you have pre-commit hooks installed?
https://sphinxcontrib-needs.readthedocs.io/en/latest/contributing.html#installing-dependencies

This will check linter and other stuff, before a final commits get created.
It will also fix the code in most cases, if something got found.

As alternaitve, you can run ``make format`` and ``make lint`` by hand.
``make format`` runs **black**, so that most linter probs get fixed.
``make lint`` checks this.

The same  command will also be executed on CI. So local finding will always be found by the CI itself :)

Issue #452, opened 2021-12-14
Request for needbar


PhilipPartsch:
I want to use bar charts to visualize my needs:
- It would be helpful, to have them visualized as bars (vertical or horizontal).
- It shall be possible to show the bars independent or stacked. 
- The data input shall be given as tables with configurable labels for x/y axis and legend.
- Output shall be rendered diagrams included in html.
- Colouring shall be customisable.
- Labels for the rendered values shall be shown. 

For convenient:
- the separator in the tables shall be configurable (e.g. ',' is even be used in a filter rule)
- It shall be possible to transpose the input table, so big tables does not have to be refactored manual, only to change x-y axis.

PhilipPartsch:
The issue #413 is even valied for needbar.

PR #453, opened 2021-12-14
inital version of needbar

PhilipPartsch:
Here is the implementation of needbar requested with #452


danwos:
Wow, really great work and already awesome docs.
I just was able to build the docs locally (without any problems) and it looks just wow.

Will start with a review soon

docs/directives/needbar.rst
```diff
@@ -0,0 +1,599 @@
...
+
+                   ,                           open ,                          in progress ,                          closed ,                          done ,                          implemented , number
+        Requirement,  type=='req' and status=='open', type=='req' and status=='in progress', type=='req' and status=='closed', type=='req' and status=='done', type=='req' and status=='implemented', 5
+               Test, type=='test' and status=='open',type=='test' and status=='in progress',type=='test' and status=='closed',type=='test' and status=='done',type=='test' and status=='implemented', 7
+      Specification, type=='spec' and status=='open',type=='spec' and status=='in progress',type=='spec' and status=='closed',type=='spec' and status=='done',type=='spec' and status=='implemented', 9
+
+.. needbar:: A more real bar chart
+   :legend:
+
+                ,                           open ,                          in progress ,                          closed ,                          done ,                          implemented , number
```
danwos, line 44: Must the first line always contain the x-axis labels.

There is also ``:xlabels:``, can both be used?

I think a small paragraph about how the content must be designed is missing.

danwos: I'm also have the feeling that we now have two ways of defining the labels.
In content or via option.

I would like to have only one, to make it a little bit easier or more clear for the user.
And then I prefer to have ``xlabels`` only, as this is already the way how other directives are asking for "column" names.
But that is just my thought, any opinions are welcome.

danwos (commented):
Have to stop the review for the moment and the code is still waiting for it.
But overall, a really good documentation and already really nice features and configuration options. :+1: 
Looking forward to have it in 0.7.5 :)

docs/directives/needbar.rst
```diff
@@ -0,0 +1,599 @@
...
+
+.. code-block:: rst
+
+   .. needbar:: A more real bar chart
+      :legend:
+
+                   ,                           open ,                          in progress ,                          closed ,                          done ,                          implemented , number
+        Requirement,  type=='req' and status=='open', type=='req' and status=='in progress', type=='req' and status=='closed', type=='req' and status=='done', type=='req' and status=='implemented', 5
+               Test, type=='test' and status=='open',type=='test' and status=='in progress',type=='test' and status=='closed',type=='test' and status=='done',type=='test' and status=='implemented', 7
+      Specification, type=='spec' and status=='open',type=='spec' and status=='in progress',type=='spec' and status=='closed',type=='spec' and status=='done',type=='spec' and status=='implemented', 9
```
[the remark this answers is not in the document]
PhilipPartsch: No, the white spaces are not needed. But I want to show, that it is possible to format the tables easily readable.

docs/directives/needbar.rst
PhilipPartsch: Yes, you are right, both options are here. 
Reason: In the future I want to generate the complete content table during runtime with python scripting or reuse of needlist (or other new need option).
To do so, it is possible to define the labels in content or with an extra option like in the other needs. 
Maybe it is not clearly described in the description, but we need to have xlabel / ylabel. ylabel is been used as values for the legend. Even more if you are using transpose, they change the meaning. So please read x & y as columns and rows labels in the defined content data table. Maybe I have to rename them accordingly? Any opinions on?

PhilipPartsch: Your are right, I shall provide a chapter on the content itself and how it is been interpreted.

danwos: Ok, I see here 3 scenarios, which we need to support:

1) Data with no labels for x/y and no option
2) Data with labels  in 1. row/column
3) Data with no labels, but ``x/ylabels`` option

We should support case 1), as this is the easiest way for the user to get something done with needbar.
Adding labels is for me the next step.
So we need a flag or another way to indicate, that needbar shall extract the labels from the data.

As we already have ``xlabels``, we can maybe use a magic word for this, e.g. ``FROM_DATA``.
So if ``xlabels`` is not set, not labels will be set.
``xlabels == FROM_DATA`` takes first row for labels.
``xlabels set, but != FROM_DATA``, use the value as comma separated list for the labels.

Does this makes sense for you?

docs/directives/needbar.rst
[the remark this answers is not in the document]
danwos: Maybe in this case we should show 2 examples.
First without any whitespace and then the second with this nice feature.

docs/directives/needbar.rst
PhilipPartsch: I'm currently thinking about _1. Data with no labels for x/y and no option_.
Do you like this rendering?
need_bar_5e2f9

danwos: Yes, it's fine.
But I think the argument of ``.. needbar::`` should be also allowed to be empty, so that "Test empty label" is not shown.

I know that these (title, labels) are missing information in a good documentation, but we never know how a user wants to use it and what the first steps are to get familiar with this function. So asking for as less information as possible is helpful.

PhilipPartsch: If you give no title to the current implementation, you do not get any in the picture, so it works already fine. I will add a chapter title to the documentation, too.

PhilipPartsch:
I want to bring an older question back again:
So please read x & y as columns and rows labels in the defined content data table. Maybe I have to rename them accordingly? What is your opinion?

I'm currently even not sure what is the best name for them, as I'm not an English native speaker. 

Shall I rename xlabel / ylabel to rowlabel / columnlabel? 
Is it so more understandable that we name the elements in the table and not the elements in the diagram?

danwos:
I'm fine with xlabel and ylabel.
Also matplotlib seems to go in this direction.

Row/column makes sense for the data table in the directive content.
But for the diagram itself it does not make much sense and X/ylabel is fine.

Also naming is not soo important, as long as the docs and examples are good. What is already the case.

docs/directives/needbar.rst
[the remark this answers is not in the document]
PhilipPartsch: I added many more examples even for content.

danwos:
Thanks for all the work.
This is really great work and a big, awesome feature for Sphinx-Needs.

PR #460, opened 2022-01-07
fix: 459 - issue when loading needs from URL

David-Le-Nir:
This PR fixes a typo in the code that made calling needs.json URL ending in a failure.

Note that no test have been modified because there is not any for the remote URL testing. And I'm not sure about the way you prefer to mock the http call to get a json file.


danwos:
Looks good for me.
Feel free to add your name also to the AUTHORS file, if you like.

Lets wait what the CI is saying...

David-Le-Nir:
I have one question for test_doc_build_html test because it seems that using subprocess.run(["sphinx-build" ... bypass the mock set with requests_mock.
So the test fails. 
I can either:

- separate my changes in a dedicated test case (i.e. do not change the conf.py to use remote needs in test_doc_build_html)
- change code to not crash if the URL cannot be reached
- do any of your suggestion

what do you think ?

danwos:
I think a separate test case makes sense, which gets build with the @app directive so that mocking should work.
And I think you can simply reuse the same sphinx-test-project. So it gets build twice. Once for checking the console output via subprocess, and once for mocking and checking the build results (no console check possible)

David-Le-Nir:
I think I have to use another test project as I need a conf.py in which I do not use json_url for external needs to keep the test_doc_build_html running.
But maybe I don't understand.
What I think I have to do:

- left the existing tests unchanged
- create a new test, specific for testing remote external needs, with a dedicated test project (conf.py) and requests_mock usage

Do we agree ?

danwos:
I have not fully understood, why an additional test project is needed.
But I also see no drawbacks to have one, so that the test is really focused.
So just go with an additional project and use requests_mock.

And you need to rebase your branch to the latest master branch, as I have already solved some other issues.

David-Le-Nir (commented):
Seems good to me.

danwos:
Thanks again for the PR! :+1:

Issue #458, opened 2022-01-06
Remove logging that cause an error on attribute that should not be defined


elnino1:
When using needs_external_needs, the json_path attribute shall not be defined.
However we have to define an empty value else an error (KeyError: 'json_path') occurs on line 
https://github.com/useblocks/sphinxcontrib-needs/blob/master/sphinxcontrib/needs/external_needs.py#L34

needs_external_needs = [
  {
      "base_url": "http://localhost",
      "json_url": 'http://localhost/needs.json',
      # if not defined, an error occurs
      "json_path": "",
      "version": "0.1.0"
  }
]

danwos:
Thanks for reporting and for the detail report.
I will try to fix it with the next release.

David-Le-Nir:
If you want I'm working on a PR for that som you may want to assign me.
(my implementation is just to remove the line 34 highlighted in description)
I just wait for the 459 to be fixed then I'll rebase to get the unit tests part that I'm missing.

danwos:
Ohh that would be nice.
Thanks for the commitment.

PR #465, opened 2022-01-11
Fix 458

David-Le-Nir:
This PR fix an issue happening when using _json_url_ for external needs access.
In this case, some logging is called on _json_path_ value which is not required (thus may not be defined) causing a failure (KeyError).

### DIFF
--- a/tests/test_needs_external_needs_build.py
+++ b/tests/test_needs_external_needs_build.py
@@ -12,7 +12,7 @@ def test_doc_build_html(app, status, warning):
     src_dir = "doc_test/doc_needs_external_needs"
     out_dir = Path(app.outdir)
     output = subprocess.run(["sphinx-build", "-M", "html", src_dir, out_dir], capture_output=True)
-    assert not output.stderr
+    assert not output.stderr, f"Should not have stderr: {output.stderr}"
 
     # run second time and check
     output_second = subprocess.run(["sphinx-build", "-M", "html", src_dir, out_dir], capture_output=True)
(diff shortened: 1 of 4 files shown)


danwos:
Maybe for the future: You should also add ``Fixes #458`` to a commit, so that the PR gets automatically linked to the issue.
And the issue gets automatically closed, when the PR gets merged.
Saves some clicks :)

Details: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

David-Le-Nir (commented):
Looks good to me.

tests/test_needs_external_needs_build.py
```diff
@@ -12,7 +12,7 @@ def test_doc_build_html(app, status, warning):
     src_dir = "doc_test/doc_needs_external_needs"
     out_dir = Path(app.outdir)
     output = subprocess.run(["sphinx-build", "-M", "html", src_dir, out_dir], capture_output=True)
-    assert not output.stderr
+    assert not output.stderr, f"Should not have stderr: {output.stderr}"
```
danwos, line 15: What does this part do: ``f"Should not have stderr: {output.stderr}"``?

Gets ``Should not have stderr:`` written by Sphinx?

Or is it a comment? Then use ``#``.

danwos (changes requested):

tests/test_needs_external_needs_build.py
David-Le-Nir: where is is something in the stderr, assertion fails. If so it displays the given assertion error message with the content of the stderr.
This helped me to understand problem in unit test.
I forgot to remove it when pushed. I find it makes sense to display the stderr if we are expecting to not have any.
If you want I remove this.

danwos: Ohh cool, learned something new.
Didn't know that this gets used as output for a failed assertion.

So no, it can stay :)

danwos (approved):

PR #468, opened 2022-01-12
Add need value

PhilipPartsch:
I liked the idea of @Jattin in #439 
So I created this draft of need_value.
Please give it a try.


sphinxcontrib/needs/roles/need_value.py
```diff
@@ -0,0 +1,110 @@
...
+from sphinxcontrib.needs.logging import get_logger
+
+log = get_logger(__name__)
+
+
+class NeedValue(nodes.Inline, nodes.Element):
+    pass
+
+
+def process_need_value(app, doctree, fromdocname):
```
PhilipPartsch, line 18: @danwos I reused the implementation of the role need_ref. Could we maybe define a more universal function to handle this. Even it would be nice to have in need_ref access to self define options. This is here implemented.

danwos:
I like the idea and the implementation.

But I think the name ``need_value`` may confuse a little bit, as it doesn't give access to a free selectable value of a need.
So you can't use ``:need_value:`value_bread_req_1` ``  twice for different values of the same need.

It is more feeling like a fix , textual representation of the need.
So something like ``__repr__`` or better ``__str__``[ (docs)](https://docs.python.org/3/reference/datamodel.html#object.__str__) of a python class.

My suggestion is to rename ``need_value`` to ``need_repr``.
What do you think?


But a ``need_value`` feature is still be needed ;)
I'm thinking about extending the ``need`` role to something like 
`` The bread is :need:`value_bread_req_1:value` :need:`value_bread_req_1:unit` long ``
But lets discuss this in #439 .

Overall, thanks for your great PR :+1: .

danwos:
@PhilipPartsch: Any thoughts on my review?

PhilipPartsch:
> @PhilipPartsch: Any thoughts on my review?

I like your input. I'm currently implementing a solution based on role need_ref. Please reject the pull request. 

One question: could you have a look why the build is read. My implementation is not touching any work items, so I confused what is the issue with my source code.

danwos:
@PhilipPartsch : I also have no idea, why the build is red.
It claims about a duplicated need object (same id), but your code does nothing in this direction.

One idea: Maybe the id/ref calculation for docutils nodes has not thrown out unique ids for ``need_value``.
And that the shown error is somehow misleading.  But code looks good for me.

I retriggered the CI, lets see if the problem is reproducible.

danwos:
Closed by wish from @PhilipPartsch

Issue #400, opened 2021-10-29
filter code support for needpie


danwos:
Currently only ``needtable`` and ``needflow`` support "filter code" in the content area, which returns the needed filter result for the related representation.

The same feature shall be available for ``needpie``.

As ``needpie`` already needs the content for the definition of multiple filter strings, a new option ``filter_content`` shall be introduced.

The code in the content area must return a list of filter-result. One entry per piece of cake

PR #472, opened 2022-01-18
Support filter_func for  needpie

haiyangToAI:
#400

### DIFF
--- a/sphinxcontrib/needs/directives/needpie.py
+++ b/sphinxcontrib/needs/directives/needpie.py
@@ -1,3 +1,4 @@
+import copy
 import os
 
 import matplotlib
@@ -13,6 +14,7 @@ import matplotlib.pyplot
 from docutils.parsers.rst import directives
 
 from sphinxcontrib.needs.logging import get_logger
+from sphinxcontrib.needs.utils import check_and_get_external_filter_func
 
 logger = get_logger(__name__)
 
@@ -43,6 +45,9 @@ class NeedpieDirective(FilterBase):
         "text_color": directives.unchanged_required,
         "shadow": directives.flag,
     }
+
+    # Update the options_spec with values defined in the FilterBase class
+    option_spec.update(FilterBase.base_option_spec)
 
     def run(self):
         env = self.state.document.settings.env
@@ -94,6 +99,7 @@ class NeedpieDirective(FilterBase):
             "shadow": shadow,
             "text_color": text_color,
         }
+        env.need_all_needpie[targetid].update(self.collect_filter_attributes())
 
         return [targetnode] + [Needpie("")]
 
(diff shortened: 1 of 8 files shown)


danwos (changes requested):

tests/test_needs_filter_data.py
```diff
@@ -67,3 +67,6 @@ def test_doc_needs_filter_code(app, status, warning):
 
     # code func data
     assert "extern_filter_test_003" in code_html
+
+    # check needpie filter func code data
+    assert '<img alt="_images/need_pie_' in code_html
```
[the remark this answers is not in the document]
haiyangToAI: I couldn't find a way to modify the 'alt' tag in needpie for now.

sphinxcontrib/needs/directives/needpie.py
```diff
@@ -156,10 +154,16 @@ def process_needpie(app, doctree, fromdocname):
                 if filter_func:
                     filter_func(**context)
                 sizes = context["results"]
+                # check items in sizes
+                for item in sizes:
+                    if not isinstance(item, int) and not isinstance(item, float):
+                        logger.error(f"{item} in sizes {sizes} is not a integer or float number.")
```
danwos, line 160: "sizes" is an internal name from the code, the user does not know it.
So the error message is not really helpful.
Better : The returned values from the given filter_func is not valid. It must be a list with items of type int/float.

sphinxcontrib/needs/directives/needpie.py
```diff
@@ -156,10 +154,16 @@ def process_needpie(app, doctree, fromdocname):
                 if filter_func:
                     filter_func(**context)
                 sizes = context["results"]
+                # check items in sizes
+                for item in sizes:
```
danwos, line 158: Check also if "sizes" is really a list or something else

sphinxcontrib/needs/directives/needpie.py
```diff
@@ -156,10 +154,16 @@ def process_needpie(app, doctree, fromdocname):
...
+                    if not isinstance(item, int) and not isinstance(item, float):
+                        logger.error(f"{item} in sizes {sizes} is not a integer or float number.")
             except Exception as e:
                 raise e
-        if current_needpie["filter_func"] and content:
+        elif current_needpie["filter_func"] and content:
             logger.error("filter_func and content can't be used at the same time for needpie.")
+        else:
+            logger.error("Both filter_func and content not used for needpie.")
```
danwos, line 166: I like the old message more:
"filter_func and content can't be used at the same time for needpie.""

danwos (changes requested):

sphinxcontrib/needs/directives/needpie.py
haiyangToAI: The old message is still there. This one is for the situation, where no content and no filter_func used at all.

danwos: Ahh okay, was too stupid to read the diff the right way. all good.

sphinxcontrib/needs/directives/needpie.py
```diff
@@ -155,15 +155,23 @@ def process_needpie(app, doctree, fromdocname):
                     filter_func(**context)
                 sizes = context["results"]
                 # check items in sizes
+                if not isinstance(sizes, list):
+                    logger.error(
+                        f"The returned values {sizes} from the given filter_func {filter_func} is not valid."
```
danwos, line 160: You should not print the content of "sizes" here.
You don't know how big it is and it may make the log message unreadable.

sphinxcontrib/needs/directives/needpie.py
```diff
@@ -155,15 +155,23 @@ def process_needpie(app, doctree, fromdocname):
...
+                        " It must be a list."
+                    )
                 for item in sizes:
                     if not isinstance(item, int) and not isinstance(item, float):
-                        logger.error(f"{item} in sizes {sizes} is not a integer or float number.")
+                        logger.error(
+                            f"The returned values {sizes} from the given filter_func is not valid. "
```
danwos, line 166: same here:  no return values in log messages.

danwos (changes requested):

sphinxcontrib/needs/directives/needpie.py
haiyangToAI: ah you are right, will fix.

sphinxcontrib/needs/directives/needpie.py
```diff
@@ -157,13 +157,13 @@ def process_needpie(app, doctree, fromdocname):
...
                     logger.error(
-                        f"The returned values {sizes} from the given filter_func {filter_func} is not valid."
+                        f"The returned values from the given filter_func {filter_func} is not valid."
                         " It must be a list."
                     )
                 for item in sizes:
                     if not isinstance(item, int) and not isinstance(item, float):
                         logger.error(
-                            f"The returned values {sizes} from the given filter_func is not valid. "
+                            f"The returned values from the given filter_func {filter_func} is not valid. "
```
danwos, line 166: I'm not sure if ``{filter_func}`` is useful in log message, as it is a function.
Have you tested this and is the output okay?

Maybe ``filter_func.__name__`` is better?

danwos (changes requested):

danwos (approved):

Issue #471, opened 2022-01-17
Migrate Tests to pytest


danwos:
All tests shall be migrated to use pytest instead of nose.
Reasons for this are explained in #467.

Right now, our tests are using the external ``sphinx-testing`` package: https://github.com/sphinx-doc/sphinx-testing
But this package is deprecated and nose itself also.

So time for a change and a more modern testing framework.

### Technical proposal
We should use the sphinx internal package ``sphinx.testing``, which is not well documented. But a little bit here:
https://www.sphinx-doc.org/en/master/internals/contributing.html?highlight=testing#unit-tests

Also **Sphinx-Needs Enterprise** is using it and has one working test case:
https://github.com/useblocks/sphinx-needs-enterprise/tree/main/tests

However, I'm not 100% satisfied how the fixture gets configured and what parameters are available.
So there may be some reasons to create our own sphinx-fixture (based on sphinx.testing).

But lets see and we start by trying to get everything done with using sphinx.testing only.

### Tasks
* Get an ``app`` fixture
*  Set the ``rootdir/`` doc test project`` path
* Try to keep our file and folder structure
* Find solution for the used ``warnings`` return value of ``sphinx-testing`` in some test cases.
* Define a "tag/marker" to mark test cases, which shall or shall not be used for y pytest run. Makes it easier to run pytest for selected test-cases only during migration.

PR #484, opened 2022-01-26
Example migration to pytest

haiyangToAI:
#471


tests/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def clean_up_tmpdir(targetdir):
+    shutil.rmtree(targetdir, True)
+
+
+@pytest.fixture()
+def create_app(make_app, sphinx_test_tempdir, srcdir):
+    # copy plantuml.jar to current test temdir
+    plantuml_jar_file = path(__file__).parent.abspath() / "doc_test/utils"
```
danwos, line 24: Do we need to change to copy it everytime for each test?

Maybe symbolic link would do the same on linux.

Or just set the right, relative path in the conf.py files of each test_docs-project

tests/test_add_sections.py
```diff
@@ -1,11 +1,15 @@
...
+import pytest
 
 
-@with_app(buildername="html", srcdir="doc_test/add_sections")
-def test_section_is_usable_in_filters(app, status, warning):
+@pytest.mark.parametrize("buildername, srcdir", [("html", "doc_test/add_sections")])
+def test_section_is_usable_in_filters(create_app, buildername):
+    make_app = create_app[0]
+    srcdir = create_app[1]
+    app = make_app(buildername, srcdir=srcdir)
```
danwos, line 11: Can we return the result of `make_app(buildername, srcdir=srcdir)` directly from the fixture?
So that `make_app` ais laready the app to use?

tests/test_add_sections.py
```diff
@@ -1,11 +1,15 @@
...
-from sphinx_testing import with_app
+import pytest
 
 
-@with_app(buildername="html", srcdir="doc_test/add_sections")
-def test_section_is_usable_in_filters(app, status, warning):
+@pytest.mark.parametrize("buildername, srcdir", [("html", "doc_test/add_sections")])
+def test_section_is_usable_in_filters(create_app, buildername):
+    make_app = create_app[0]
+    srcdir = create_app[1]
```
danwos, line 10: Can't we take this one as test_parameter? Or do we even need it here? See next comment about app-init in fixture and not in test

danwos (changes requested):

pyproject.toml
```diff
@@ -45,9 +45,9 @@ sphinx-data-viewer = "^0.1.1"  # needservice debug output
 pre-commit = "^2"
 # test dependencies
 lxml = "^4.6.5"
+pytest = "^6.2.5"
```
danwos, line 48: Can you set `>6.2.5`please, otherwise e.g. `6.3.1`is not supported and I don't think that upcoming releases will make trouble.

tests/test_add_sections.py
```diff
@@ -1,11 +1,12 @@
...
 
-from sphinx_testing import with_app
+import pytest
 
 
-@with_app(buildername="html", srcdir="doc_test/add_sections")
-def test_section_is_usable_in_filters(app, status, warning):
+@pytest.mark.parametrize("create_app", [{"buildername": "html", "srcdir": "doc_test/add_sections"}], indirect=True)
+def test_section_is_usable_in_filters(create_app):
+    app = create_app
```
danwos, line 9: 2 small thinks:
1) Can we rename `create_app` to app, test_app, or sphinx_app?
The app is already created and usable when it arrives in the "tests".

2) `app = create_app` is not needed, we can directly use `create_app.build()...`

danwos (changes requested):

danwos (approved):

Issue #432, opened 2021-12-01
Interaction between needextend and template


adegroote:
I don't know if it is a well-defined behavior or something less expected, but template handling seems to be executed after template execution. For example,

assuming the node_template

```
Related to node {{node}}
```

and the needs

```
.. impl:: Basic req
   :id: foo
   :template: node_template

.. needextend:: foo
   :node: bar

```

the current result is something like

```
Related to node ""
```

while I expect more something like

```
Related to node bar
```

So, is the behaviour expected ?

danwos:
Ohh what a nice use case and thanks for the good bug report.
I just checked the code and indeed there is a problem.

The template gets rendered, when the need gets created. So it only has access to the current need data.

A `needextend` gets executed later, with no affect to the template.

I need to change to the execution order of some commands. This may affect a lot of stuff (or maybe not).
Will try to fix it soon.

danwos:
Made some tests and this feature is not possible.

Templates or the resulting content needs to be rendered by Sphinx, which needs to be done during handling the need-directive (so during creation of it).
The content is somehow fix when `needextend` starts to change values.

I see no way to recreate the content after everything got read, as Sphinx only supports changing the resulting docutils node-tree during this build-step.
In this step we could only replace the content-node with text or simple docutils-only based rst content (so no support for any sphinx directives or extensions). But this would have to many drawbacks and would confuse a lot.

If anyone has an idea, please share :)

adegroote:
Thanks for the feedback (and sorry for the delay, notification were not pushed on the right email).

PR #487, opened 2022-01-31
WIP: needextend event switch

danwos:
A fist test.

A special test case is missing. 

Fix for bug #432


danwos:
Does not solve the problem of #432 . So no merge

PR #489, opened 2022-02-02
Adds needs_build_json config var

### LINKED ISSUE #485: Build needs.json during a HTML build
need.json could be created quite easy also during HTML build, which can save a lot of time.

A config var like ``needs_build_json = True `` could trigger it.

danwos:
Fixes #485


arwedus:
works like a charm, thanks

Issue #486, opened 2022-01-29
embedded needs in embedded needs are not supported


PhilipPartsch:
An embedded need in an embedded need is not supported:
```
.. req:: Bread
   :id: bread_req_2
   :value: 30
   :unit: cm
   
   Our bread includes sausage.
   
   .. req:: Sausage
      :id: sausage_req_2
      :value: 30
      :unit: cm

      The sausage is delicious and on top is some ketchup.

      .. req:: Ketchup
         :id: Ketchup_req_2

          The ketchup is tasty.

   .. req:: Onion
      :id: onion_req_2

      The onions are delicious.
```

In the html or needs.json the parent_needs are always empty. 

Secondly, the algorithm to get the parents /childs,  is fetching even parents of the parents and so on. In a model of the relations, this is not good practice. We shall only provide the real parents of need (and do not give back the grands as parents, too). This is especially bad, as the grands would mark grandchilds as children, too.

PhilipPartsch:
@danwos can you close this ticket?

PR #494, opened 2022-02-10
Support embedded needs in embedded needs

PhilipPartsch:
Here I want to solve #486


sphinxcontrib/needs/directives/need.py
```diff
@@ -206,7 +206,7 @@ def get_sections_and_signature_and_needs(need_info):
         # Checking for a signature defined "above" the need.
         # Used and set normally by directives like automodule.
         # Only check as long as we haven't found a signature
-        if signature and current_node.parent and current_node.parent.children:
+        if not signature and current_node.parent and current_node.parent.children:
```
danwos, line 209: Not sure, but does the code still detects signatures, if `not signature` is used?
I have the feeling it does not.

danwos (changes requested):
The signature comment is for me not clarified.

danwos:
@PhilipPartsch, the PR looks good, I only need some feedback to may last question :)

sphinxcontrib/needs/directives/need.py
danwos: must be `None`, as `signature` is initialized  with `none` and therefore this `if` is always False

danwos (approved):

PR #493, opened 2022-02-10
Change matplotlib (+ numpy) to optional extra

OverkillGuy:
Remove need for matplotlib by default (and NumPy, transitively), per #222.

To enable matplotlib usage, install via

    pip install sphinx-needs[matplotlib]
    # equivalent via [all]
    pip install sphinx-needs[all]

Add a note about needbar/needpie also needing matplotlib in pyproject.toml (not visible till now)

Needbar and needpie directives both need matplotlib, but could fail imports if the optional dependency is not installed. Accomodate failure by raising a specific and useful error message via ImportError

Directive-specific docs updated with `versionchanged` warning about that dependency now being optional.

### High impact, high priority
- [X] Evaluate impact of feature-gating on test suite (disable/run tests differently? tests for optional import?)
- [X] Check with maintainers what level of documentation would be good enough to announce this big (breaking) change
- [X] Set up local dev environment to run tests in (see comments)

### Cosmetic, minor
- [X] Confirm the optional/extra lib name "matplotlib" is OK (I considered `needpie` for feature that needed it, but then discovered `needbar` also uses it, and more may use it in future)
- [X] Confirm with maintainers the importerror message is good enough
- [X] Confirm with maintainers the directive docs entry is good enough
- [X] Check with maintainers what to do about implicit (transitive) dependency on NumPy (see comment in needbar.py)
- [X] Update the `nextversion` message in directives for `versionchanged` block
- [x] Add `docs/changelog.rst` entry
- [x] Add myself to Authors

Issue #222, opened 2021-03-16
Matplotlib pulls numpy: make lib optional


OverkillGuy:
I was surprised to find `numpy` as part of my dependencies when I don't do any maths.

NumPy is a fantastic tool, but it's heavy and clearly unnecessary for my project. It turns out, numpy is a required dependency of `matplotlib`, which is now part of `needs` deps since 0.5.0.

I won't personally be doing any pie-chart-ing, so wouldn't really benefit from use of matplotlib, nor numpy. Can we make the `matplotlib` an optional dependency please? This way, people who are in more "constrained" systems don't need to pull in the whole maths ecosystem.

From [setuptools docs on Optional Dependencies](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies), adapted for us:
```python3
    extras_require={
        "matplotlib":  ["matplotlib"],
    }
```

As for gating the feature in python, from [SO](https://stackoverflow.com/a/15928170):

```python3
try:
    import matplotlib
except ImportError:
    # dependency missing, issue a warning
    import warnings
    warnings.warn('matplotlib not found, please install 'matplotlib' extra to enable pie-charts')
```

I realize there's a documentation overhead to doing this, but think it's worth it.

danwos:
Yes, I agree.
There are some (huge) dependencies, which are used for some directives only and therefore are often not needed for most use-cases.

Instead of a warning I would like to throw an exception, as the final documentation build will not contain the expected content (e.g. no pie charts).
Otherwise I'm afraid that a user may miss the warning, as sphinx-build logs are normally not so small.

This could also be configurable, e.g. ``needs_missing_import_exceptions = False`` (``True`` is default).

Any concerns about raising an exception and stopping build in such a case?

OverkillGuy:
100% agreed. I have no issues with errors instead of warnings, the only reason I showed warnings was when searching for "python check extra_require imports" docs, which got me a warnings-based example.

Another example I can think of is sphinx-plantuml (which I enjoy a lot!) but isn't necessary unless using graphs. It requires a jar file that isn't trivial to install/configure (a small pain of its own, sadly), and a lot of installs can just skip it.

OverkillGuy:
Wrote up a (draft) PR to do this, as I just bumped into NumPy again on a new project =)

[...]

PR #495, opened 2022-02-13
Add explicit plantuml dependency as optional

### LINKED ISSUE #482: Make sphinxcontrib.plantuml a fix dependecy
There are some imports in the code for `sphinxcontrib.plantuml`,  which don't check if plantuml is even installed and available.

In the past, it was an optional installation, if `needflow` shall be used.
But right now, it becomes a major dependency and should be part of pyproject.toml file.

OverkillGuy:
Related to #222 for matplotlib + depending on PR #493.
Fixes #482 by adding plantuml as (optional) dependency.

Update docs for new optional deps.

Open questions:
- [ ] What to do about no_plantuml func, see diff
- [ ] What is minimum plantuml version to allow? (currently latest = 0.22)
- [ ] What about the documentation

Note: This PR is meant to be on top of #493, but I couldn't make the PR actually diff from that PR (on my repo) to this HEAD without the PR being on my repo = invisible to yours. So now this PR is from master of upstream, hence contains both matplotlib PR + plantuml PR jumbled together as sequence of commits. Ugh. Feel free to untangle either/both feature proposals as you feel is best, this is just a simple implementation proposal with little actual thought put in, hoping to help your (busy) team.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -21,3 +21,5 @@ Daniel Eades <danieleades@hotmail.com>
 Philip Partsch <philip.partsch@googlemail.com>
 
 David Le Nir <david.lenir.e@thalesdigital.io>
+
+Jb Doyon <jb@jiby.tech>
--- a/README.rst
+++ b/README.rst
@@ -45,14 +45,18 @@ Using poetry
 ------------
 ::
 
+    poetry add sphinxcontrib-needs[all]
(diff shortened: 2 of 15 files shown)


OverkillGuy:
As mentioned on the ticket, the PR is unnecessary as the plantuml feature is now mandatory (rather than optional like it was implied to be before).
Closing PR as we won't need it.

PR #501, opened 2022-02-24
Delete temporary dir at the end of build

Youcef-DZ:


sphinxcontrib/needs/builder.py
```diff
@@ -86,3 +87,10 @@ def build_needs_json(app, exception):
     needs_builder = NeedsBuilder(app)
     needs_builder.set_environment(app.env)
     needs_builder.finish()
+    
+    
+def delete_temp_dir(app, exception):
```
danwos, line 92: `builder.py` is more used for the ``needs`` builder.
Maybe we put this small function directly inside ``needs.py`` or the ``urils.py`` file?

sphinxcontrib/needs/builder.py
```diff
@@ -86,3 +87,10 @@ def build_needs_json(app, exception):
...
+
+    tempdir = os.path.join(app.srcdir, "_temp_needs")
```
danwos, line 94: can we put this magic path to ``defaults.py``? E.g. ``TEMP_FoLder``
So other functions can reuse it and we have one single place to change it if needed later.

sphinxcontrib/needs/needs.py
```diff
@@ -317,6 +317,7 @@ def setup(app):
     app.connect("doctree-resolved", process_need_outgoing)
     app.connect("doctree-resolved", process_need_count)
     app.connect("doctree-resolved", process_need_func)
+    app.connect("build-finished", delete_temp_dir)
```
danwos, line 320: I think we also need a ``create_temp_dir`` and connect to an early event.
I know it gets created also inside ``needpie.py`` line 236, but I think we should be safe here and create this central folder at a central place. :)

danwos (changes requested):
Ohh @Youcef-DZ , thanks for this really fast Follow-up PR. :+1: 

I like it a lot and have only 2-3 remarks.

I think we should also have a small test case, which checks if ``_temp_needs`` gets created and gets also deleted.

I'm also not yet sure if we really should delete the temp folder, as it may be get reused again by Sphinx for an incremental build. 
Have you tested this scenario?

PR #510, opened 2022-03-02
Fixes buggy file registration

### LINKED ISSUE #455: On Windows, calling build() multiple times on the same sphinx.application.Sphinx instance will multiply sphinx-needs css and js entries.
On windows:
Calling build() multiple times on the same sphinx.application.Sphinx instance will add additional sphinx-needs css and js each time, resulting in multiplication of entries in produced html.
Root cause is the usage of pathlib.Path in environment.py  safe_add_file ([environment.py#L34](https://github.com/useblocks/sphinxcontrib-needs/blob/af76b5b4e28e50f569c95a7cf8e9c733f7da5e2b/sphinxcontrib/needs/environment.py#L34])), which will cause a WindowsPath to be rendered.

Forward slashes (/) should be used for src= and href= html entries also on windows.

The problem can bee seen in the debug output:
debug

danwos:
Fixes #455

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,6 +12,8 @@ License
 
 * Bugfix: Correct image registration to support build via Sphinx API.
   (`#505 <https://github.com/useblocks/sphinxcontrib-needs/issues/505>`_)
+* Bugfix: Correct css/js file registration on windows.
+  (`#455 <https://github.com/useblocks/sphinxcontrib-needs/issues/455>`_)
 
 0.7.6
 -----
--- a/sphinxcontrib/needs/environment.py
+++ b/sphinxcontrib/needs/environment.py
@@ -17,7 +17,7 @@ if parse_version(sphinx_version) >= parse_version("1.6"):
 IMAGE_DIR_NAME = "_static"
 
 
-def safe_add_file(filename: PurePosixPath, app: Sphinx):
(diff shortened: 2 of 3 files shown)


StephanKuempel:
You might want to rename the test since from test_sphinx_api_needpie, since it is not related to needpie

StephanKuempel:
You might also rename the function according to the filename (still is test_sphinx_api_needpie():)

StephanKuempel:
filename is still not converted to posix on windows

Use 
´´´app.add_js_file(filename.as_posix())´´´ and ´´´app.add_css_file(filename.as_posix())´´´

StephanKuempel:
Use like this:

    static_data_file = PurePosixPath("_static") / filename
    filename = PurePosixPath("filename")

    if filename.suffix == ".js":
        # Make sure the calculated (posix)-path is not already registered as "web"-path
        if hasattr(app.builder, "script_files") and str(static_data_file) not in app.builder.script_files:
            app.add_js_file(str(filename))

    elif filename.suffix == ".css":
        if hasattr(app.builder, "css_files") and str(static_data_file) not in app.builder.css_files:
            app.add_css_file(str(filename))
    else:
        raise NotImplementedError(f"File type {filename.suffix} not support by save_add_file")

Issue #390, opened 2021-10-18
Need more stable URL to reference need elements in HTML 


twodrops:
Sphinx-Needs uses the standard HTML URLs of Sphinx-generated HTMLs to refer to elements with IDs
For example, https://server/sphinx-needs-example-folder/sw_requirements.html#REQ_SwCps_Requirement1

If the folders/files are moved, then the reference to the elements are broken.

It would be nice to have another way to uniquely reference elements with just ID
For example, https://server/needs?query=REQ_SwCps_Requirement1

arwedus:
@danwos can you narrow down a date for this feature? We have several use cases for it, would be cool to get this in the next few months.

danwos:
Next few months is possible ;)

The sponsored development (by Prototype Fund) of Open-Needs starts on March and there we need a react based JS Frontend, which will have such a Feature.
As the Open-Needs Server will not be directly available, I think we can use a generated needs.json.
And maybe have something at mid/end of April (if develpmend is done in this context only.)

But maybe we can speed things up, by discussing it with @twodrops internally.

r-o-b-e-r-t-o:
Hi Daniel, do you have an update on this issue?

twodrops:
@r-o-b-e-r-t-o We are thinking of creating a reactjs-based interface parallel to a sphinx-theme. This means, instead of static HTMLs, the HTMLs will be rendered dynamically using javascript. It goes in the direction of this [idea](https://github.com/useblocks/sphinxcontrib-needs/discussions/211#discussioncomment-492378). This does not replace HTML based theme approach, but is more suitable for usecases which expect dynamic rendering like better responsive UI, filtering etc. The above issue matchings the reactjs approach better as there has to be a javascript based backend which translates the queries.  AFAIF, @danwos wanted to start with a protoype of this together with his colleague.



@danwos

r-o-b-e-r-t-o:
@twodrops Thanks for the update. This seems to be a big feature. May it be possible to do the backend part first and afterwards the UI stuff? This way we could already use permalinks, which we urgently need for get our bi-directional traceability across tools working.

twodrops:
@r-o-b-e-r-t-o Yes, I guess it should be possible to start with this feature in the existing setup as well and port it to reactjs theme later. @danwos What do you think?

danwos:
I agree and I also have thought about it in the last days.

So here is my 
 
### technical concept

An additional page `permalink.html` is added to the project-html build.
[...]

PR #526, opened 2022-03-31
Issue 390

r-o-b-e-r-t-o:
Implement permalink feature as described here: #390


danwos:
Thanks for the PR :+1: .

There is an error with `PackageLoader("needs")`, where "needs" is unknown to Jinja.

Also, is this PR still a draft, or could it be merged, when CI is green?

Oh and 2-3 additional hints:
- If you like you can add your name to the AUTHORS file
- Please add also a line to `docs/changelog.rst`.
- Maybe also some docs are needed?

But the last two steps I can also do after this PR got merged.

Thanks again for your contribution.

r-o-b-e-r-t-o:
> Also, is this PR still a draft, or could it be merged, when CI is green?

Thanks for your input.
It's still a draft. Once I'm done I change this to a normal PR and then it can be merged once CI is green.

sphinxcontrib/needs/layout.py
```diff
@@ -777,18 +783,6 @@ def image(self, url, height=None, width=None, align=None, no_link=False, prefix=
         # Otherwise the images gets not copied to the later build-output location
         self.app.env.images.add_file(self.need["docname"], url)
 
-        # Okay, this is really ugly.
```
r-o-b-e-r-t-o, line 780: @danwos I removed this because using image link did not work as expected with this. Probably one could also remove https://github.com/useblocks/sphinxcontrib-needs/blob/d40897e023349b9e301de4cccbbf71241d12cd80/sphinxcontrib/needs/libs/html/sphinx_needs_collapse.js#L50 . I did not do it myself because I was afraid of side effects.

sphinxcontrib/needs/layout.py
```diff
@@ -829,7 +823,7 @@ def link(self, url, text=None, image_url=None, image_height=None, image_width=No
         link_node = nodes.reference(text, text, refuri=url)
 
         if image_url:
-            image_node = self.image(image_url, image_height, image_width)
+            image_node = self.image(image_url, image_height, image_width, no_link=True)
```
r-o-b-e-r-t-o, line 826: @danwos I had to change this to make image links work.

sphinxcontrib/needs/layout.py
```diff
@@ -887,6 +881,43 @@ def collapse_button(self, target="meta", collapsed="Show", visible="Close", init
...
+        if image_url is None and text is None:
+            image_url = "icon:share-2"
+            image_width = "17px"
+
+        config = self.app.config
+        permalink = config.needs_permalink_file
+        id = self.need["id"]
+        permalink_url = permalink + "?id=" + id
+
+        return self.link(
```
r-o-b-e-r-t-o, line 912: @danwos Here is the use of the modified image link function.

docs/conf.py
```diff
@@ -336,6 +346,9 @@
     },
 }
 
+# build needs.json to make permalinks work
+needs_build_json = True
```
r-o-b-e-r-t-o, line 350: @danwos I activated this to make permalinks work in the needs docs.

r-o-b-e-r-t-o:
@danwos Unless you want me to change something this PR can be merged once CI is green.

docs/conf.py
```diff
@@ -294,11 +294,21 @@
     "example": {
         "grid": "simple_side_right_partial",
         "layout": {
-            "head": ['**<<meta("title")>>** for *<<meta("author")>>*'],
+            "head": ['**<<meta("title")>>** for *<<meta("author")>>*>>'],
```
danwos, line 297: `>>`, I think  added by mistake

docs/conf.py
danwos: makes sense

docs/configuration.rst
```diff
@@ -1498,6 +1498,45 @@ Default: False
    The created ``needs.json`` file gets stored in the ``outdir`` of the current builder.
    So if ``html`` is used as builder, the final location is e.g. ``_build/html/needs.json``.
 
+.. _needs_permalink_file:
+
+needs_permalink_file
```
danwos, line 1503: Is this option really needed?
I mean, do you see any use cases where a user wants to specify his/her own permalink.html file?

For me, only the used file name in the final `_build/html/ for `permalink.html? to be configurable. (e.g. `sttic_ids.html` instead of permalink)
The file content itself can be always the same.

What do you think?

sphinxcontrib/needs/layout.py
```diff
@@ -767,7 +768,12 @@ def image(self, url, height=None, width=None, align=None, no_link=False, prefix=
 
             url = file_path
 
-        image_node = nodes.image(url, classes=["needs_image"], **options)
+        if no_link:
+            classes = ["needs_image", "no-scaled-link"]
```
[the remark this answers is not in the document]
danwos: Makes sense and good solution.

sphinxcontrib/needs/layout.py
danwos: interesting. Needed some time to understand this, but I think it makes sense as there were 1-2 bugs in the code of image() and link handling links with images. Thanks.

sphinxcontrib/needs/layout.py
```diff
@@ -887,6 +881,43 @@ def collapse_button(self, target="meta", collapsed="Show", visible="Close", init
...
+            <<permalink(image_url="icon:link")>>
+            <<permalink(text='¶')>>
+        """
+
+        if image_url is None and text is None:
+            image_url = "icon:share-2"
+            image_width = "17px"
+
+        config = self.app.config
+        permalink = config.needs_permalink_file
```
danwos, line 908: I guess here is a small bug.

The default value is `/permalink.html`, which is used for the physical file location and the relative URL.
These two may not be the same, as docs may be stored under e.g. `examples-com/docs/v1.2/...`.

So the docs-root, used for the build, is not the same root, as used on the webserver.

I also think that the link to the permalink-file can not be absolute/static and can't be based on the webserver-root.
This is mainly because the weberser-root path may change or be different.

Scenario: docs get deployed twice, but were build only once:
1. example.com/latest/...
2. example.com/releases/1.1/...

So I think we need to calculate here a relative link to `permalink.html`, based on the current location of the need in the url-path.
I think something like this is already done at 1-2 places in the code and as far as I remember it was just "counting" the `/` (slahes) in need.docname and then `amount_slashes * "../" + "permalink.html"`.

danwos (changes requested):
Great PR.
Unfortunately, I was not able to test the permalink feature in the docs, based on a URL-calculation bug (details in one of the comments.)

We should also think about adding a small test case, which tests at least the existence of the permalink-link and the correct URL. But this could be part of another PR, after this one got merged :)

sphinxcontrib/needs/layout.py
danwos: Ohh and the given path here may be also Windows-specific, so you would have backslashes in your URL.
This check may get irrelevant with the needed changes from above.

docs/configuration.rst
r-o-b-e-r-t-o: I mainly wanted the path to be configurable. 
In our project the web server does not host the _build/html directory, but the _build directory.
So I would need to change the needs_permalink_file option from the default "/peramlink.html" to "/html/permalink.html".

sphinxcontrib/needs/layout.py
r-o-b-e-r-t-o: I changed the default value. Now the permalink file option is meant to just change the file name as you suggested. This file is expected to be in the html build dir, i.e. next to index.html.

docs/configuration.rst
r-o-b-e-r-t-o: Agreed and changed. See below comment.

sphinxcontrib/needs/layout.py
danwos: I think this is not enough.
The permalink-link can not be static or relative to web-root, as root may change.

When I build the Sphinx-Needs docs locally, I get e.g. such locations in my browser:
``file:///home/daniel/workspace/sphinx/sphinxcontrib-needs/docs/_build/html/examples/index.html``

In this html-file, a need with a permalink is available. But the link is:
``file:///permalink.html?id=EX_REQ_5``

I came to the conclusion that we will not be able to calculate a link from any need object to the permalink page for each use case during build time.

First of all the calculated link must not contain any relative parts, as this link may not work anymore if the folder structure changes. E.g. `company.com/docs/part/chapter/../../permalink.html` will not work, if e.g. ``chapter`` folder gets renamed.

Second, we can not create an absolute link during build time, as this would mean that we know under which URL the final documentation gets deployed. And even if we know it, it would not work for local builds and server deployments with one build. Only one correct "base URL" could be set.

For me this means, we need to calculate the final permalink-url via JavaScript, as only then we are able to figure out, under which URL the docs got deployed.
Untested JS code:
```js
/* ID must get extracted via JS somehow */

$('.need-permalink').attr("href", URL("permalink.html?id=" + need_id, document.baseURI).href)
```
Then it should not matter, how often and under which domains the docs got deployed.

What do you think, makes this sense?

docs/configuration.rst
```diff
@@ -1498,6 +1498,45 @@ Default: False
...
+~~~~~~~~~~~~~~~~~~~~
+
+The option specifies the name of the permalink html file,
+which will be copied to the html build directory during build.
+
+The permalink web site will load a ``needs.json`` file as sepcified
```
danwos, line 1509: sepcified - > specified

sphinxcontrib/needs/templates/permalink.html
```diff
@@ -0,0 +1,68 @@
...
+                        docname = need['docname'];
+                        return;
+                    }
+                });
+
+                window.location.replace(pathname + docname + '.html#' + id);
+            });
+        }
+
+        function getParameterByName(name, url = window.location.href) {
```
danwos, line 51: Just a small hint, by default jQuery is available on Sphinx docs.

So I think some stuff could be solved much easier/shorter with jQuery.
Just for the case, you need to touch some js-parts again.
Otherwise it's fine :)

danwos (changes requested):

sphinxcontrib/needs/templates/permalink.html
r-o-b-e-r-t-o: > What do you think, makes this sense?

Totally. With the long break I somehow forgot about this complication ;-)
Thanks for your thorough review 👍 

I changed the computation of the link in `layout.py` to a relative link based on the need docname as you described earlier.
The web browser automatically resolves this to an absolute URL when the web page is rendered.
I tested with a local webserver one time hosting _build/html and another test hosting _build.
It worked.
I also tested opening without web server directly from Windows file explorer. Here the link looked good as well but the permalink page itself does not work because of CORS error when loading needs.json.

So I guess I can get arround more JS "magic" ;-)

Please test for yourself again.

I hope I understood you correctly and it is ok now.
Otherwise just tell me and I hope for your continued patience ;-)

sphinxcontrib/needs/layout.py
r-o-b-e-r-t-o: > What do you think, makes this sense?

Totally. With the long break I somehow forgot about this complication ;-)
Thanks for your thorough review 👍 

I changed the computation of the link in `layout.py` to a relative link based on the need docname as you described earlier.
The web browser automatically resolves this to an absolute URL when the web page is rendered.
I tested with a local web server one time hosting _build/html and another test hosting _build.
It worked.
I also tested opening without web server directly from Windows file explorer. Here the link looked good as well but the permalink page itself does not work because of CORS error when loading needs.json.

So I guess I can get around more JS "magic" ;-)

Please test for yourself again.

I hope I understood you correctly and it is OK now.
Otherwise just tell me and I hope for your continued patience ;-)

danwos (approved):
Works now :)
And a huge "Thank you!" for this cool, new feature :+1:

Issue #531, opened 2022-04-27
Docker for Sphinx-Needs


twodrops:
An "official" docker image for sphinx-needs is missing.
https://github.com/tlwt/sphinxneeds-docker does not seem to be maintained anymore.

I would like to contribute the one we are using internally since an year now. 
Incase you would like to have it, could you please create a new repo, for example,  "sphinxneeds-docker"

danwos:
I'm not sure if a dedicated docker image for Sphinx needs makes sense.

Maybe a more generic image for Sphinx in Industry Environments would be more helpful.
For sure it would contain Sphinx-Needs as well

twodrops:
Ok.

These are the contents of the two base images we are using. I do not see anything specific for Industry in these and IMHO this could come in handy for anyone wanting to start with Sphinx and Sphinx-Needs.

## Image Variants
The sphinxneeds images come in two flavors, each designed for a specific use case containing sphinx-needs. 

### `sphinxneeds:<version>`
This is the defacto image (size ~ 460M). If you are unsure about what your needs are, you probably want to use this one. It is designed to be used both as a throw away container (mount your source code and start the container to start your app), as well as the base to build other images off of.

#### Included Tools

The image includes the following tools.

| Tools                                       |   
| ---------------------------------------------- | 
| graphviz |
| jre |
| latexmk |
| texlive |
| open-needs-ls |
| plantuml |
| python |
| sphinx |
| sphinxcontrib-needs |
| sphinxcontrib-plantuml |

### `sphinxneeds:<version>-slim`
This image does not contain the common packages contained in the above version and only contains the minimal packages needed to run sphinxneeds (size ~ 200M). Unless you are working in an environment where only the sphinxneeds image will be deployed and you have space constraints, we highly recommend using the default image of this repository.
 
#### Included Tools

The image includes the following tools.

| Tools                                       |   
| ---------------------------------------------- | 
| python-slim |
| sphinx |
| sphinxcontrib-needs |

**NOTE** Features such as report generation or PDF generation are not supported in this image. Please use the full-version above for such usecases.

twodrops:
@danwos 

Could you please add the useblocks Docker Hub credentials as secrets to this repo?
```
secrets.DOCKERHUB_USERNAME
secrets.DOCKERHUB_TOKEN
```

danwos:
@twodrops: Dockerhub secrets got added.

PR #532, opened 2022-05-02
Add Support for Docker

twodrops:
Implements #531 

- Dockerfile
- Script for local build
- Pipeline update for Docker
- Documentation changes

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,4 +1,4 @@
-name: Docker Build Image
+name: Docker-Image
 
 on:
   # Allow manual trigger
@@ -71,7 +71,7 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           push: ${{ github.event_name != 'pull_request' && steps.deploycheck.outputs.value == 'y' }}
-          file: Dockerfile
+          file: docker/Dockerfile
           build-args: |            
             NEEDS_VERSION=${{ env.NEEDS_VERSION }}
             BASE_IMAGE=${{ matrix.base-image }}                 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,4 +1,4 @@
-name: Docker Build HTML Image
+name: Docker Build Image
 
 on:
   # Allow manual trigger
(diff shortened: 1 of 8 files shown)


.github/workflows/docker.yaml
```diff
@@ -0,0 +1,71 @@
...
+        default: 'n'
+  push:
+    tags:
+      - '[0-9].[0-9]+.[0-9]+' 
+      
+env:
+  DOCKER_REGISTRY_PATH: useblocks
+  IMAGE_NAME: sphinxneeds
+  BASE_IMAGE: sphinxdoc/sphinx:latest
+  NEEDS_VERSION: 0.7.8
```
twodrops, line 20: @danwos Is the current release version available during the pipeline run, to avoid hard-coding here? This is needed to version to docker image

twodrops, line 20: @danwos Is the current release version available during the pipeline run, to avoid hard-coding here? This is needed to version to docker image

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,71 @@
+name: Docker Build HTML Image
```
twodrops, line 1: Could you please create me a feature branch in useblocks repo `feature/531_add_docker_support`?
I will do a temporary merge to the feature branch so that I can test this workflow file. I do not think I can test a workflow otherwise on the useblocks CI.

twodrops, line 1: Could you please create me a feature branch in useblocks repo `feature/531_add_docker_support`?
I will do a temporary merge to the feature branch so that I can test this workflow file. I do not think I can test a workflow otherwise on the useblocks CI.

danwos: No problem, I just gave you the WRITE role.
So you should be able to create and push any needed branch.

.github/workflows/docker.yaml
danwos: Not sure, maybe a task in the workflow is needed like `pip install .` to install it and make it available.

twodrops: ah sorry. I just need the version number of the release version, not the tool. I do not want to hardcode the version number and forget it update it during release. So a variable or a config file where the current version is available so that I can reference it

danwos: Got it.
There are files containing this number, but I'm not sure if there is a way to get it easily.
The best file would be the pyproject.toml : https://github.com/useblocks/sphinxcontrib-needs/blob/master/pyproject.toml

But for me, it would be ok to raise it by hand with each release.
We already need to do it for pyproject.toml, needs.py, and conf.py.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+on:
+  # Allow manual trigger
+  workflow_dispatch:
+    inputs:
+      # Ask if docker depoyment needed when manually triggered
+      deploy:      
+        description: 'Deploy Docker Image (y/n)?'
```
twodrops, line 9: The workflow file has to be on master once to trigger the workflow even for testing even with a dispatch I guess, I will test it once it is on master if the pipeline build of the image works.

twodrops: @danwos I added a trigger to build on feature branch (temporarily) to trigger the build and it worked.
The image is building but the push to docker.io registry failed.
https://github.com/useblocks/sphinxcontrib-needs/runs/6263751379?check_suite_focus=true#step:7:1755
Could you please confirm that the Docker Token you added has the rights to push?

twodrops: @danwos Could you please let me know your dockerhub username? I assumed it is useblocks but cannot find the pushed image under this user on dockerhub :)

danwos: Ohh I didn't know that there is a useblocks user.
I will try to reactivate it ...

danwos (approved):

danwos:
Thanks for the work and the PR :+1:

PR #539, opened 2022-05-06
Adds intial esbonio config and docs

danwos:


haiyangToAI:
Close in favor of #584

PR #546, opened 2022-05-12
Fix references in needtables when building with rinohtype

JMAlego:
While using Sphinx-Needs with a [rinohtype](https://github.com/brechtm/rinohtype)-based builder I discovered that links in needtables did not link properly inside the generated PDF while links in need definitions did. It seems this is caused by references in needtables being created without the `internal=True` attribute which is added to most other references automatically becuase they are created using the `make_refnode` utility ([defined here](https://github.com/sphinx-doc/sphinx/blob/31eba1a76dd485dc633cae48227b46879eda5df4/sphinx/util/nodes.py#L552)).

This change simply adds the `internal=True` attribute to needtable references if they are not marked as external. This means that needtables will build properly with the rinohtype builder (and possibly other builders which rely on the `internal` attribute) and should not affect other builders negatively.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -27,3 +27,5 @@ Baran Barış Yıldızlı <arisbbyil@gmail.com>
 Roberto Rötting <roberto.roetting@gmail.com>
 
 Nirmal Sasidharan <nirmal.sasidharan@de.bosch.com>
+
+Jacob Allen <jacob.allen@etas.com>
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,6 +12,7 @@ License
 
 * Bugfix: Support embedded needs in embedded needs.
   (`#486 <https://github.com/useblocks/sphinxcontrib-needs/issues/486>`_)
+* Bugfix: Correct references in :ref:`needtable`s to be external or internal instead of always external.
 
 0.7.9
 -----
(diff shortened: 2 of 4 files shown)


danwos:
Thanks for the PR and the detailed description 👍 
Please add also a short line in the docs/changelog.rst.
This would help others with similar problems to find this fix.

And for sure feel free to add yourself to the AUTHORS file. 🥇 

As this is your first PR in this project, I need to manually allow a CI run. This sometimes may need some time....

danwos:
Ohh the CI tests found something.
The important line is this one: 
`` assert '<a class="reference external" href="#impl1">impl1</a>' in code_args_html`` in test_doc_needs_filter_code
https://github.com/useblocks/sphinxcontrib-needs/runs/6410631322?check_suite_focus=true#step:7:113

I think a simple HTML code comparison does not match anymore, as ``internal=True`` sets different HTML classes.

JMAlego:
I've updated the test to match the HTML as generated now. The link in the test is within the same page and so the `internal` class is expected.

I've also updated the changelog and added myself to the authors file.

JMAlego:
It looks like all the checks have passed this time 🎉

danwos:
Looks good and merged. Thanks for fixing this.

PR #549, opened 2022-05-14
set up mypy hook for incremental adoption

danieleades:
this pull request sets up mypy to allow for incremental adoption of type annotations.

it does this by adding a 'mypy' pre-commit hook, and a 'whitelist' to the `pyproject.toml` file such that all the existing errors are ignored. Over time, this whitelist can be reduced as modules are incrementally 'typed'. New modules should be fully-typed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,3 +25,13 @@ repos:
       - id: pyupgrade
         args:
           - --py36-plus
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.950
+    hooks:
+      - id: mypy
+        pass_filenames: false
+        additional_dependencies:
+          - types-requests==2.27.25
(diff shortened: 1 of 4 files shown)


danieleades:
@danwos interested to gauge how you feel about this one. On the one hand, type annotations prevent a lot of errors, and allow an IDE to provide many more code completions. On the other hand, sphinx is an ageing codebase, and it's type annotations are not the best.

i've also currently have 'strict' mode enabled, which *requires* that all functions have type annotations. with strict mode set to `false`, it will only check for type errors where the annotations are present. Personally i'm in favour of 'strict' type checking, but it could be disabled.

danwos:
Type annotation is fine, if you are starting with it :)
Adding it to existing code means a lot of work.
But for sure it makes sense to have this and I like your approach of making this optional for the moment.

Thanks, for PR :+1:

danwos (approved):

PR #551, opened 2022-05-16
Theme update into 1.0

danwos:


danwos:
Argh sh**, there seem to be problems with running pytest-tests, as they seem to keep their "old" sphinx-env from the last test case run.

PR #561, opened 2022-05-18
use python 3.10 in CI

danieleades:
As you can see, the specified version of `numpy` fails to build with Python 3.10


danwos:
We can close this PR, as it got already implemented with 1.0 branch. Which got merged today.
https://github.com/useblocks/sphinxcontrib-needs/blob/master/.github/workflows/ci.yaml#L13

PR #555, opened 2022-05-16
slightly improve typing in 'needsfile'

danieleades:

PR #565, opened 2022-05-21
slightly improve typing in 'needsfile'

danieleades:
converts an untyped dictionary to a class

### DIFF
--- a/sphinxcontrib/needs/needsfile.py
+++ b/sphinxcontrib/needs/needsfile.py
@@ -7,6 +7,7 @@ import json
 import os
 import sys
 from datetime import datetime
+from typing import Any, List
 
 from jsonschema import Draft7Validator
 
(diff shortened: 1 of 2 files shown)


sphinxcontrib/needs/needsfile.py
```diff
@@ -131,7 +132,12 @@ def load_json(self, file):
             self.log.debug(f"needs.json file loaded: {file}")
 
 
-def check_needs_file(path):
+class Errors:
+    def __init__(self, schema_errors: List[Any]):
```
danieleades, line 136: @danwos what is the 'type' of the schema errors? I'd prefer to avoid using `Any` if i can

danwos:
Isn't this the same as #555?

I'm wondering because github shows here some diffs, but the code changes are already part of master.
Or do I miss something?

danieleades:
> Isn't this the same as #555?
> 
> I'm wondering because github shows here some diffs, but the code changes are already part of master. Or do I miss something?

huh. not sure what's happened there. Maybe i pushed another commit after you merged? i think this can be closed.

PR #568, opened 2022-05-22
improve typing annotations

danieleades:
- improves type annotations throughout the codebase
- bumps versions in pre-commit config
- de-globs and tightens mypy whitelist


danieleades:
Finding it a little challenging to make much further progress on this... might need the author to fill in some of the blanks!
(@danwos ;) )

sphinxcontrib/needs/api/configuration.py
```diff
@@ -118,7 +124,13 @@ def my_function(app, need, needs, *args, **kwargs):
...
 
 
-def add_warning(app, name, function=None, filter_string=None):
+# 'Need' is untyped, so we temporarily use 'Any' here
+WarningCheck = Callable[[Any, SphinxLoggerAdapter], bool]
+
+
+def add_warning(
+    app: Sphinx, name: str, function: Optional[WarningCheck] = None, filter_string: Optional[str] = None
+) -> None:
```
danieleades, line 133: hey @danwos can you check my work here? the function signature (and purpose) of the 'warning check' function don't seem to be documented anywhere, but there are examples, which is where i've tried to infer the signature from.

I've used `Any` for the 'need' since the need is currently untyped anyway.

danieleades: on an unrelated note, this function can accept a `function` or a `filter_string` arg, and has checks to ensure that there is at least one set, and not both. You could remove these checks by changing the signature to

```python
def add_warning(app: Sphinx, name: str, check: Union[WarningCheck, str]) -> None:
```

and then doing an `isinstance` check to see which one was used.

that's a breaking change, so doesn't belong in this PR

danwos: This looks good.
And oh yes, docs are totally missing for the warning functions.

It's really great to see typing annotation being added. Especially as I'm still learning a lot of tricks from you.
So thanks for all the free lessons and for sure also for the PRs :+1:

sphinxcontrib/needs/directives/need.py
```diff
@@ -107,9 +123,9 @@ def run(self):
         )
         return need_nodes
 
-    def read_in_links(self, name):
+    def read_in_links(self, name: str) -> List[str]:
         # Get links
-        links_string = self.options.get(name, [])
+        links_string = self.options.get(name)
```
danwos, line 128: MMh wait, why does this work?
Old code was ``links_string = self.options.get(name, [])`` 

For me, it makes sense to have a default value here because as far as I know not all options are available if not set in the directive.
Same thoughts for all the `self.options.get()` calls above?

Is this change related to the typing annotations or is it generally not needed?

sphinxcontrib/needs/layout.py
```diff
@@ -506,15 +507,15 @@ def meta(self, name, prefix=None, show_empty=False):
             for index, element in enumerate(data):
                 if index > 0:
                     spacer = nodes.inline(classes=["needs_spacer"])
-                    spacer += nodes.Text(", ", ", ")
+                    spacer += nodes.Text(", ")
```
danwos, line 510: Oh ok, no second parameter anymore.
This also has not worked for me in the past.
But I had last checked with Sphinx version <4 I think....
So nice simplification

danwos (commented):
Wow, what a huge and wonderful PR.
Love it a lot and to be honest, I would never find the time to do this kind of enhancement.
So how long have you worked on this? : )

Thanks a lot for this awesome PR.
I only have 1-2 comments, to understand some stuff a little better.

sphinxcontrib/needs/directives/need.py
danieleades: `dict.get` has a default default value of `None`, so it's not necessary to set a value here unless `None` is not appropriate.
i removed the default list because it doesn't make sense (from a type-safety perspective) to assign an empty list to a variable which would otherwise be used to represent a `str`

sphinxcontrib/needs/layout.py
danieleades: yeah i wasn't sure when it was deprecated. I was concerned it would break some of the sphinx versions in the test matrix, but it doesn't seem to?

danwos (approved):

PR #569, opened 2022-05-22
prefer 'pathlib' to 'os.path'

danieleades:


danieleades:
i'll close this for now. this is exactly the kind of refactoring that's dramatically easier in a type-safe code-base

PR #573, opened 2022-05-23
Replaced logo with the current one. Adds also favicon and header logo

danwos:


iSOLveIT:
Looks good. Although, in my opinion, the current logo for the homepage is nice.

iSOLveIT (approved):
Looks good. Although, in my opinion, the current logo for the homepage is nice.

danwos:
I agree, also like it a lot. But the official logo was also missing.
Next try, now both of them are combined.
I think it is good enough to start with for the moment.

PR #578, opened 2022-05-23
apply inting to docs config

danieleades:


danwos (approved):

danwos:
Thanks for this little, helpful PR :+1:

docs/conf.py
```diff
@@ -545,18 +559,18 @@
 ]
 
 rst_epilog = """
-.. |ex| replace:: **Code** 
+.. |ex| replace:: **Example**
 
-.. |out| replace:: **Output** 
+.. |out| replace:: **Result**
```
danieleades, line 564: hi @danwos was this change expected? could be the result of a missing rebase?

danieleades: Ha I don't have any skin in the game here. They seem like synonyms to me

iSOLveIT: I used Result instead of Output because it describes what is beneath better than Output.

Also, Code described the content beneath it better than Example.

But all words are synonymous though.

iSOLveIT: I will be sending a PR to update theme's CSS so I can add this change to it.

PR #580, opened 2022-05-24
fix typing in 'environment'

danieleades:
this is an experiment. `PurePosixPath` is mixed with `Path`, which causes mypy to complain. Would be possible to fix the annotations *and* use `PurePosixPath` but first I wanted to check if it was really necessary.


danieleades:
i guess this is needed in windows...

back to the drawing board!

any background you can give me on this issue would be greatly appreciated

PR #584, opened 2022-05-25
Support ide language features

haiyangToAI:


sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+    index = 0
+    for word in words:
+        length = length + len(word)
+        if col <= length + index:
+            return index
+        index = index + 1
+    return index - 1
+
+
+def get_lines(ls, params) -> List[str]:
```
danieleades, line 42: type annotations?

danieleades (changes requested):

danieleades:
what happens if you add the `esbonio` package to the list of extra dependencies for mypy in `pre-commit-config.yaml`?

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+            kind=CompletionItemKind.Snippet,
+        ),
+    ]
+
+
+class NeedlsFeatures(LanguageFeature):
+    """Sphinx-Needs features support for the language server."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
danieleades, line 254:
```suggestion
    def __init__(self, rst: RstLanguageServer):
        super().__init__(rst)
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+
+        # if word starts with '..', complete_directive
+        if word.startswith(".."):
+            return complete_directive(self, context, lines, word)
+
+        return []
+
+    hover_triggers = [re.compile(r".*")]
+
+    def hover(self, context: HoverContext):
```
danieleades, line 295:
```suggestion
    def hover(self, context: HoverContext) -> str:
```

sphinx_needs/lsp/needs_store.py
```diff
@@ -0,0 +1,110 @@
...
+            return False
+
+        return True
+
+    def set_docs_root(self, docs_root: str) -> None:
+        if not os.path.exists(docs_root):
+            raise FileNotFoundError(f"Docs root directory not found: {docs_root}")
+        self.docs_root = docs_root
+
+    def set_conf_py(self, conf_py_path: str) -> None:
```
danieleades, line 41: would be better to use `pathlib.Path` instead of `str` everywhere that it represents a filepath.

sphinx_needs/lsp/needs_store.py
```diff
@@ -0,0 +1,110 @@
...
+
+class NeedsStore:
+    """Abstraction of needs database."""
+
+    def __init__(self):
+        self.docs_per_type = {}  # key: need type, val: list of doc names (str)
+        self.needs_per_doc = {}  # key: docname; val: list of needs
+        self.types = []  # list of need types actually defined in needs.json
+        self.declared_types = {}  # types declared in conf.py: {'need directive': 'need title'}
+        self.needs = {}
```
danieleades, line 23: prefer type annotations to comments where possible.

ie

```python
self.docs_per_type: Dict[str, List[str]] = {}
```

sphinx_needs/lsp/needs_store.py
```diff
@@ -0,0 +1,110 @@
...
+        self.needs_initialized = False
+        self.docs_root = None
+        self.conf_py_path = None
```
danieleades, line 26: these should have annotations too

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+            label=":need:",
+            detail="need role",
+            insert_text="need:`${1:ID}` $0",
+            insert_text_format=InsertTextFormat.Snippet,
+            kind=CompletionItemKind.Snippet,
+        ),
+    ]
+
+
+class NeedlsFeatures(LanguageFeature):
```
danieleades, line 250: class name should use PascalCase:

```suggestion
class NeedLsFeatures(LanguageFeature):
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+    hash_part = generate_hash(user_name, doc_uri, need_prefix, line_number)
+    need_id = need_prefix + "_" + hash_part
+    # re-generate hash if ID is already in use
+    while need_id in ls.needs_store.needs:
+        hash_part = generate_hash(user_name, doc_uri, need_prefix, line_number)
+        need_id = need_prefix + "_" + hash_part
+    return need_id
+
+
+def complete_directive(ls, params, lines: List[str], word: str):
```
danieleades, line 208:
```suggestion
def complete_directive(ls, params, lines: List[str], word: str) -> List[CompletionItem]:
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+        except AttributeError:
+            return "ID"
+
+    need_prefix = need_type.upper()
+    hash_part = generate_hash(user_name, doc_uri, need_prefix, line_number)
+    need_id = need_prefix + "_" + hash_part
+    # re-generate hash if ID is already in use
+    while need_id in ls.needs_store.needs:
+        hash_part = generate_hash(user_name, doc_uri, need_prefix, line_number)
+        need_id = need_prefix + "_" + hash_part
```
danieleades, line 204:
```suggestion
        need_id = f"{need_prefix}_{hash_part}"
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+    if not need_type:
+        try:
+            match = re.search(".. ([a-z]+)::", lines[line_number - 1])
+            need_type = match.group(1)
+        except AttributeError:
+            return "ID"
+
+    need_prefix = need_type.upper()
+    hash_part = generate_hash(user_name, doc_uri, need_prefix, line_number)
+    need_id = need_prefix + "_" + hash_part
```
danieleades, line 200:
```suggestion
    need_id = f"{need_prefix}_{hash_part}"
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+def generate_hash(user_name, doc_uri, need_prefix, line_number):
+    salt = os.urandom(blake2b.SALT_SIZE)  # pylint: disable=no-member
+    return blake2b(
+        f"{user_name}{doc_uri}{need_prefix}{line_number}".encode(),
+        digest_size=4,
+        salt=salt,
+    ).hexdigest()
+
+
+def generate_need_id(ls, params, lines: List[str], word: str, need_type: str = None) -> str:
```
danieleades, line 184: mypy 'strict' checking is enabled, so i'm not sure why it's not flagging this one... 'strict' should imply 'no implicit optionals', so the signature here should be:

```suggestion
def generate_need_id(ls, params, lines: List[str], word: str, need_type: Optional[str] = None) -> str:
```

also `ls`, and `params` are missing type annotations

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+    for sub_path in sub_paths:
+        if sub_path.find(".rst") > -1:
+            kind = CompletionItemKind.File
+        else:
+            kind = 19  # Folder
+        items.append(CompletionItem(label=sub_path, kind=kind, detail="path to needs doc"))
+    return items
+
+
+def complete_need_link(ls, params: CompletionContext, lines: List[str], line: str, word: str):
```
danieleades, line 127: return type?

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+    words = line.split()
+    index = col_to_word_index(col, words)
+    return words[index]
+
+
+def get_lines_and_word(ls, params) -> Tuple[List[str], str]:
+    return (get_lines(ls, params), get_word(ls, params))
+
+
+def get_need_type_and_id(ls, params) -> Tuple[str, str]:
```
danieleades, line 66:
```suggestion
def get_need_type_and_id(ls, params) -> Tuple[Optional[str], Optional[str]]:
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+                            new_text="",
+                        )
+                    ],
+                )
+                for need in ls.needs_store.needs_per_doc[requested_doc]
+                if need["type"] == requested_type
+            ]
+
+
+def generate_hash(user_name, doc_uri, need_prefix, line_number):
```
danieleades, line 175: type annotations?

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+    definition_triggers = [re.compile(r".*")]
+
+    def definition(self, context: DefinitionContext):
+        """Return location of definition of a need."""
+        # load needs.json
+        needs_json = os.path.join(self.rst.app.confdir, "_build/needs/needs.json")
+        self.needs_store.load_needs(needs_json)
+
+        if not self.needs_store.is_setup():
+            return
```
danieleades, line 327:
```suggestion
            return None
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+            description = self.needs_store.needs[need_id]["description"]
+            hover_value = f"**{title}**\n\n```\n{description}\n```"
+            return hover_value
+        except KeyError:
+            # need is not in the database
+            return ""
+
+    definition_triggers = [re.compile(r".*")]
+
+    def definition(self, context: DefinitionContext):
```
danieleades, line 320: what's the return type?

is it `Optional[List[Location]]`?

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,378 @@
...
+            return None
+
+        doc_path = os.path.join(self.rst.app.confdir, need["docname"])
+        if os.path.exists(doc_path + ".rst"):
+            doc_path = doc_path + ".rst"
+        elif os.path.exists(doc_path + ".rest"):
+            doc_path = doc_path + ".rest"
+        else:
+            return None
+        doc_uri = pathlib.Path(doc_path).as_uri()
```
danieleades, line 344: mixing `os.path` and `pathlib`

would be better to use `pathlib` throughout

sphinx_needs/lsp/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class NeedlsConfigException(BaseException):
```
danieleades, line 1: this should use PascalCase. also, pep8 says errors should have the suffix 'error' rather than 'exception'.

```suggestion
class NeedLsConfigError(BaseException):
```

pyproject.toml
```diff
@@ -95,6 +96,8 @@ module = [
   'sphinx_needs.functions.common',
   'sphinx_needs.functions.functions',
   'sphinx_needs.layout',
+  'sphinx_needs.lsp.esbonio',
+  'sphinx_needs.lsp.needs_store',
```
danieleades, line 100: new modules should not be added to the mypy whitelist. This is for pre-existing modules that are not (yet) "type-safe".

danwos:
Ohh I like the discussion here, as it marks the beginning of having type annotations in every code change from now on :)

Thanks @danieleades for all the comments.
I will go through them together with @haiyangToAI on Monday and as far as I admire your work, every change request will get implemented :+1:

.flake8
```diff
@@ -2,4 +2,4 @@
 max-line-length = 120
 per-file-ignores =
     tests/data/service_github.py: E501
```
danieleades, line 4: you could maybe remove the `per-file-ignores` line while you're here? it's redundant given that `E501` is in the 'extend-ignore' list

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,385 @@
...
+import os
+import pathlib
+import re
+from hashlib import blake2b
+from typing import List, Optional, Tuple, Union
+
+from esbonio.lsp import LanguageFeature
+from esbonio.lsp.rst import CompletionContext, DefinitionContext, HoverContext
+from esbonio.lsp.sphinx import SphinxLanguageServer
+from pygls.lsp.types import (
```
danieleades, line 12: i don't see the `pygls` dependency in `pyproject.toml`. Where is this getting pulled in?

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,385 @@
...
+    Location,
+    Position,
+    Range,
+    TextEdit,
+)
+
+from sphinx_needs.lsp.needs_store import NeedsStore
+
+
+class NeedlsFeatures(LanguageFeature):  # type: ignore[misc]
```
danieleades, line 25: this error occurs because mypy can't 'see' the `esbonio` types, since the esbonio package is not installed in the pre-commit environment.

There are two solutions

1. add the esbonio package to the mypy additional dependencies in `pre-commit-config.yaml`
2. modify the mypy pre-commit hook to be a 'local' hook, which is run from the dev environment, not a dedicated pre-commit environment.

I'll make a PR for option 2, since i think that's a more robust general solution.

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,385 @@
...
+    """Sphinx-Needs features support for the language server."""
+
+    def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
+        super().__init__(*args, **kwargs)
```
danieleades, line 29: is there any reason to use `args, **kwargs)` here? the arguments for `super().__init__(...)` are well-defined.

https://github.com/swyddfa/esbonio/blob/6d898fc6ad691bf1d1d63ec496c6c453b5699fa2/lib/esbonio/esbonio/lsp/rst/__init__.py#L227 

then you can remove the `type: ignore` comment here

sphinx_needs/lsp/esbonio.py
danieleades: i've created #588 with this change.

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,385 @@
...
+        index = index + 1
+    return index - 1
+
+
+def get_lines(ls: NeedlsFeatures, params: Union[CompletionContext, DefinitionContext, HoverContext]) -> List[str]:
+    """Get all text lines in the current document."""
+    text_doc = params.doc
+    ls.logger.debug(f"text_doc: {text_doc}")
+    source = text_doc.source
+    return source.splitlines()  # type: ignore[no-any-return]
```
danieleades, line 167: this is a false positive. not sure why it's occurring. can fix it with

```python
    ...
    lines: List[str] = source.splitlines()
    return lines
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,385 @@
...
+def get_word(ls: NeedlsFeatures, params: Union[CompletionContext, DefinitionContext, HoverContext]) -> str:
+    """Return the word in a line of text at a character position."""
+    line_no, col = params.position
+    lines = get_lines(ls, params)
+    if line_no >= len(lines):
+        return ""
+    line = lines[line_no]
+    words = line.split()
+    index = col_to_word_index(col, words)
+    return words[index]  # type: ignore[no-any-return]
```
danieleades, line 179: this is another odd false positive. Can be fixed with an additional annotation

```suggestion
def get_word(ls: NeedlsFeatures, params: Union[CompletionContext, DefinitionContext, HoverContext]) -> str:
    """Return the word in a line of text at a character position."""
    line_no, col = params.position
    lines = get_lines(ls, params)
    if line_no >= len(lines):
        return ""
    line = lines[line_no]
    words: List[str] = line.split()
    index = col_to_word_index(col, words)
    return words[index]
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,385 @@
...
+            if new_path not in sub_paths:
+                sub_paths.append(new_path)
+    sub_paths.sort()
+
+    items = []
+    for sub_path in sub_paths:
+        if sub_path.find(".rst") > -1:
+            kind = CompletionItemKind.File
+        else:
+            kind = 19  # Folder
```
danieleades, line 244: does this not work?

```suggestion
            kind = CompletionItemKind.Folder
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,385 @@
...
+    """Generate a need ID including hash suffix."""
+
+    user_name = getpass.getuser()
+    doc_uri = params.doc.uri
+    line_number = params.position.line
+
+    if not need_type:
+        try:
+            match = re.search(".. ([a-z]+)::", lines[line_number - 1])
+            need_type = match.group(1)  # type: ignore[union-attr]
```
danieleades, line 322: this type error is occurring because `re.search` can return `None`, and the `None` case is not handled.

if there some invariant in the code here which means that `re.search` will *never* return None? in that case you can using `typing.cast` to force it to be not `None`, and you should add a comment stating why this invariant holds.

If you can't guarantee through some invariant that the search will *always* return a value, then failing to handle the `None` case is a bug and mypy is correct to flag it.

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,385 @@
...
+        except AttributeError:
+            return "ID"
```
danieleades, line 324: perhaps this is a better construction-

```suggestion
        match = re.search(".. ([a-z]+)::", lines[line_number - 1])
        if match:
            need_type = match.group(1)
        else:
            return "ID"
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,385 @@
...
+
+    need_prefix = need_type.upper()  # type: ignore[union-attr]
```
danieleades, line 326: with the above suggested change, you can also remove this 'type: ignore' comment

sphinx_needs/lsp/needs_store.py
```diff
@@ -0,0 +1,111 @@
...
+        logging.info(f"Loading need_types from {conf_py_name}...")
+
+        spec = importlib.util.spec_from_file_location(module_name, conf_py_path)
+        if spec is None:
+            raise ImportError(f"Created module spec {spec} from {conf_py_name} not exists.")
+
+        module = importlib.util.module_from_spec(spec)
+        sys.modules[module_name] = module
+        try:
+            spec.loader.exec_module(module)  # type: ignore[union-attr]
```
danieleades, line 64: ideally this 'type: ignore' comment should be removed as it is obscuring a real potential bug.

`spec.loader` can be `None`, but this case is not explicitly handled in your code- hence the error.

sphinx_needs/lsp/esbonio.py
danwos: I agree, it should be part of the ``pyproject.toml``, but without any version pinning, as this is the job from the "parent"-dep, in this case ``esbonio``.

danwos (changes requested):
@haiyangToAI thanks for the update.
And @danieleades  thanks for the great review and the proposals how to fix most of the findings.

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,390 @@
...
+    Location,
+    Position,
+    Range,
+    TextEdit,
+)
+
+from sphinx_needs.lsp.needs_store import NeedsStore
+
+
+class NeedlsFeatures(LanguageFeature):
```
danieleades, line 25: I haven't got the faintest idea why these imports are causing type errors in mypy. Especially given pylance has no problem with them.

Nevertheless, these ignore statements will get us out of trouble for now

```suggestion
from esbonio.lsp import LanguageFeature  # type: ignore[attr-defined]
from esbonio.lsp.rst import (  # type: ignore[attr-defined]
    CompletionContext,
    DefinitionContext,
    HoverContext,
)
from esbonio.lsp.sphinx import SphinxLanguageServer  # type: ignore[attr-defined]
from pygls.lsp.types import (
    CompletionItem,
    CompletionItemKind,
    InsertTextFormat,
    Location,
    Position,
    Range,
    TextEdit,
)

from sphinx_needs.lsp.needs_store import NeedsStore


class NeedLsFeatures(LanguageFeature):  # type: ignore[misc]
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -0,0 +1,390 @@
...
+
+def get_need_type_and_id(
+    ls: NeedlsFeatures, params: Union[DefinitionContext, HoverContext]
+) -> Tuple[Optional[str], Optional[str]]:
+    """Return tuple (need_type, need_id) for a given document position."""
+    word = get_word(ls, params)
+    for need in ls.needs_store.needs.values():
+        if need["id"] in word:
+            return (need["type"], need["id"])
+    return (None, None)
```
danieleades, line 196: would this not be better returning a `Optional[Tuple[str, str]]`?

as in-

```suggestion
def get_need_type_and_id(
    ls: NeedlsFeatures, params: Union[DefinitionContext, HoverContext]
) -> Optional[Tuple[str, str]]:
    """Return tuple (need_type, need_id) for a given document position."""
    word = get_word(ls, params)
    for need in ls.needs_store.needs.values():
        if need["id"] in word:
            return (need["type"], need["id"])
    return None
```

obviously you'll need to cascade that change everywhere it's called, but i think it might make your null-checks a little tidier

haiyangToAI: Don't agree. This function needs to return a Tuple (None, None), not just None. Otherwise, it will make the code in other place where use this function more complicated.

danieleades: If you say so!

I'm curious which bits will become more complex?

The reason I suggest using Optional[Tuple] is that the 'optional' monad never needs to be distributed over the tuple elements. That is, you *never* have [Some, None] or [None, Some], you only have [Some, Some] or [None, None]. That is, the type you've chosen has invalid variants

danwos:
For me this PR is fine and the functions are well developed.
Small findings regarding naming and type annotations can be fixed with another PR.
No need to be too perfect here :)

Thanks for the work. 
I'll will write some additional docs for it in the next days.

PR #586, opened 2022-05-26
add 'sphinx' package to mypy context

danieleades:
adding the `sphinx` package to the mypy context used by pre-commit significantly improves the type inference. This removes some errors, and adds a whole bunch of new ones.

have added an 'unwrap' method (modelled after Rust's approach) to coerce `Optional` types which are known to be not `None`


danieleades:
changing to draft. this PR is effectively replaced by #588, so they shouldn't both be merged.

PR #587, opened 2022-06-01
fix poetry lock file

danieleades:
I noticed that the poetry lock file was not quite aligned with the pyproject.toml file.

This PR fixes that misalignment, and adds a pre-commit hook to prevent it from happening again


danieleades:
@danwos is the 'link check' job actually failing because of changes in this PR or is it just a little flaky?

PR #592, opened 2022-06-01
Add docu for needextend default option values

arwedus:
resolves useblocks/sphinxcontrib-needs#313

@danwos: I also documented the "docname" option now, as the use case will often require using it. Needextend was also prominently missing from the "list of directives that support filters" ;-)


danwos (approved):
Perfect, thanks :+1:

And congratulation to your first PR in this project :1st_place_medal:

PR #595, opened 2022-06-02
Adds sphinx-plantuml as dev dep

danwos:


danieleades:
should `tabulate` also be added?

PR #599, opened 2022-06-07
flatten an unnecessary exception

danieleades:
this exception is immediately caught and handled, so might as well be a simple if statement.

@danwos i'm not sure this is reachable anyway, since `sphinxcontrib.plantuml` is already imported right at the top of the file. If the extension wasn't available in the build environment, you'd have already thrown an `ImportError` by this point.

This exception can only fire if you have installed the extension, but haven't included it in your config.

What are the intended semantics?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,6 +53,8 @@ jobs:
       - name: Install Poetry
         run: |
           curl -sL https://install.python-poetry.org | python -
+      - name: Update Pip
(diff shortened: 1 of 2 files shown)


danwos:
You are right, as PlantUML gets already imported, the later check makes no sense.

But as PlantUML is currently no dependency of Sphinx-Needs, we also need to secure this import or make it a dependency.

And sorry for my late response. Totally missed this.

danwos:
Already solved or the situation has changed.

PR #605, opened 2022-06-11
Handle overlapping labels in needpie

### LINKED ISSUE #498: Piecharts with overlapping labels
With contents containing no data (or datasets where one content is a lot bigger than the others) , the labels of a `needpie::` are overlapping.

PhilipPartsch:
fix #498


PhilipPartsch:
Looks like that the useblocks server does not answer to the Linkcheck script executed by the pipeline. Maybe the useblocks server assumes wrongly an attack?

danwos:
> Looks like that the useblocks server does not answer to the Linkcheck script executed by the pipeline. Maybe the useblocks server assumes wrongly an attack?

Yes I know. It is a bug by our provider and ticket is open 
Therefore the link check is currently not needed for a merge.

danwos (approved):

Issue #581, opened 2022-05-24
Check warnings/constraints during earlier build phase


danwos:
**Goals**

1. Check constraints during build
2. Highlight related need with check_status (e.g. red border if not passed)
3. Overview of needs with specific check status
4. Stop the build or keep it running based on check-results (During need build or at the end  
   Based on the given config, e.g. local=keep it running, release = stop

**Technical Requirements**

- Perform check during need-creation (event + process_need_constraints)
- Have a global dict of check results
- Register checks somehow via config
- Map checks to specific need objects (e.g. all need of type X shall run checks A,B,D)
- Define needed checks via need-option (e.g :constraints: list)

**Spec**

- New global option “constraints”.
  - “constraints” take a list of names
- new config options “need_constraints”
  - takes a dict
  - key = name of the constraint (can be used in the “constraints” option)
  - item is a dict
    - key is the name of the constraint-element
    - item is  a filter-string (see need_warnings)
      - maybe extra info can be added to define
        - the color / weight of this check
         - needed behavior (e.g. stop build)
            - Maybe should be defined somewhere else to keep constraints-config more shareable between projects

**Example**

RST-file

```rst
.. need:: My REQ
   :id: REQ_001
   :constraints: PROC; Q-Gate, TEAM  
```

conf.py

```py
needs_constraints = {
    "PROC": {
      "Check_1": "type=='REQ' and status in ['open', 'done']"
      "Check_X": ....
    }
    ...
}
```

Hint: Check if this concept can be an update of need_warnings!

PR #607, opened 2022-06-13
Adds constraints option

mawieland:
implements #581


danwos (approved):
Fixed some minor type annotation stuff on my own.

The rest looks all good. But I have to test it in action, so I will merge and start using it :) 

Thanks for the work :+1:

PR #609, opened 2022-06-13
[WIP] fix some more type errors

danieleades:


danieleades:
@danwos i just wanted to check some assumptions with you, and make sure i'm on the right track.

In the directives code (i'm currently looking at "needbar", the `option_spec` on the directive makes use of the `directives.unchanged_required` parser. This parser means that the values *must* be present.

There are then null checks for each value as they are indexed from `self.options`. Am i right in thinking those null checks can be removed, since the `option_spec` parser will ensure the value is never None?

danwos:
I would like to close this PR, as nobody is working on it and it is out of any with master.
Any concerns?

Looks like the work stopped as I haven't given any feedback :( Sorry!
Any chance to work on it again, if I'm more responsive?

PR #611, opened 2022-06-13
needref customizable link text

PhilipPartsch:
try to implement  #439


PhilipPartsch:
Can some body help with the error from Lint?

sphinx_needs/roles/need_ref.py
```diff
@@ -2,16 +2,53 @@
...
+
+from typing import Dict
+
 log = get_logger(__name__)
 
 
 class NeedRef(nodes.Inline, nodes.Element):
     pass
 
+def transform_need_to_dict(need: Need) -> Dict[str, str]:
```
PhilipPartsch, line 20: Do we maybe want to shift this function to a more generic place? I belive there is a reuse in other algorithms, too

danwos (changes requested):
Great PR.
I only have some concerns regarding the string-algorithm.

I would like to use jinja and maybe configure it the way we need (e.g. replace the pre/postfix characters ``{{..}}`` --> ``[[...]]``)

sphinx_needs/roles/need_ref.py
PhilipPartsch: A need is a dictionary, but the value is sometimes a dictionary, list, ... . It is not nice to see this python syntax in teh output, so I translate a need in a dictionary of str, str. Why do I need a dictionary of str, str. This is the input to a str.format() function.

sphinx_needs/roles/need_ref.py
```diff
@@ -61,19 +102,19 @@ def process_need_ref(app: Sphinx, doctree: nodes.document, fromdocname: str) ->
...
-                    links=";".join(target_need["links"]),
-                    links_back=";".join(target_need["links_back"]),
-                )
+                dict_need = transform_need_to_dict(target_need)
+                # we set title and id, after the transform_need_to_dict to overwrite the values
+                dict_need["title"] = title
+                dict_need["id"] = ref_id_complete
+
+                if ref_name and prefix in ref_name:
+                    # if ref_name is set and has prefix to process, we will do so
```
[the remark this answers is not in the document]
PhilipPartsch: In a discussion we made the discision to give in "[[" as prefix and "]]" as post fix. So theses have to be ranslated to "[[" => "{" and "]]" => "}" to forward the string later to the str.format() function

sphinx_needs/roles/need_ref.py
danwos: In which cases do we have a `dict` as values of a need option?
And even if this is the case, are these values needed in a `need` representation?

We use jinja for all string-manipulations in Sphinx-Needs. And it would give the user more power, e.g. cutting long strings.
So I would like to stay with one, tested solution in the code and try to avoid a new one.

sphinx_needs/roles/need_ref.py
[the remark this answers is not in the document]
danwos: Ahh ok, got it. Thanks.

sphinx_needs/roles/need_ref.py
PhilipPartsch: I'm not sure if we talk about the same. In the current implementation of needref is this [line](https://github.com/useblocks/sphinxcontrib-needs/blob/master/sphinx_needs/roles/need_ref.py#L64).
`                link_text = app.config.needs_role_need_template.format(
                    title=title,
                    id=ref_id_complete,
                    type=target_need["type"],
                    type_name=target_need["type_name"],
                    status=target_need["status"],
                    content=target_need["content"],
                    tags=";".join(target_need["tags"]),
                    links=";".join(target_need["links"]),
                    links_back=";".join(target_need["links_back"]),
                )`
What I did:
Instead of doing the mapping everytime explicit (with maintance effort on all places, if a new option is added), I create the function `def transform_need_to_dict(need: Need) -> Dict[str, str]:` which is doing the mapping and use the current format function to create the string. See:
link_text = app.config.needs_role_need_template.**format**(

I'm not a jinja expert and so I cannot do the needed refactoring of the current needref to reflect all string manipulation with Jinja code.

danwos: Argh I didn't know that this part is so "old" and is not using jinja implementation.
And before we introduce here a new way to use need-data inside strings, I would like to go directly with jinja.

There are several locations in the code, where it gets used. But in general this is how it works:
```python
from jinja2 import Template
ref_template = Template(ref_string, autoescape=True)
final_ref_name = content_template.render(**need_info)
 ```
So I think this should even reduce the lines of code and make really all the need-data available.
What do you think?

File: docs/roles.rst [hunk 3]
```diff
@@ -34,16 +39,29 @@ Example
...
    :ref:`needs_role_need_max_title_length`.
+   If we find a ``[[`` in the customized string, we handle it 
+   according to Python's ``{`` `.format() <https://docs.python.org/3.4/library/functions.html#format>`_ 
+   function.
+   Please see https://pyformat.info/ for more information.
+   RST-attributes like ``**bold**`` are **not** supported.
+
+.. warning::
+
+   If your refer to an :ref:`external need <needs_external_needs>`, the algorithm is different
```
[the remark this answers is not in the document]
PhilipPartsch: Resolved

danwos (approved):
Looks great, thanks for the work and the tough discussion :) :+1:

Issue #612, opened 2022-06-13
customizable grid layouts


PhilipPartsch:
I have a few use cases, there I have to have other gird layouts for a need.
I thougth about a more sophisticated grid layout configuration.
How is the interest about this?
Who could test it with me and help me with reviews?

PR #614, opened 2022-06-16
initial customizable grids

PhilipPartsch:
try to implement #612

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -285,6 +285,38 @@ needs_warnings = {
 needs_default_layout = "clean"
 needs_default_style = None
 
+needs_grids = {
+    "customer_grid3": {
+        "widths": "auto" ,
+        "layout": {
+            "head":    {"head_left": {},    "head_center": {},    "head_right": {}    },
+            "meta":    {"meta_left": {},    "meta_center": {},    "meta_right": {}    },
+            "mid":     {"mid_left": {},     "mid_center": {},     "mid_right": {}     },
+            "footer":  {"footer_left": {},  "footer_center": {},  "footer_right": {}  },
+            "sfooter": {"sfooter_left": {}, "sfooter_center": {}, "sfooter_right": {} },
+        },
+    },
+    "customer_grid": {
+        "widths": "auto" ,
+        "layout": {
+            "head":   {"head_left": {},                        "head_right": {}   },
+            "meta":   {"meta_left": {},   "meta_center": {},   "meta_right": {}   },
+            "mid":    {"mid_left": {},    "mid_center": {},    "mid_right": {}    },
+            "footer": {"footer_left": {}, "footer_center": {}, "footer_right": {} },
+            "sub_footer": {"sub_footer_left": {}, "sub_footer_center": {}, "sub_footer_right": {} },
+        },
+    },
+    "customer_grid2": {
+        "widths": [20, 30, 30, 20,],
+        "layout": {
+            "head":   {"head_left":   {'colspan': 1,}, "head_center":  {'colspan': 1,},              "head_center2": {'colspan': 1,}, "head_right":   {'colspan': 1,}, },
+            "meta":   {"meta_left":   {'colspan': 1,}, "meta_center":  {'colspan': 1,},              "meta_center2": {'colspan': 1,}, "meta_right":   {'colspan': 1,}, },
+            "mid":    {"mid_left":    {'colspan': 1,}, "content":   {'colspan': 2, 'rowspan': 2,},                                 "mid_right":    {'colspan': 1,}, },
+            "footer": {"footer_left": {'colspan': 1,},                                                                                "footer_right": {'colspan': 1,},  },
+        },
+    },
+}
+
 needs_layouts = {
     "example": {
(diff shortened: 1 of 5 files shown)


sphinx_needs/layout.py
```diff
@@ -354,8 +373,11 @@ def _func_replace(self, section_nodes):
                     func_def = line_element[0]
                     # Check if normal string was detected
                     if len(text) > 0 and len(func_def) == 0:
-                        node_line += nodes.Text(text, text)
-                        result = nodes.Text(text, text)
+                        if len(section_nodes) == 1:
+                            node_line = nodes.Text(text, text)
+                        else:
+                            node_line += nodes.Text(text, text)
+                        #result = nodes.Text(text, text)#dead code?
```
PhilipPartsch, line 380: I belive it is dead code and can be removed.

sphinx_needs/layout.py
```diff
@@ -297,6 +311,10 @@ def get_section(self, section: str) -> Optional[nodes.line_block]:
         # Needed for PDF/Latex output, where empty line_blocks raise exceptions during build
         if len(lines) == 0:
             return None
+        elif len(lines) == 1:
+            line_parsed = self._parse(lines[0])
+            line_ready = self._func_replace(line_parsed)
+            return line_ready
```
PhilipPartsch, line 317: Here we now return in element with out lines_container (block). This leads to missing <span> elements in the ouput. Is this an issue?

sphinx_needs/layout.py
```diff
@@ -407,14 +429,21 @@ def _func_replace(self, section_nodes):
...
+                                node_line += result
                     else:
                         raise SphinxNeedLayoutException(
                             f"Error during layout line parsing. This looks strange: {line_element}"
                         )
 
                 return_nodes.append(node_line)
-        return return_nodes
+
+        if len(section_nodes) == 1:
```
PhilipPartsch, line 443: Ugly, but currently functional. I can refactor it, if we have an working solution.

sphinx_needs/layout.py
```diff
@@ -1190,6 +1231,32 @@ def _grid_content(self, colwidths, side_left, side_right, footer):
...
+        for width in customer_format["widths"]:
+            node_colspec = nodes.colspec(colwidth = width)
+            node_tgroup += node_colspec
+
+        for r_k, r_v in customer_format["layout"].items():
+            row = nodes.row(classes=[r_k])
+            self.node_tbody += row
+
+            for c_k, c_v in r_v.items():
+                if "colspan" in c_v and "rowspan" in c_v:
```
PhilipPartsch, line 1247: At first I had a for loop to add all elemnts, but I have to decrease these by one, so it is easier understable for the "user". Overall I have to think about, what is better:
- forward everything, without checking
- copy the know attributes

PhilipPartsch:
Tests are failing, because they have to be adpated, becasue of breaking change.

sphinx_needs/layout.py
```diff
@@ -134,6 +134,11 @@ def build_need(layout, node, app: Sphinx, style=None, fromdocname: Optional[str]
     # We need to replace the current need-node (containing content only) with our new table need node.
     node.parent.replace(node, new_need_node)
 
+# read configuration
+# ensure dict of array of arrays of tuble
+def getgridsfromconfig(app: Sphinx,):# -> dict:
```
danieleades, line 139: this should use proper type annotations, not a type comment

sphinx_needs/layout.py
```diff
@@ -407,14 +429,21 @@ def _func_replace(self, section_nodes):
...
 
                 return_nodes.append(node_line)
-        return return_nodes
+
+        if len(section_nodes) == 1:
+            return return_nodes[0]
+        else:
+            return return_nodes
 
     def _replace_place_holder(self, data):
```
danieleades, line 448: type annotations?

sphinx_needs/layout.py
```diff
@@ -956,6 +985,18 @@ def permalink(
             prefix=prefix,
         )
 
+    def content(self):
```
danieleades, line 988: type annotations?

sphinx_needs/needs.py
```diff
@@ -454,6 +455,129 @@ def visitor_dummy(*_args, **_kwargs) -> None:
     """
     pass
 
+#todo: test isattributeset
+def isattributeset(grid: {}, attributes: []) -> bool:
```
danieleades, line 459:
```suggestion
def is_attribute_set(grid: {}, attributes: []) -> bool:
```

also, missing type annotations

danieleades (changes requested):
use snake case for function names. A few missing type annotations as well

docs/conf.py
```diff
@@ -285,6 +285,38 @@
 needs_default_layout = "clean"
 needs_default_style = None
 
+needs_grids = {
```
danieleades, line 288: these are really complex, stringly-typed dictionaries. Is it possible to replace any of the dicts used in this PR with classes?

danwos:
I would like to close this PR, as nobody is working on it and it is out of any with master.
Any concerns?

PR #618, opened 2022-06-22
CI fix for old setuptool deps

danwos:


danwos:
Not needed, as already implemented by #608

PR #620, opened 2022-06-23
remove some potentially unused objects

danieleades:
removing dead code.

I'm abusing the CI to test this incrementally as i strip code out


danieleades:
there are also unused args in the `api.add_need` function. These should probably have been removed before the move to 1.0, along with any deprecated directives. I've excluded any API changes from this PR for now though

danwos (approved):
Thanks for cleaning up the code :+1:

PR #622, opened 2022-06-23
Improve language features docs

haiyangToAI:


docs/IDE/index.rst
```diff
@@ -3,18 +3,14 @@
 IDE Support
 ===========
 
-Language features
------------------
+The following supported sphinx needs language features are intergrated into `Esbonio language server <https://github.com/swyddfa/esbonio>`_.
```
danwos, line 6: I think we need some docs for IDE extension installation too.
Otherwise it is not clear what the user shall do.

Can you add a chapter `Installation` and a sub-chapter `VsCode`?
Then 1-2 lines of docs for installing the needed RST-Plugin and get it running for Sphinx-Needs.

Document only the Sphinx-Need-specific part. Everything else shall be just a link to the original extension docs

danwos (changes requested):
Good docs, 1-2 findings only.

Can we rename the folder from `IDE` to `ide`?

docs/ide/index.rst
```diff
@@ -1,16 +1,25 @@
...
 
    Currently as long as needs.json exists under {confdir}/_build/needs in your sphinx needs project, all the features are available
    if your IDE extension or plugin supports `Esbonio language server <https://github.com/swyddfa/esbonio>`_.
 
+
+To use all the Sphinx-Needs language featues,
+
+* Install IDE extension or plugin, see current supported IDE extension in :ref:`ide_installation`.
+
+* Build needs.json file under path {confdir}/_build/needs in your Sphinx-Needs project, see how to build needs.json in :ref:`builders`.
```
danwos, line 18: Put paths this between 2 ` chars

docs/ide/installation.rst
```diff
@@ -0,0 +1,14 @@
+.. _ide_installation:
```
danwos, line 1: I think we do not need an extra file for installation.
I would put it all inside `ide/index.rst`

docs/ide/installation.rst
```diff
@@ -0,0 +1,14 @@
...
+Installation
+============
+
+VsCode
+------
+
+The VsCode extension `reStructuredText <https://github.com/vscode-restructuredtext/vscode-restructuredtext>`_ supports all the Sphinx-Needs
+language features and is available at `Visual Studio Marketplace <https://marketplace.visualstudio.com/items?itemName=lextudio.restructuredtext>`_.
+
+To install and config this extension, see details in
```
danwos, line 12: config = configure

docs/ide/features.rst
```diff
@@ -0,0 +1,93 @@
+Language features
```
danwos, line 1: I think this should be part of `index.rst`

Is not soo much content, so need for extra files.

docs/ide/index.rst
```diff
@@ -0,0 +1,25 @@
+.. _ide:
+
+IDE Support
+===========
+
+The following supported Sphinx-Needs language features are intergrated into `Esbonio language server <https://github.com/swyddfa/esbonio>`_.
```
danwos, line 6: `The following ... features`? Where are they?
I would think a list is following...

docs/ide/index.rst
```diff
@@ -0,0 +1,25 @@
...
+
+.. note::
+
+   Currently as long as needs.json exists under {confdir}/_build/needs in your sphinx needs project, all the features are available
+   if your IDE extension or plugin supports `Esbonio language server <https://github.com/swyddfa/esbonio>`_.
+
+
+To use all the Sphinx-Needs language featues,
```
danwos, line 14: Maybe this can be partf of an extra chapter "Usage".

So in the end we would have:
- Installation
- Usage
- Features (maybe this at top as selling argument)

danwos (changes requested):
Nice docs, but I think we should put all the information together on a single page.

danwos (changes requested):
Thanks for the refactoring.

I think some parts should be moved from "Features" to usage so that "Features" are short and more selling it instead of explaining the details.

PR #627, opened 2022-07-01
Added test cases for lsp features

haiyangToAI:


tests/test_lsp/test_lsp.py
```diff
@@ -0,0 +1,227 @@
+"""Test Sphinx-Needs language features."""
+
+import os
+import sys
+
+import pytest
+import pytest_lsp
+from pygls.lsp.types import MarkupContent, MarkupKind, Position, Range
+from pytest_lsp import ClientServerConfig
```
danwos, line 9: `pytest_lsp` got already imported.

So use `pytest_lsp.ClientServerConfig` in your code

danwos (commented):
wow, these tests are looking very promising.

Like the way, it gets tested via `client` class in pytest_lsp.

danwos (approved):

PR #630, opened 2022-07-04
Added open-needs service

iSOLveIT:


docs/conf.py
```diff
@@ -355,6 +373,10 @@
 # build needs.json to make permalinks work
 needs_build_json = True
 
+# Config Variables for Open Needs service
+open_needs_test_env = False  # Set to True to use patched data instead of using Open-Needs server when testing
```
danwos, line 377: Don't like the solution, as this option is normally not needed by any user.

It would be better to activate the service-call via a rst-directive, if certain values are set.

I think we can reuse here the same mechanism as for codebeamer in the sphinx-needs-enterprise package.

danwos (changes requested):
Great work. We only should try to find a better way for test-handling

PR #635, opened 2022-07-07
Improvement: Added open-needs service

iSOLveIT:
The open-needs service will allow users to read/import `need` objects from an Open-Needs Server.


docs/roles.rst
```diff
@@ -31,6 +31,8 @@ Example
    :value: 20
    :unit: slices
 
+.. _needref:
```
danwos, line 34: Does this make sense here?

If yes, then a second ``:`` is missing at the end.

danwos (changes requested):
Really great work and good test cases (incl. the monkeypatch part).
Only a small finding in the docs.

Overall next time pls try to separate docs-enhancements and feature development :)
It's always hard to make reviews of huge PRs.

docs/roles.rst
iSOLveIT: This

> Does this make sense here?
> 
> If yes, then a second `:` is missing at the end.

Yes, it makes sense. It creates a link to that section so a second `:` is not needed.

iSOLveIT:
> Really great work and good test cases (incl. the monkeypatch part). Only a small finding in the docs.
> 
> Overall next time pls try to separate docs-enhancements and feature development :) It's always hard to make reviews of huge PRs.

The documentation hasn't been added yet. The current content is just to test out the output for the feature.
I am now writing the complete docs for the feature.

docs/roles.rst
danwos: Argh, sure. My fault.
Was thinking on a directive and not the basic sphinx reference feature.
Sorry :)

danwos (approved):

PR #646, opened 2022-07-15
Added option key for needuml

haiyangToAI:


docs/directives/needuml.rst
```diff
@@ -173,6 +227,32 @@ Please read :ref:`need_diagram` for details.
    {{uml("COMP_001")}}
    {{uml("FEATURE_1")}}
 
+Key argument
+++++++++++++
+
+:ref:`uml() <jinja_uml>` supports ``key`` argument to define which PlantUML code to load from the Sphinx-Need object.
+``key`` value by default is ``diagram``. If no key argument given, then load the PlantUML code from ``diagram``.
```
danwos, line 234: If no key argument given, then the PlantUML code is loaded from ``diagram``.

sphinx_needs/api/need.py
```diff
@@ -412,16 +412,31 @@ def run():
...
+    node_need_needumls_key_names = []
     for child in node_need_content.children:
         if isinstance(child, Needuml):
             needuml_id = child.rawsource
             try:
                 needuml = env.needs_all_needumls.get(needuml_id)
-                needs_info["diagram"] = needuml["content"]
+                if needuml["key"]:
+                    key_name = needuml["key"]
+                    if key_name not in node_need_needumls_key_names:
```
danwos, line 425: MAybe we should check, if `key` is already set in the need or if it is defined in needs_extra_options or needs_extra_links.

In all cases, this is an error, as we overwrite some stuff.

danwos (changes requested):

danwos (approved):

Issue #535, opened 2022-05-05
Enhance needtable to also support needs_string_links


StephanKuempel:
Render needs_string_links enhanced options also as a clickable link in needtable colums.
Currently it is just rendered as text.

danwos:
Solved by #652

PR #652, opened 2022-07-22
Improvement: Support of needs_string_links in needtable.

iSOLveIT:
- [x] Render ``needs_string_links`` enhanced options also as a clickable link in ``needtable`` columns. Fix for #535. 
- [x] Support for assigning multiple values to a needs_string_links option.


tests/test_needtable.py
```diff
@@ -76,6 +76,32 @@ def test_doc_needtable_options(test_app):
 
     assert column_order in html
 
+    # Checks for needs_string_links in needtable
+    assert "EXAMPLE_STRING_LINKS" in html
+    assert (
+        '<a class="reference external" href="https://github.com/useblocks/sphinxcontrib-needs/issues/404">'
+        "GitHub #404</a>" in html
```
danwos, line 83: I miss the check for the second value `303` here.

danwos (changes requested):
Really great work.
Can you add also a changelog.rst entry for this change?

And I think the tests for the new "additional value" needs some minor additions.

tests/test_needtable.py
iSOLveIT: > I miss the check for the second value `303` here.

Changes have been effected.

sphinx_needs/utils.py
```diff
@@ -161,6 +186,29 @@ def row_col_maker(
                 else:
                     ref_col.append(text_col)
                     para_col += ref_col
+            elif matching_link_confs:
+                try:
+                    link_name = None
```
PhilipPartsch, line 191: Hi @iSOLveIT,
Really cool Idea and implementation.
After a first look, the code here and the code in layout.py codeline 482 mostly look the same.
Is it possible to define a function once and reuse ist in both places?

iSOLveIT: Yeah I realised that but I will have to check that out properly. Anyway, thank you for the suggestion.

iSOLveIT: @PhilipPartsch 
I have implemented the suggestions in recent push changes.
Thanks again.

danwos (approved):
Thanks for the PR and the good work.
Can't wait to see the results in some more complex projects and their tables :)

PR #657, opened 2022-07-26
Added option save for needuml

haiyangToAI:


docs/directives/needuml.rst
```diff
@@ -152,6 +152,39 @@ and can't be ``diagram``.
       B -> C: Hi
       C -> B: Hi there
 
+save
+~~~~
+
+Allows to store generated Plantuml-code of ``needuml`` to given file path. The given file path must be
+relative path to ``conf.py`` directory. For example, ``_build/_out/my_needuml.puml``.
```
PhilipPartsch, line 159: Hello @haiyangToAI,
Really cool Feature and implementation.
I have my doubts to store the puml relative to the conf.py. I believe ist shall be relative to the out folder. 
What do you think?
Nice regards

danwos: For me it's hard to answer this question, as I see use cases and drawbacks for both locations.

My first thought was, that storage should be independent of the build location.
So  a ``make html`` stores the file at the same location as a ``make json``. Maybe somewhere on the root folder of the project, outside the docs folder.

But then I thought, what if it shall be part for the build artifact. Then this will never work.

Another question was: Do we really need to store the plantuml-output for each doc-build (and overwrite existing files)?

I came to the conclusion, that may be a special sphinx-builder would be better. For instance: ``make needumls``.
This then creates all puml files under ``_build/needumls``.

Maybe combined with a new config option `needs_build_needumls`, which takes a path, which is relative to the build-location. So it can be triggered  during a normal ``make html`` as well (similar to creating the needs.json during a HTML build.)

@PhilipPartsch: What do you think?

@haiyangToAI: Sorry for adding new ideas to this PR :/

PhilipPartsch: I really like the Idea of dedicated builder would be helpful and clean.

haiyangToAI: I also like the idea. I will adapt the implementation. No worries.

docs/builders.rst
```diff
@@ -137,3 +137,24 @@ Format
...
 
+.. _needumls_builder:
+
+needumls
+--------
+
+The **needumls** builder saves each :ref:`needuml` generated plantuml code to a file, and stores all the files into a single folder during the build.
+The file is created only if the option ``:save:`` from :ref:`needuml` is configured.
+
+The build creates a folder called **needumls** and files that store the generated plantuml code from :ref:`needuml` inside the given build-folder, e.g. `_build/needumls`.
```
danwos, line 148: Rephrase this line and maybe make it simpler? E.g.:
The build creates a folder called **needumls** inside the build folder, for instance _build/needumls.

docs/builders.rst
```diff
@@ -137,3 +137,24 @@ Format
...
+
+Usage
++++++
+
+.. code-block:: bash
+
+    sphinx-build -M needumls source_dir build_dir
```
danwos, line 155: mention also "make needumls" here. Maybe as first option, as it is simpler :)

docs/builders.rst
```diff
@@ -137,3 +137,24 @@ Format
...
+
+.. hint::
+
+    As an alternative, you can set the :ref:`needs_build_needumls` parameter in **conf.py** to create a folder for saved plantuml files
```
danwos, line 159: Rephrase:
As an alternative, you can set the config option :ref:`needs_build_needumls` to export the needumls files during each build.

docs/changelog.rst
```diff
@@ -21,6 +21,8 @@ Under development
   (`#659 <https://github.com/useblocks/sphinx-needs/issues/659>`_)
 * Bugfix: Fixed :ref:`needextract` filter options issue involved with :ref:`need_part`.
   (`#651 <https://github.com/useblocks/sphinx-needs/issues/651>`_)
+* Improvement: Added `save` option for :ref:`needuml`.
+* Improvement: Added builder :ref:`needumls_builder` and config option `needs_build_needumls` in `conf.py`.
```
danwos, line 25: :ref:``needs_build_needumls`

ref missing

docs/configuration.rst
```diff
@@ -1663,6 +1663,26 @@ Example:
    The created ``needs.json`` file gets stored in the ``outdir`` of the current builder.
    So if ``html`` is used as builder, the final location is e.g. ``_build/html/needs.json``.
 
+.. _needs_build_needumls:
+
+needs_build_needumls
+~~~~~~~~~~~~~~~~~~~~
+
+Creates a folder during other builds, like ``html``.
```
danwos, line 1671: Rephrase:
Export needuml data during each build.

docs/configuration.rst
```diff
@@ -1663,6 +1663,26 @@ Example:
...
+
+This option like :ref:`needs_build_json` allows to have one single Sphinx-Build for two output formats.
```
danwos, line 1673: This option works like :ref:`needs_build_json`.

Rest of the sentence can be deleted, as it is written already above.
dont' mention that there are two output format,s that sound confusing.
Just write, what it does. It exports needuml data.

docs/configuration.rst
```diff
@@ -1663,6 +1663,26 @@ Example:
...
+.. _needs_build_needumls:
+
+needs_build_needumls
+~~~~~~~~~~~~~~~~~~~~
+
+Creates a folder during other builds, like ``html``.
+
+This option like :ref:`needs_build_json` allows to have one single Sphinx-Build for two output formats. 
+
+All the files that stores the generated plantuml code from :ref:`needuml` will be saved in this created folder
```
danwos, line 1675: You should mention what kind of data this option can take above.

First always describe how the user needs to configure something.
Then explain what it does (max. one sentence for the function as introduction as the beginning.)
I was confused first, has I thought it gets a boolean value.

danwos: And keep the sentences more simple.
It's hard to follow the parts with "that"

docs/configuration.rst
```diff
@@ -1663,6 +1663,26 @@ Example:
...
+in the ``outdir`` of the current builder, e.g. ``_build/html/my_needumls/`` 
+
+This option takes relative folder path as input. Default value is: ``""``.
```
danwos, line 1678: This sentence should be near the beginning

docs/configuration.rst
```diff
@@ -1663,6 +1663,26 @@ Example:
...
+
+needs_build_needumls
+~~~~~~~~~~~~~~~~~~~~
+
+Creates a folder during other builds, like ``html``.
+
+This option like :ref:`needs_build_json` allows to have one single Sphinx-Build for two output formats. 
+
+All the files that stores the generated plantuml code from :ref:`needuml` will be saved in this created folder
+in the ``outdir`` of the current builder, e.g. ``_build/html/my_needumls/``
```
danwos, line 1676: If you use "my_needumls" here, you should show the example first, as it tells the user that "my_needumls" is not a hard configured name but can be configured by the user.

docs/directives/needuml.rst
danwos, line 159: Is this still valid?

I thought it is relative to the current builder output folder, e.g _build/needumls

docs/directives/needuml.rst
```diff
@@ -152,6 +152,39 @@ and can't be ``diagram``.
...
+
+If given file path folder not exists, it will be created. If given file path already exists, it will be
```
danwos, line 161: ...does not exist

sphinx_needs/builder.py
```diff
@@ -92,3 +92,58 @@ def build_needs_json(app: Sphinx, _exception: Exception) -> None:
...
+
+    def write_doc(self, docname: str, doctree: nodes.document) -> None:
+        pass
+
+    def finish(self) -> None:
+        env = unwrap(self.env)
+        needumls = env.needs_all_needumls.values()
+
+        for needuml in needumls:
+            if needuml["save"]:
```
danwos, line 108: Maybe some out put to the user would be nice.
E.g. A line like "Storing needuml to {file}"

sphinx_needs/directives/needuml.py
```diff
@@ -68,6 +69,14 @@ def run(self) -> Sequence[nodes.Node]:
         if key_name == "diagram":
             raise NeedumlException(f"Needuml option key name can't be: {key_name}")
 
+        save_path = self.options.get("save", None)
+        plantuml_code_out_path = None
+        if save_path:
+            if os.path.isabs(save_path):
+                raise NeedumlException(f"Given save path: {save_path}, is not relative path.")
```
danwos, line 76: .. is not **a** relative path

danwos (changes requested):
Thanks for the update.
Most change requests are regarding the docs.

docs/directives/needuml.rst
```diff
@@ -152,6 +152,44 @@ and can't be ``diagram``.
...
+      :id: INT_001
+
+      .. needuml::
+         :save: _build/_out/my_needuml.puml
+
+         Alice -> Bob: Hi Bob
+         Bob --> Alice: Hi Alice
+
+In this example, if builder :ref:`needumls_builder` is used, the plantuml-code will be exported to file at `outdir` of current builder,
+e.g. `_build/needumls/_build/_out/my_needuml.puml`.
```
danwos, line 180: Not so nice examples having twice `_build` here and `_out.
Use the same example folder names as in the other descritpions.

sphinx_needs/builder.py
```diff
@@ -92,3 +92,59 @@ def build_needs_json(app: Sphinx, _exception: Exception) -> None:
...
+                puml_content = needuml["content_calculated"]
+                # check if given save path dir exists
+                save_path = os.path.join(self.outdir, needuml["save"])
+                save_dir = os.path.dirname(save_path)
+                if not os.path.exists(save_dir):
+                    os.makedirs(save_dir, exist_ok=True)
+
+                log.info(f"Storing needuml data: {needuml} to file {save_path}.")
```
danwos, line 116: How does this look like?
Isn't ``{needuml}`` an object? How long can the string be?
I think writing the file path is enough information

danwos (changes requested):

danwos (approved):

Issue #659, opened 2022-07-27
Nested needs and custom layout with collapse_button


ubmarco:
The `collapse_button` setting `inital=True` does not work correctly with nested needs.

`conf.py`
```python
needs_layouts = {
    "common": {
        "grid": "simple",
        "layout": {
            "head": [
                '<<meta("type_name")>> **<<meta("title")>>**'
                ' <<collapse_button("meta",collapsed="icon:arrow-down-circle",'
                ' visible="icon:arrow-right-circle", initial=True)>> '
            ],
            "meta": [
                'id: <<meta_id()>>',
                '<<meta_all(no_links=True, exclude=["layout","style","template"])>>',
                "<<meta_links_all()>>",
            ],
        },
    },
}
```
`index.rst`
```rst
.. story:: Test story
   :id: ST_lvl1
   :status: open
   :layout: common

   .. story:: Test story
      :id: ST_lvl2
      :status: open
      :layout: common

      .. story:: Test story
         :id: ST_lvl3
         :status: open
         :layout: common

         .. story:: Test story
            :id: ST_lvl4
            :status: open
            :layout: common

            .. story:: Test story
               :id: ST_lvl5
               :status: open
               :layout: common
```
`Output`
![image](https://user-images.githubusercontent.com/839934/181277832-abf75ded-6b0b-47e7-ba53-b5339d04eb3b.png)
So every second need is closed an document load. The setting `inital=True` should, however, affect all nested needs.

Also the collapse button toggling does not work correctly as one click also triggers all nested needs below:
![nested_needs_collapse_affect_others](https://user-images.githubusercontent.com/839934/181278595-4ee32210-f3ed-4cfb-ac49-2e37d9b0b42c.gif)

ubmarco:
I added a minimum working example showing the a new branch into the doc_basic test data:
https://github.com/useblocks/sphinx-needs/tree/mh-nested-needs-collapse-example/tests/doc_test/doc_basic

PR #662, opened 2022-07-30
Bugfix: Fixed the collapse button issue for needs.

iSOLveIT:
Fixed the collapse button issue for needs which also affected nested needs. This fixes issue #659.

### DIFF
--- a/sphinx_needs/css/blank/styles.css
+++ b/sphinx_needs/css/blank/styles.css
@@ -90,6 +90,6 @@ table.needs_style_blue_bar {
     background: linear-gradient(to right, transparent 90%, #0fbcf9 10%);
 }
 
-.is_visible{
+.collapse_is_hidden {
     display: none;
 }
--- a/sphinx_needs/css/blank/styles.css
+++ b/sphinx_needs/css/blank/styles.css
@@ -90,3 +90,6 @@ table.needs_style_blue_bar {
     background: linear-gradient(to right, transparent 90%, #0fbcf9 10%);
 }
 
+.is_visible{
+    display: none;
+}
--- a/sphinx_needs/css/dark/styles.css
+++ b/sphinx_needs/css/dark/styles.css
@@ -115,3 +115,7 @@ table.needs_style_yellow_bar {
 table.needs_style_blue_bar {
     background: linear-gradient(to right, transparent 90%, #0fbcf9 10%);
 }
+
+.is_visible{
+    display: none;
+}
--- a/sphinx_needs/css/dark/styles.css
+++ b/sphinx_needs/css/dark/styles.css
@@ -116,6 +116,6 @@ table.needs_style_blue_bar {
     background: linear-gradient(to right, transparent 90%, #0fbcf9 10%);
 }
 
-.is_visible{
+.collapse_is_hidden {
     display: none;
 }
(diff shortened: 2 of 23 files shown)


sphinx_needs/css/blank/styles.css
```diff
@@ -90,3 +90,6 @@ table.needs_style_blue_bar {
     background: linear-gradient(to right, transparent 90%, #0fbcf9 10%);
 }
 
+.is_visible{
```
ubmarco, line 93: This is quite a generic class selector that might conflict with custom end-user CSS.
Can you make it more specific to a need/table/header/... or change the name to something like 'collape_is_visible'?
Also the name is inverted somehow.

sphinx_needs/css/dark/styles.css
```diff
@@ -115,3 +115,7 @@ table.needs_style_yellow_bar {
 table.needs_style_blue_bar {
     background: linear-gradient(to right, transparent 90%, #0fbcf9 10%);
 }
+
+.is_visible{
```
ubmarco, line 119: same here

sphinx_needs/css/modern/styles.css
```diff
@@ -124,3 +124,6 @@ table.needs_style_yellow_bar {
 table.needs_style_blue_bar {
     background: linear-gradient(to right, transparent 90%, #0fbcf9 10%);
 }
+.is_visible{
```
ubmarco, line 127: same here

ubmarco (changes requested):

sphinx_needs/css/blank/styles.css
iSOLveIT: Changes implemented in the new commit.

sphinx_needs/css/dark/styles.css
iSOLveIT: Changes implemented in the new commit.

sphinx_needs/css/blank/styles.css
```diff
@@ -90,3 +90,6 @@ table.needs_style_blue_bar {
     background: linear-gradient(to right, transparent 90%, #0fbcf9 10%);
 }
 
+.collapse_is_visible {
```
ubmarco, line 93: I just compared this class with below jQuery selector. Would the selector here still work if prefixed like this
```
span.needs.collapse_is_visible {
```
? If yes, I'd appreciate to use a more detailed selector, just to be sure it does not clash with user-space CSS.

ubmarco (approved):
Thanks for the quick fix, this is working as intended.
Thanks also for the other small fixes (typos).

sphinx_needs/css/blank/styles.css
iSOLveIT: Yes it will.

ubmarco: Ok, thanks for the update.
Then I'd vote for
```
span.needs.collapse_is_hidden {
    display: none;
} 
```
The selector is as specific as possible.
A class that makes something disappear should have something like 'hidden' in its name.
Any thoughts?

iSOLveIT: Yeah, collapse-is-hidden is ok too

PR #669, opened 2022-08-19
Added needarch directive

haiyangToAI:


danwos (approved):
wow, really nice and almll solution to introduce needarch.
Looks all good, thanks!!

PR #671, opened 2022-08-20
Variants for need options

iSOLveIT:
This PR implements variant handling based on need options.

**Usecase 1**
In `conf.py`:
```python
needs_variants = {
  "var_a": "'var_a' in sphinx_tags"  # filter_string
  "var_b": "assignee == 'me'"
}
```
In any `.rst` file:
```rst
.. req:: Example
   :id: x
   :status: var_a:open, var_b:closed, unknown
```

From the above example,  there are pre-defined variants, which have a “filter string” that defines to which variant a need belongs in the current situations.

**Usecase 2 - Filter in option**
There shall also be a way to use the filter string inside the options

```rst
.. req:: Example
   :id: x
   :status: ['var_a' in tags]:open, [assignee == 'me']:closed, unknown 
```

**Default option**
The last option can exist without a filter, so we consider it if no other options fulfil their filter rules.


docs/configuration.rst
```diff
@@ -1897,8 +1897,81 @@ Each entry describes in an "on_fail" action what to do:
 
 If "force style" is set, all other styles are removed and just the constraint_failed style remains.
 
+.. _needs_variants:
 
+needs_variants
+~~~~~~~~~~~~~~
+.. versionadded:: 1.0.1
```
danwos, line 1904: 1.0.2

docs/configuration.rst
```diff
@@ -1897,8 +1897,81 @@ Each entry describes in an "on_fail" action what to do:
...
+The dictionary consists of key-value pairs where the key is a string used as a reference to the value.
+The value is a string which consists of a Python-supported "filter string".
+
+Default: ``{}``
+
+.. _needs_variant_options:
+
+needs_variant_options
+~~~~~~~~~~~~~~~~~~~~~
+.. versionadded:: 1.0.1
```
danwos, line 1929: 1.0.2

docs/configuration.rst
```diff
@@ -1897,8 +1897,81 @@ Each entry describes in an "on_fail" action what to do:
...
+
+      * ``status``, ``tags`` and ``links`` options
+      * options specified in :ref:`extra options <needs_extra_options>`
+      * options specified in :ref:`extra links <needs_extra_links>`
+
+.. _needs_variant_data:
+
+needs_variant_data
+~~~~~~~~~~~~~~~~~~
+.. versionadded:: 1.0.1
```
danwos, line 1960: 1.0.2

docs/configuration.rst
```diff
@@ -1897,8 +1897,81 @@ Each entry describes in an "on_fail" action what to do:
...
+      :ref:`extra options <needs_extra_options>` or :ref:`extra links <needs_extra_links>`.
+   2. By default, if ``needs_variant_options`` is empty, we apply variants handling to the following options:
+
+      * ``status``, ``tags`` and ``links`` options
+      * options specified in :ref:`extra options <needs_extra_options>`
+      * options specified in :ref:`extra links <needs_extra_links>`
+
+.. _needs_variant_data:
+
+needs_variant_data
```
danwos, line 1958: I think we don't need need_variant_data.

Forgot we already have [needs-filter-data](https://sphinx-needs.readthedocs.io/en/latest/configuration.html#needs-filter-data) (Sorry!)

I think this is doing the same, or not?

docs/directives/need.rst
```diff
@@ -38,12 +38,127 @@ but you must set a title as an argument (i.e. if you do not specify :ref:`needs_
     By default, the above example works also with ``.. spec::``, ``.. impl::``, ``.. test::`` and all other need types,
     which are configured via :ref:`needs_types`.
 
+.. _needs_variant_support:
+
+Variants for options support
+----------------------------
+.. versionadded:: 1.0.1
```
danwos, line 45: 1.0.2

docs/directives/need.rst
```diff
@@ -38,12 +38,127 @@ but you must set a title as an argument (i.e. if you do not specify :ref:`needs_
     By default, the above example works also with ``.. spec::``, ``.. impl::``, ``.. test::`` and all other need types,
     which are configured via :ref:`needs_types`.
 
+.. _needs_variant_support:
+
+Variants for options support
```
danwos, line 43: I love these docs!

docs/directives/need.rst
```diff
@@ -38,12 +38,127 @@ but you must set a title as an argument (i.e. if you do not specify :ref:`needs_
...
+|out|
+
+.. req:: Variant options
+   :id: VA_003
+   :status: ['variants' in tags and not collapse]:enabled, disabled
+   :tags: variants;support
+   :collapse: true
+
+   Variants for need options in action
+
```
danwos, line 155: Use Case 3 missing :)

Using [Sphinx-Tags](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-t)

If docs get built via `sphinx-build -b html -t tag_a . _build`.
```rst
  .. req:: Example
      :id: VA_001
      :status: tag_a:open, tag_b:closed, unknown
```
"open" is set for status.

danwos (changes requested):
Only small findings in the docs.

We should also add already the Sphinx-Tags support.
Then the feature is already complete and users can go on bug hunt.

docs/configuration.rst
iSOLveIT: Okay will change it.

docs/directives/need.rst
iSOLveIT: Thank you.

docs/directives/need.rst
iSOLveIT: I will read about it. Thanks.

docs/configuration.rst
iSOLveIT: Yeah, it does the same so I will pick the data from there instead.

docs/directives/need.rst
iSOLveIT: Included in update.

danwos (approved):
Thanks for the great work.

PR #676, opened 2022-08-29
Pydantic modeling

ubmarco:


danwos:
Will be part of an external extension

Issue #604, opened 2022-06-10
Jinja support for need content


danwos:
The content of a need should be able to get handled by jinja (incl. all injected Sphinx-Needs data).

The result would be the same as if you are using the Sphinx-Needs template mechanism: https://sphinxcontrib-needs.readthedocs.io/en/latest/directives/need.html#template

But it would be much easier to use, because:

* No extra template file needs to be created
* The content is specified only once, so a need-specific template file can be avoided.

Example:
```rst
.. req:: My requirement
   :links: REQ_1, REQ_23 
   :jinja_content:  True / False # if set, support gets activated 
 
   This requirement has status: **{{status}}**.
   It links to:
   {% for link in links %}
   *  {{needs[link]["title"]}}    
   {% endfor %}

```

Jinja support shall not be activated by default, as this can have some strange side-effects if the user uses characters like ``{{`` without knowing that this gets handled by jinja.

The options shall be able to get set via global_option filters as well.
Option name must be discussed ;)

twodrops:
Thanks for creating the issue.

Maybe as part of this or a separate issue the documentation of Jinja support in sphinx-needs could be extended. A standalone jinja page with references to other jinja related features in sphinx-needs and materials on internet on Jinja. This is a very powerful feature which many sphinx-needs users does not know yet (so it needs some marketing  😄 )

The places I found in Sphinx-Needs until now where Jinja is used are
- sphinx-needs jinja templates
- within `conf.py` as part of `needservice` - Is this coming from sphinx-needs by the way?
- needuml

Perhaps there are more.

PR #678, opened 2022-08-30
Added Jinja support for need content via the `jinja_content` option

iSOLveIT:
Tasks
- [x] * Provide a need option (flag) to activate jinja-parsing for the content (this is need-specific)
- [x] * Render the content with jinja

As the jinja context, the following should be available
* [needs_filter_data](https://sphinxcontrib-needs.readthedocs.io/en/latest/configuration.html#needs-filter-data)
* need-data (if possible, as rst-rendering is done very early)

This PR is a fix for issue #604.


danwos (changes requested):
Looks great, thanks :+1: 

I'm only missing an entry in the changelog :)

twodrops:
Thanks for implementing this in short time.
Could you please take care of the documentation part I mentioned [here](https://github.com/useblocks/sphinx-needs/issues/604#issuecomment-1152239751), as part of this PR or a separate issue?

iSOLveIT:
Thanks. I will add it to the changelog and do the necessary changes.

danwos (approved):

Issue #640, opened 2022-07-08
Defining Snippet templates for different need types


haiyangToAI:
Issue moved from https://github.com/open-needs/open-needs-ide/issues/69.

MohSam1:
Sure this makes sense to move it here

haiyangToAI:
Hi @MohSam1,

Regarding to your question, the use case you mentioned has been supported in the directive snippet feature.

twodrops:
@danwos @haiyangToAI 

I would like to know a bit more on this snippet feature.

The current snippet feature is useful only to know which need-types (directives) are available. The snippet that is created has only the mandatory `extra_options`  (`id` and `status`), which is not often useful in projects with several types and internally defined "mandatory" `extra_options` and `extra_links`.

We therefore use the VSCode snippet feature to create complete snippets for each need-type. This is however a duplication of need definition in conf.py and VSCode snippet and hard to maintain. We are therefore creating an alternate need definition possibility using json/schema or RDF (instead of `conf.py`), so that more useful snippets can be created out of these. 

Is this task addressing these current limitations in some way?  Thats is,
- Have need definition only once
- Use this definition to auto-create snippets

danwos:
No, the problem of defining somehow multiple times the same is not addressed by this.
Also, there is no automation of snippet-creation/definition planned.

So it provides the user only the possibility to define the snipper per hand for each need-type.

Example (not a spec ;) ):
```python

needs_ide_snippets = {
'req': 
""".. req:: Example
      :id:
      :status: 
      :custom_1: 
""",

'test': 
""".. test:: Test Case 1
      :id:
      :custom_1: 
      :custom_2: 
"""
}
```
Maybe this definition does support Jinja in the future, so that can provide jinja-functions for ID-Generation or other magic.

The key (like `req`) could also be a regex, to address multiple need-types with one definition.

twodrops:
Thanks for the clarification @danwos.

This means that the need-definitions and `needs_ide_snippets ` needs to be kept in sync manually. 
I find this better than keeping it in sync with VsCode snipprts. Atleast both are in `conf.py`.

Also we can then auto-create `needs_ide_snippets` from our json-schema/rdf solution which is also good.

@StephanKuempel fyi

MohSam1:
@danwos : Any rough idea when this issue is planned to implemented ?

MohSam1:
Looks like this feature is now implemented, can we close the issue?

PR #680, opened 2022-09-01
Support custom directive snippets for IDE features

haiyangToAI:
#640

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1210,6 +1210,41 @@ or
 
 {% endraw %}
 
+.. _needs_ide_directive_snippets:
+
+needs_ide_directive_snippets
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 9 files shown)


docs/configuration.rst
```diff
@@ -1183,6 +1183,41 @@ or
...
+    .. test:: Test Title
+       :id: TEST_
+       :status: open
+       :custom_option: something
+
+       test directive content.
+    """,
+    }
+
+If ``needs_ide_directive_snippets`` is not configured or empty, the default directive snippets
```
twodrops, line 1218: @haiyangToAI Could you please add a short hint that the snippets are not automatically synced with the need definitions in `conf.py` and it is up to the user to keep them in sync.

danwos (changes requested):
Really impressive work. 
using the esbonio API the right way does not look easy :+1: 

Just add the wish from @twodrops  to the docs, then we can merge this PR.

danwos (approved):

PR #681, opened 2022-09-01
Added jinja func import() for needarch

haiyangToAI:


sphinx_needs/directives/needuml.py
```diff
@@ -171,6 +196,18 @@ def process_needuml(app, doctree, fromdocname):
         id = node.attributes["ids"][0]
         current_needuml = env.needs_all_needumls[id]
 
+        parent_need_id = None
+        # Check if current needuml is needarch
+        if current_needuml["is_arch"]:
+            # Check if needarch is only used inside a need
+            from sphinx_needs.directives.need import Need
```
danwos, line 203: This import should be at the top of the file.
It does not save really any time, if done in this if-statement.

sphinx_needs/directives/needuml.py
```diff
@@ -208,9 +245,22 @@ def process_needuml(app, doctree, fromdocname):
         mem_template = Environment(loader=BaseLoader).from_string(uml_content)
 
         # Get all needed Jinja Helper Functions
-        jinja_utils = JinjaFunctions(app, fromdocname)
+        jinja_utils = JinjaFunctions(app, fromdocname, parent_need_id)
         # Make the helpers available during rendering
-        data = {"needs": all_needs, "uml": jinja_utils.uml, "flow": jinja_utils.flow, "filter": jinja_utils.filter}
+        if current_needuml["is_arch"]:
+            # Add jinja function import() only for needarch
+            data = {
```
danwos, line 252: Define `data` only once and add "import" only if `is_arch`.
Currently, we define the data twice with nearly the same content, hard to maintain.

danwos (changes requested):
Thanks for this PR. :+1: 
Looks already really good, only have 1-2 remarks.

danwos (approved):

PR #708, opened 2022-09-23
Matrix test includes docutils versions

danwos:


noxfile.py
```diff
@@ -1,8 +1,9 @@
 import nox
 from nox_poetry import session
 
-PYTHON_VERSIONS = ["3.8", "3.9.7", "3.10"]
-SPHINX_VERSIONS = ["5.0.2"]
+PYTHON_VERSIONS = ["3.8", "3.10"]
+SPHINX_VERSIONS = ["5.0.2", "4.5"]
```
twodrops, line 5: @danwos How about including Sphinx 5.1.1 as well? Or does Sphinx-Needs dependency exclude it?

danwos: Will update the list to newest releases, when the CI is working :)

Try to not have too many tests here
If we have 2 versions per tool/lib, we already have 2x2x2 = 8 test runs.

Issue #701, opened 2022-09-19
Support integration of external needs exported from Codebeamer via SNE


Rubyfi:
When importing a needs.json file which was exported from Codebeamer via SNE, the generated links aren't inferred correctly.

Codebeamer uses the URL format `<base-url>/issue/<item ID>` for its items (e.g. requirements).
The exported needs may look as follows:
```
"1234": {
    "description": "...",
    "id": "1234",
    "origin": "1234",
    "title": "...",
    "type": "..."
},
```

However, when importing such a needs file, e.g. by using the following `needs_external_needs` configuration
```
needs_external_needs = [
        {
            "base_url": <base_url>/issue,
            "json_url": ...
            "version": "1.0",
            "css_class": "external_link",
        },
]
```
the generated links use the format `<base-url>/issue/1234.html#1234` (this behaviour is hard-coded [here](https://github.com/useblocks/sphinx-needs/blob/b543e228840f3b70ed3f46414b0e245ad651c044/sphinx_needs/external_needs.py#L83)).

As a temporary workaround I use a custom Sphinx extension which lets users specify Codebeamer base URLs in the `conf.py`. After the external needs are set up, it iterates over all needs and removes the ".html#<ID>" suffixes from the external_url entries that start with one of the specified base URLs.

Would it be possible to add an option to needs_external_needs entries that allows users to modify the default behaviour, and e.g.  omit the .html suffix?

danwos:
Interesting use case. 
It's hard-coded because the "target" for `needexternal` was planned to be another Sphinx documentation :)

But I understand your case, and yes that should be configurable.

Maybe like this:

```python
needs_external_needs = [
        {
            "base_url": <base_url>/issue,
            "target_url": 'issue/{need["id"]}'   # <-- NEW
            "json_url": ...
            "version": "1.0",
            "css_class": "external_link",
        },
]
```
``target_url`` takes a formattable string and ``need`` contains the current need.

Would this be already helpful?

Rubyfi:
Thank you for the fast response. I think that should solve the issue.

PR #713, opened 2022-09-28
support target_url in needs_external_needs

haiyangToAI:
#701


sphinx_needs/external_needs.py
```diff
@@ -19,6 +20,10 @@ def load_external_needs(app: Sphinx, env: BuildEnvironment, _docname: str) -> No
         if source["base_url"].endswith("/"):
             source["base_url"] = source["base_url"][:-1]
 
+        target_url = ""
+        if source.get("target_url", False):
+            target_url = source["target_url"]
```
danwos, line 25: That's the same as:
```
target_url = source.get("target_url", "")
```
or not?

danwos (changes requested):
Thanks for the PR.
Only one small code enhancement :)

sphinx_needs/external_needs.py
haiyangToAI: much better, thanks.

danwos (approved):

Issue #691, opened 2022-09-13
Rendering issues with Immaterial Theme


twodrops:
The collapse/expand button of needs have some rendering issues with the [Immaterial Theme](https://jbms.github.io/sphinx-immaterial/)

![image](https://user-images.githubusercontent.com/103930/189815363-81f3fd98-6ced-407e-85dc-e27e70cb0ab5.png)

See the problem also in Sphinx-Needs Documentation
https://sphinxcontrib-needs.readthedocs.io/en/latest/directives/needextract.html#needextract

PR #716, opened 2022-09-29
Fixed rendering issues for collapse/expand button.

iSOLveIT:
Fix for issue #691.

### DIFF
--- a/sphinx_needs/layout.py
+++ b/sphinx_needs/layout.py
@@ -5,6 +5,7 @@ Based on https://github.com/useblocks/sphinxcontrib-needs/issues/102
 """
 import os
 import re
+import uuid
 from contextlib import suppress
 from typing import List, Optional
 from urllib.parse import urlparse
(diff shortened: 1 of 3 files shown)


sphinx_needs/layout.py
```diff
@@ -130,9 +133,14 @@ def build_need(layout, node, app: Sphinx, style=None, fromdocname: Optional[str]
 
     lh = LayoutHandler(app, need_data, need_layout, node, style, fromdocname)
     new_need_node = lh.get_need_table()
+    node_container.append(new_need_node)
+
+    container_id = "SNCB-" + str(uuid.uuid4())[:32]
```
danwos, line 138: I thin we do not need such long ids.
It only must be unique on the same page, or not?
So 4 randoms chars should be enough.

danwos (changes requested):

sphinx_needs/layout.py
iSOLveIT: Okay noted.

danwos (approved):

Issue #704, opened 2022-09-19
Custom jinja context


danwos:
There shall be a config option like html_context, which allows defining custom content to jinja.

This is already done for filter string, but we need it also for :
* Templates
* string-links
* Other functions, where Jinja is used.

twodrops:
@danwos Thanks for creating this issue as discussed. Please check if you can provide a detailed page on Jinja support in Sphinx-Needs as part of this issue or this one https://github.com/useblocks/sphinx-needs/issues/604#issuecomment-1152239751

PR #717, opened 2022-09-29
Added support for custom data as jinja context.

iSOLveIT:
This PR is the fix for issue #704.
It improves the plugin by adding support for custom data which we use as context when rendering Jinja templates or strings.


danwos (changes requested):
Thanks for the PR.
Looks all good.

But can you please add also a test case or update the existing ones?

iSOLveIT:
Okay, I will add the test case right away.

danwos (approved):

PR #734, opened 2022-10-16
Needuml fix import issue and loop detection

### LINKED ISSUE #730: needuml: uml() call of needarch with import() in - lead to an exception
Source code:
```
.. comp:: Import Example - Importee
   :id: COMP_T_004

.. comp:: Import Example - Importer
   :id: COMP_T_005
   :tests: COMP_T_004

   .. needuml::

      {{flow(need().id)}}
      {{import('tests')}}
      {% for test in need().tests%}
      {{need().id}} --> {{test}} : tests
      {% endfor %}

.. needuml::

   {{uml(needs['COMP_T_005'].id)}}
```

Lead to `Jinja function import is not supported for needuml.` from [here](https://github.com/useblocks/sphinx-needs/blob/427194f27e178553df96428ad04630d0dd1bc079/sphinx_needs/directives/needuml.py#L259).
Additionally, the test one line above is not good, as I could even write `{{ import(...`.

### LINKED ISSUE #731: needuml: uml() call circle leads to an exception
I could create in needuml an `uml()` circle, like `(uml('1') -> uml('2') -> uml('3') -> uml('1')`.
This can happen intended or unintended over many kind of elements.

Code:
```
.. comp:: COMP_T_001
   :id: COMP_T_001

   .. needuml::

      {% set need = needs['COMP_T_001'] %}
      {{flow(need.id)}}
      {% if variant == "A" %}
      {{uml('COMP_T_003', variant="A")}}
      usecase {{need.id}}_usecase
      {% else %}
      {{uml('COMP_T_003')}}
      {{uml('COMP_T_003', variant="A")}}
      {% endif %}

.. comp:: COMP_T_002
   :id: COMP_T_002

   .. needuml::

      {% set need = needs['COMP_T_002'] %}
      {{flow(need.id)}}
      {% if variant == "A" %}
      {{uml('COMP_T_001', variant="A")}}
      usecase {{need.id}}_usecase
      {% else %}
      {{uml('COMP_T_001')}}
      {% endif %}

.. comp:: COMP_T_003
   :id: COMP_T_003

   .. needuml::

      {% set need = needs['COMP_T_003'] %}
      {{flow(need.id)}}
      {% if variant == "A" %}
      {{uml('COMP_T_002', variant="A")}}
      usecase {{need.id}}_usecase
      {% else %}
      {{uml('COMP_T_002')}}
      {% endif %}
```

### LINKED ISSUE #732: Request: needarch: provide need() function to get "need data"
The `need()` function provides you the need information the `needarch`  is embedded in.

### LINKED ISSUE #737: needuml: flow() shall return plantuml text without newline
I want to embedded elements in other elements.
Like this:
```
@startuml
card card {
card card2
}
@enduml
```
If I do want to generate the `card card` part, flow() has to finish without a newline.
```
@startuml
{{flow("card")}} {
{{flow("card2")}}
}
@enduml
```

This is in general a breaking change, but often human do write the need newline anyway.
Example 1:
```
.. needuml::

   {{flow("e1")}}
   {{flow("e1")}}
```
has the needed newlines in the code.

Counter example to 1:
```
.. needuml::

   {{flow("e1")}}{{flow("e1")}}
```
Would in the future generate a failed plantuml syntax. 
I believe the new behaviour is intuitive, and would not generate issues.

Example 2 from scripts:
```
.. needuml::

   {% for e in filter(...) %}
   {{flow(e.id)}}
   {% endfor %}
```
Even here the script generates the needed newlines.

Counter example to 2:
```
.. needuml::

   {% for e in filter(...) %}{{flow(e.id)}}{% endfor %}
```
Would in the future generate a failed plantuml syntax. 
I believe the new behaviour is intuitive, and would not generate issues.

Any opinions?

### LINKED ISSUE #742: needuml used but "sphinxcontrib.plantuml" not installed leads to exception
If you run sphinx-needs with out extension "sphinxcontrib.plantuml" and use needuml , you get the exception:
```
NotImplementedError: <class 'sphinx_immaterial.html_translator_mixin.get_html_translator.<locals>.CustomHTMLTranslator'> departing unknown node type: Needuml
```
I expect to get an error node, where a needuml /needarch is been defined.

Code to handle this is been [commented out](https://github.com/useblocks/sphinx-needs/blob/029387ef8448c8ef6c551edd6759d024ccbe8bab/sphinx_needs/directives/needuml.py#L222).

PhilipPartsch:
I try to fix #730, fix #731, fix #732, fix #737 and fix #742.
I want to refactor the double processing in needuml [uml()](https://github.com/useblocks/sphinx-needs/blob/427194f27e178553df96428ad04630d0dd1bc079/sphinx_needs/directives/needuml.py#L148) and [process_needuml](https://github.com/useblocks/sphinx-needs/blob/427194f27e178553df96428ad04630d0dd1bc079/sphinx_needs/directives/needuml.py#L263) to a single source implementation. It shall encapsulate the Jinja handling and should be reusable in other locations.

Anyway, I want to get some feedback.


sphinx_needs/directives/needuml.py
```diff
@@ -303,6 +395,8 @@ def process_needuml(app, doctree, fromdocname):
 
         node.replace_self(content)
 
+        pro_ids = jinja_utils.get_processed_need_ids()
```
PhilipPartsch, line 398: Dead code

PhilipPartsch: Fixed

danwos (changes requested):

docs/directives/needarch.rst
```diff
@@ -41,14 +39,65 @@ Jinja context
...
 
+
+.. _needarch_jinja_need:
+
+need()
+~~~~~~
+
+.. versionadded:: 1.0.3
+
+The `need()` function provides you the need information the :ref:`needarch` / :ref:`needuml` is embedded in.
```
danwos, line 50: need() should not be accessible in needuml, as needuml can be used outside a need.

So the same code would behave differently, if used inside a need or not.
I personally don't like this :/

danwos (changes requested):
Great PR, a lot of changes.
Haven't found any problem.

Only the part of the general discussion, what kind of jinja-functions needuml shall get

docs/directives/needarch.rst
```diff
@@ -41,14 +39,65 @@ Jinja context
 
 The following Jinja functions are **only available** for :ref:`needarch`. 
 
+
+.. _needarch_jinja_need:
+
+need()
```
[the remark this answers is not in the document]
PhilipPartsch: I choose a function, so I can have a call point to check if we are embedded in a need. If not, I can inject a exception for a better usability.

docs/directives/needarch.rst
PhilipPartsch: Implemented as requested. Even covered with tests in needarch (to support need()) and needuml (to get exception in case of use need()).

danwos (approved):
Thanks for this great work. :+1: 

Found 1-2 minor stuff in the docs, but I will correct them with another PR.

Issue #743, opened 2022-10-19
l_foucs: content allignment not stable/left


kreuzberger:
**Update**: Issue is reused as bug report. Bug description starts in https://github.com/useblocks/sphinx-needs/issues/743#issuecomment-1283639247

following the documentation i can add "custom" css files to change "default" needs styling done in css.
But if i add these css "normaly" in sphinx (e.g. with app.add_css_file) the generated html include order is
- default html style
- my styles
- sphinx-needs style

so i can not "overwrite" the default sphinx-needs styles because they are applied later.
is this intended?

E.g. my custom css is named "fsp.css" and added via add_css_file the generated html is

```
<link rel="stylesheet" type="text/css" href="_static/pygments.css" />
<link rel="stylesheet" type="text/css" href="_static/nature.css" />
<link rel="stylesheet" type="text/css" href="_static/fsp.css" />
<link rel="stylesheet" type="text/css" href="_static/assets/jsonview.bundle.css" />
<link rel="stylesheet" type="text/css" href="_static/sphinx-needs/libs/html/datatables.min.css" />
<link rel="stylesheet" type="text/css" href="_static/sphinx-needs/common.css" />
<link rel="stylesheet" type="text/css" href="_static/sphinx-needs/styles.css" />
```

E.g. i want to change default focus_l style back to "former" layout with ID written not vertical.

danwos:
I'm currently not Infront of my pc to check the details, but does add_css_file have a second parameter for the order?

You could also use the way to add css via option html_css_files. Maybe these files get added later.

kreuzberger:
You are right. The add_css_file has an priority argument. But the better way would be to use html_css_files. It takes the highest priority as default.
https://sphinxcontrib-needs.readthedocs.io/en/latest/layout_styles.html#own-css-file-on-sphinx-level should be adapted by either adding priority or the html_css_files configuration option.

BTW: If i use focus_l as layout. The content of the needs is not left aligned depending on the width of the table. The shorter the text, the more it is moved to the right. Which property is responsible for this behaviour?
![Screenshot_20221019_102316](https://user-images.githubusercontent.com/579125/196640763-c3479467-3fb1-4bf3-9562-c7407ee2c99e.png)

danwos:
For the alignment problem: I don't have a clue and the behavior looks strange.
Maybe it also behaves different in a different browser.

I rename this issue, so we can keep it as bug report for the alignment problem.

kreuzberger:
[...]

PR #744, opened 2022-10-19
[#743] update documentation for custom css files

kreuzberger:
Instead of using `app.add_css_files` the better way seems to use `html_css_files` cause the files will be added **after** the extension css files by default.

Fix for #743

### DIFF
--- a/docs/layout_styles.rst
+++ b/docs/layout_styles.rst
@@ -572,8 +572,7 @@ Own CSS file on sphinx level
 If you want to use most of the sphinx-needs internal styles but only need some specific changes for single elements, you
 can provide your own CSS file by register it inside your conf.py::
 
-    def setup(app):
-        app.add_stylesheet('css/my_custom.css')  # may also be an URL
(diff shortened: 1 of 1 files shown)


docs/layout_styles.rst
```diff
@@ -572,8 +572,9 @@ Own CSS file on sphinx level
 If you want to use most of the sphinx-needs internal styles but only need some specific changes for single elements, you
 can provide your own CSS file by register it inside your conf.py::
 
-    def setup(app):
-        app.add_stylesheet('css/my_custom.css')  # may also be an URL
+    html_css_files = ['css/my_custom.css'])  # may also be an tuple
```
danwos, line 575: ")" at the end not needed.

"# may also be an tuple": I would remove this part.
Was hard for me to understand, what benefits a tuple has and why it is needed.
The later link to the docs makes it clear. But first it was confusing and this feature may not be needed for most users.

danwos (changes requested):
Thanks for the quick fix :+1: 

One small finding :)

PR #762, opened 2022-10-31
check_links() gets called only once now

### LINKED ISSUE #756: check_links needs too much time
This runtime trace is from a more complex build, which is using also Sphinx-Needs.
The overall runtime was ~4h.

image

As you may see, `check_links()` (inside ``process _need_nodes``) it taking **~24.6%** of build time (~1h), which is a lot.
We need to optimize this function.
It also looks like this function does not scale well, as in smaller projects the time consumption is less. (but often still high).

### Technical solution
tbd. No idea how the code looks like...

danwos:
* [x] Fix implemented
* [x]  Make before/after runtime tests

Fixes #756

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,6 +12,10 @@ License
 -----
 Under development
 
+* Improvement: Performance fix needs processing.
+(`#756 <https://github.com/useblocks/sphinx-needs/issues/756>`_)
+* Improvement: Performance fix for needflow.
+(`#760 <https://github.com/useblocks/sphinx-needs/issues/760>`_)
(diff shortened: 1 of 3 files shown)


danwos:
Build without fix: 781.21 s  build time

`sphinx-analysis --project needs --local /home/daniel/workspace/sphinx/sphinx-needs --pyinstrument --tree --pages 20 --needs 5 --folders 10 --depth 2 --needtables 0 --needflows 0 --needpies 0`

The ``check_link`` function takes only 5.4% or 42.4 s  of the build time.
So not as dramatic as the 4h measurement I got from a user.
Looks like my test scenario is different (maybe to less not-need related rst files)

image

PR #763, opened 2022-10-31
fix import function to work with new flow function

### LINKED ISSUE #752: Bug: Not released needarch import() function has bug
Currently import() in needarch is not adapted to new flow() function which does no longer return a newline after each element. 
Todo:
- Check in import that flow returns a newline, if not add one. 
- Enhance tests to cover this for future behavior.

PhilipPartsch:
Fix #752


danwos (approved):
Thanks for the quick fix :+1:

PR #766, opened 2022-10-31
Needtable set output table

### LINKED ISSUE #434: Table needs id
``sphinxcontrib-needs/docs/directives/needtable.rst:4: WARNING: Any IDs not assigned for table node``

With version 0.7.4 normal tables get wrapped inside a container/div.
In this case the generated docutils id goes to the container and not to the table.

This seems to make problems. Maybe we should just generate a dummy-id or set an id.

PhilipPartsch:
Fix #434

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,5 @@ need_bar_*.png
 need_pie_*.png
 output.json
 mem_out.*
+
+pyinstrument*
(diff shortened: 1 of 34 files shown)


PhilipPartsch:
No Idea why we get the warnings. Locally  do not get these

danwos:
It's strange that only this PR has this CI problems.

I understand the issue (pytet-xdist can not be installed together with pytest-benchmark) and will create a PR soon, but it's fancy that this issue has not shown up earlier.

After CI fix is implemented I will retrigger the CI runs in this PR.

However, thanks already for the PR, looks good to me. :+1:

danwos (approved):

Issue #776, opened 2022-11-04
needs_variant_options shall not be used by default


danwos:
The "variant" feature makes some trouble, as it scans nearly all option b default and detects often a variant string, where it is a normal string with a colon.
Our regular expressions may not be good enough, but it may also be tough not to have false positive detections for it.

Therefore we should deactivate variant handling by default, and only activate it for options given by [needs-variant-options](https://sphinx-needs.readthedocs.io/en/latest/configuration.html#needs-variant-options)

So the new behavior is if ``needs-variant-options`` is ``[]`` do nothing, instead of checking all options as of right now.

PR #777, opened 2022-11-04
Fixed issue with needs_variant_options #776.

iSOLveIT:
This PR is the fix for issue #776.

### DIFF
--- a/sphinx_needs/functions/functions.py
+++ b/sphinx_needs/functions/functions.py
@@ -250,33 +250,31 @@ def resolve_variants_options(env: BuildEnvironment):
         return
 
     variants_options = env.app.config.needs_variant_options
-    if len(variants_options) == 0:
-        # If variant_options is not provided
-        default_variant_options = ["status", "tags", "links"]
-        default_variant_options.extend(env.app.config.needs_extra_options)
-        extra_links = [extra_link["option"] for extra_link in env.app.config.needs_extra_links]
-        default_variant_options.extend(extra_links)
-        variants_options.extend(set(default_variant_options))
+    enable_variant_handling = False
+    if len(variants_options) != 0:
+        # If variant_options is provided or not empty then activate variant handling
+        enable_variant_handling = True
 
-    needs: Dict = env.needs_all_needs
-    for need in needs.values():
-        # Data to use as filter context.
-        need_context: Dict = {**need}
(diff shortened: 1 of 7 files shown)


sphinx_needs/functions/functions.py
```diff
@@ -250,33 +250,31 @@ def resolve_variants_options(env: BuildEnvironment):
...
-                    need[need_option] = match_variants(option_value, need_context, env.app.config.needs_variants)
-                else:
-                    option_value = need[need_option]
-                    need[need_option] = match_variants(option_value, need_context, env.app.config.needs_variants)
+    enable_variant_handling = False
+    if len(variants_options) != 0:
+        # If variant_options is provided or not empty then activate variant handling
+        enable_variant_handling = True
+
+    if enable_variant_handling:
```
danwos, line 258: Could this be done in one line?

``if variants_options: ...``.

An empty list is always False in Python3

sphinx_needs/functions/functions.py
```diff
@@ -250,33 +250,31 @@ def resolve_variants_options(env: BuildEnvironment):
...
+        needs: Dict = env.needs_all_needs
+        for need in needs.values():
+            # Data to use as filter context.
+            need_context: Dict = {**need}
+            need_context.update(**env.app.config.needs_filter_data)  # Add needs_filter_data to filter context
+            _sphinx_tags = env.app.builder.tags.tags  # Get sphinx tags
+            need_context.update(**_sphinx_tags)  # Add sphinx tags to filter context
+
+            for need_option in need:
```
danwos, line 267: Maybe we should also change this loop to check not all need-options, but the options in variants-need only:

```python
for var_option in variants_options:
   if var_option in need:
       # do all the magic
```
This would save some loops :)

danwos (changes requested):
Thanks for the fix, 1-2 smaller changes would be nice.

Maybe it makes also sense to add a test case, where an option-value, which looks like a variant-string, is in the final HTML, if ``variant_options`` is empty.
Just to be sure that option-values stay untouched.

sphinx_needs/functions/functions.py
iSOLveIT: Okay sure

danwos (approved):

PR #789, opened 2022-11-09
Add support for link in needuml directive

adegroote:
It is useful for sequence or activity diagramm

For now, it is mainly to discuss the idea. I will add a test and update the documentation if you feel it useful for the project.


danwos:
Thanks for the PR :+1: 

Do you have some example code, how `links()` shall be used, and maybe an image of how what it looks like?
Hard imagine just from the code.

adegroote:
For example, the following snippet

```
FOO
=====


.. req:: Base requirement
   :id: FOO_0001


.. spec:: A complex specification
   :id: FOO_0004
   :links: FOO_0001

   .. spec:: Positive refinement of the specification
      :id: FOO_0004_1

   .. spec:: Negative refinement of the specification
      :id: FOO_0004_2


.. needuml::
   :debug:

   start
   repeat
   if ({{link('FOO_0004', 'title')}}) then (true)
    :{{link('FOO_0004_1', 'title')}};
   else (false) 
    :{{link('FOO_0004_2', 'title')}};
   endif
   repeat while (end_of_universe?)
   stop
```

would lead to something like that (broken link of course)

foo_activity

danwos:
Ohh that looks great and I see the use case.

I'm not sure if we should call the function `link`, because  I have to think about links between needs.
Would it be okay to call it `ref`? Like the Sphinx `ref` role?

And we maybe need a third parameter, which is just a custom text for the link, if this shall not be taken from the linked need.
`link(id, content=None, text=None)`
Use it like: `link('NEED-123', text='Custom link text')`.

Would this be okay for you?

adegroote:
Fine for me,  I would prepare something in this direction.

danwos:
The code looks good to me. :+1: 
Now only docs (incl. changelog!) and test-case are missing.
And the most important point, if you like, put yourself in the AUTHORS file.

Thanks for your contribution!

adegroote:
I have submitted a proposal, through I fail to build the doc locally (even without my change). Do you have some specific instruction for dev ? (most of the Makefile target just fails locally :()

docs/directives/needuml.rst
```diff
@@ -345,6 +345,32 @@ Finds a list of Sphinx-Need objects that pass the given filter string.
       {% endfor %}
 
 
+.. _needuml_jinja_ref
+
+ref(id, content, text)
```
danwos, line 350: Sorry, to bother you here again.
I have thought about ``content`` and I think it's not 100% clear what it means, as it describes what it is used for and not what the source of the data is.
Can we rename it to ``option``, as all need-values in your documentation are referenced as ``option``?

danwos (changes requested):
Thanks for the update.
Only one naming suggestion :)

Docs and Tests are looking fine

danwos:
> I have submitted a proposal, through I fail to build the doc locally (even without my change). Do you have some specific instruction for dev ? (most of the Makefile target just fails locally :()

It should work locally, if you install the pinned version from ``poetry.lock`` file.
So just use poetry to install the project and setup the env.
``poetry install``

docs/directives/needuml.rst
```diff
@@ -345,6 +345,32 @@ Finds a list of Sphinx-Need objects that pass the given filter string.
...
+   .. needuml::
+
+      Alice -> Bob: {{ref(COMP_001, content='title')}}
+      Bob -> Alice: {{ref(COMP_002, text='A completely free text')}}
+
+|out|
+
+.. needuml::
+
+      Alice -> Bob: {{ref(COMP_001, content='title')}}
```
danwos, line 371: The used need-ID must be valid.
So use an existing one, like ``FEATURE_1``, or create your own example-needs for the docs.

danwos:
For the liniting problems, you can also install the pre-commit hooks.
See for details: https://sphinx-needs.readthedocs.io/en/latest/contributing.html#installing-dependencies

adegroote:
> > I have submitted a proposal, through I fail to build the doc locally (even without my change). Do you have some specific instruction for dev ? (most of the Makefile target just fails locally :()
> 
> It should work locally, if you install the pinned version from `poetry.lock` file. So just use poetry to install the project and setup the env. `poetry install`

Thanks for the feedback. I don't have much experience with poetry. I can now run `make test`, but `make html-docs` still fails, with sphinx_copybutton missing.

danwos:
> > > I have submitted a proposal, through I fail to build the doc locally (even without my change). Do you have some specific instruction for dev ? (most of the Makefile target just fails locally :()
> > 
> > 
> > It should work locally, if you install the pinned version from `poetry.lock` file. So just use poetry to install the project and setup the env. `poetry install`
> 
> Thanks for the feedback. I don't have much experience with poetry. I can now run `make test`, but `make html-docs` still fails, with sphinx_copybutton missing.

For building the docs, you must install additionally also the documentation requirements, as they are not needed by Sphinx-Needs itself. Just execute ``pip install -r docs/requirements.txt` and you should be able to build the docs.

adegroote:
I still have trouble to build the documentation (sphinx-copybutton is installed through pip, but sphinx-needs still complains about it), hence I'm not able to progress much more on the documentation build.

adegroote:
@danwos I progress on the documentation build, but I still have failure when trying to compile it, relating to services I guess. It fails with 

`
requests.exceptions.ConnectionError: HTTPConnectionPool(host='127.0.0.1', port=9595): Max retries exceeded with url: /auth/jwt/login (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fcbf5792fb0>: Failed to establish a new connection: [Errno 111] Connection refused'))
`

Do you have any guideline to help me fixing locally my build ?

danwos:
Building the docs is normally not needed and needs also some special setup, as it fetches some data from GitHub during the build, so a github-api key needs to be set.

But if you really want to build the docs, you can set `ON_CI=true` and `FAST_BUILD=true` and ENV vars. These skip any actions with external services.

I just checked the contribute-docs and this is totally missing. Sorry for the confusion.

So running the tests via `make tests` and the linter via `make lint` is normally enough.

danwos (approved):
Looks all good and thanks for the PR :+1: 

Feel free to add yourself to the AUTHORS file, if you like :)

Issue #796, opened 2022-11-17
NotImplementedError: <class 'sphinx.writers.latex.LaTeXTranslator'> departing unknown node type: NeedRef


Pierre-Gronau-ndaal:
Exception occurred:
  File "/usr/local/lib/python3.10/site-packages/docutils/nodes.py", line 2040, in unknown_departure
    raise NotImplementedError(
NotImplementedError: <class 'sphinx.writers.latex.LaTeXTranslator'> departing unknown node type: NeedRef
The full traceback has been saved in /var/folders/vf/rxgmxgc5225gcwgk4g9c8yq40000gn/T/sphinx-err-o5uk2ndy.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
make: *** [latexpdf] Error 1

apears with

Sphinx 5.2.3, 5.3, 6.0b
docutils 0.19

on command make singlehtml or make latexpdf

# Python version: 3.10.8 (CPython)
# Docutils version: 0.19 
# Jinja2 version: 3.1.2
# Last messages:
#   _legal/Information
#   _finalchapter/figs
#   _finalchapter/tables
#   _finalchapter/bib
#   _finalchapter/final
#   
#   resolving references...
#   erledigt
#   writing...
#   failed
# Loaded extensions:
#   sphinx.ext.mathjax (6.0.0b2) from /usr/local/lib/python3.10/site-packages/sphinx/ext/mathjax.py
#   sphinxcontrib.applehelp (1.0.2) from /usr/local/lib/python3.10/site-packages/sphinxcontrib/applehelp/__init__.py
#   sphinxcontrib.devhelp (1.0.2) from /usr/local/lib/python3.10/site-packages/sphinxcontrib/devhelp/__init__.py
#   sphinxcontrib.htmlhelp (2.0.0) from /usr/local/lib/python3.10/site-packages/sphinxcontrib/htmlhelp/__init__.py
#   sphinxcontrib.serializinghtml (1.1.5) from /usr/local/lib/python3.10/site-packages/sphinxcontrib/serializinghtml/__init__.py
#   sphinxcontrib.qthelp (1.0.3) from /usr/local/lib/python3.10/site-packages/sphinxcontrib/qthelp/__init__.py
[...]

PR #805, opened 2022-11-21
nodes: Attach a dummy visitor for each node

tim-nordell-nimbelink:
Fixes https://github.com/useblocks/sphinx-needs/issues/796.

I don't know sphinx extensions well enough to know if this is the correct way of fixing it, but it fixes it here.


tim-nordell-nimbelink:
Closing this since it doesn't actually fix the underlying issue; just masks it.

PR #815, opened 2022-11-23
Feature/vector need figures

tim-nordell-nimbelink:
This adds support for saving the need_bar_* and need_pie_* into SVG and PDF formats as well.  The builder will pick the suitable format for the output document format.  (E.g. with HTML it ends up with the SVG, and with latexpdf it ends up with the PDF.)


danwos:
Thanks for the PR :+1: 
And sorry for the late response, I have totally missed it.

Looks good to me and I also learned something new about image candidates.
Happy to get such a change into Sphinx-Needs.

Could you also add a small entry into the `docs/changelog.rst`, and I think a tiny hint in the related directives about the automatically selected image format would also be good :)

tim-nordell-nimbelink:
Changelog and docs updated.

I saw this was mentioned in another issue but in the future it'd be nice if this supported detecting the document type, and emitted the appropriate image directly into the target build folder.  I haven't tried to fix that here since I can live with the images in the source folder for now, but I wonder if one could create a new node type that has the various visitors implemented (e.g. html, latex, texinfo, etc).  The node type could contain the "in progress" figure, and then once the visitor is invoked for the node, it could at that point emit the file to the filesystem.  Some of the other types that render to a figure (like mermaid) generate the image during the visitor (see https://github.com/mgaitan/sphinxcontrib-mermaid/blob/master/sphinxcontrib/mermaid.py).

tim-nordell-nimbelink:
Added the following changes:
- Rebased to `master`
- Added the new vector files to the .gitignore for the doc folder
- Fixed the test case by changing it to look for `need_pie_dba00.svg` instead of `need_pie_dba00.png` in the hard coded string in the test case.

danwos:
I have made some tests with this PR and built the Sphinx-Needs docs:

With this PR: 33s build time, 2.7 MB `_images` folder size
Just master: 32s build time, 1.6 MB `_images` folder size.

As our documentation is relatively small compared to the ones I know from some companies (~25k rst files, 30min build time), I'm afraid that this change would have a negative impact at least on the `_images` folder size and maybe on the build time by ~5-10%.

Do you see any way to make this part more concrete?
https://github.com/useblocks/sphinx-needs/blob/49d8a3bc8a8a0f7d7c03f7b8336f115d7a36018a/sphinx_needs/utils.py#L359-L364
So creating the files only if needed?
And if this can't be done automatically, can it be a user option in the `conf.py`, so that the user can decide what kind of format to use?

You mentioned the visitors creating the files in the needed format. 
I somehow try to avoid visitors, as they are builder specific and it would be up by us to provide a solution for each possible builder, even the special ones, which get introduced by extensions (like the `needs` builder.)

Could we just ask during the above code, what format makes sense? Roughly like:
```python
#Pseudo code!
   
if app.builder == 'PDF':
      file_formats = [("pdf", "application/pdf")]
else:  # For all other (HTML, latex, needs)
      file_formats = [("svg", "image/svg+xml"), ("png", "image/png")] # multiple formats still supported, if needed

 for ext, mimetype in file_formats:
        rel_file_ext_path = f"{rel_file_path}.{ext}"
        candidates.update({mimetype: rel_file_ext_path})
        if rel_file_ext_path not in env.images:
            figure.savefig(os.path.join(env.app.srcdir, rel_file_ext_path))
            env.images.add_file(fromdocname, rel_file_ext_path)
```

tim-nordell-nimbelink:
I'll see what I can dig out.  I mean, ultimately, it'd be best if it also was written into the build folder, too.

FYI - you can see the list of "supported image types" per builder here: https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.html.StandaloneHTMLBuilder.supported_image_types.  (Each builder has a supported_image_types property.  It also happens to be in preferential order, so I'll see if it's possible to get at that information at the point the image generator runs.

tim-nordell-nimbelink:
I modified it to generate the "best" output format supported by the builder by inspecting the `supported_image_types` property of the builder.  If there aren't any matches, it falls back to the older behavior.

I also added a 3rd commit to the series that writes the images into the output folder directly.  This appears to work for the `html` and `latexpdf` builders; I haven't tried any others.  That could be split into a second PR that is dependent on this one, if preferred, but each commit is a single logical unit of changes:

1. Move the save plot code into a common spot
2. Select the best image output type based on the builder
3. Save into the output folder

danwos (approved):
Thanks for the changes, I like them a lot. :+1: 
As soon as the CI is green this PR gets merged via a merge commit

danwos:
I was a little bold and updated the PR to contain updated poetry.lock file and removed `py` as a dependency (I saw no code using it.). Thi was needed, as the CI claims about dependency conflicts in the `poetry.lock` file.

danwos:
PR looks good now. Thanks again for your help here :1st_place_medal: 
Feel free to add yourself to the `AUTHORS` file if you like.

PR #817, opened 2022-11-26
Enhance uml flow

adegroote:
A little proposal to allow more customization on needuml flow:

* recursive=True permits to generate automatically the parts and childs of the underlaying need. False by default so it does not change the actual behavior.

* color and style allows to override the default color and style without the burden to write the whole logic by hand.

If you feel it interesting, I will update the documentation accordingly.

### DIFF
--- a/sphinx_needs/directives/needflow.py
+++ b/sphinx_needs/directives/needflow.py
@@ -1,6 +1,7 @@
 import html
 import os
 import re
+from functools import partial
 from typing import Iterable, Sequence
 
 from docutils import nodes
@@ -166,10 +167,7 @@ def get_need_node_rep_for_plantuml(
 
 
 def walk_curr_need_tree(
-    app: Sphinx,
-    fromdocname: str,
-    current_needflow: dict,
-    all_needs: list,
+    render_func,
     found_needs: list,
     need: dict,
 ):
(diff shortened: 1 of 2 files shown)


PhilipPartsch:
I understand your idea. But I do not like to make flow to specific. You can store specific architecture view with needarch in the need you want to show and reflect their child's and parts customer specific there; and with ref even with links.
You can use templates to define them even once, so they occur for all your elements of the same type. 
With the config parameter needs_render_context you can even define your own functions to implement such a specific graphical function and you can use it in needuml and needarch.

adegroote:
Thanks for your feedback. I don't know that `needs_render_context` allows to use any free functions. Maybe we just want to have the part which make walk_on_tree more generic, so one can use it in their functions (instead of fully reimplementing it).

adegroote:
@PhilipPartsch by the way, is there any way to access to app object in needs_render_context (to be able to reuse some function such as `compute_link`) ? 

My current goal is to "customize" needflow to display both functional decomposition and organic decomposition (either using color code for each subsystem or put all the spec from a subsystem into some "package") ?

PhilipPartsch:
@adegroote I'm currently working on this. The idea is to have a config parameter like `needs_render_context` but it extends [class JinjaFunctions](https://github.com/useblocks/sphinx-needs/blob/56b4835ef6b3044e9bbde59681998bfa247fb549/sphinx_needs/directives/needuml.py#L218)
In this class, you would have access to everything, even you get the knowledge which is the parent need of your `needarch`.

danwos:
@adegroote: Is this PR still valid, or are the suggested solutions also useable for your use case?

adegroote:
I have the feeling that proposal of @PhilipPartsch would fit my use case (when it is ready) (even through I fear that I will almost duplicate the logic of needflow on my side, but it is a different problem). We probably can close this one so.

PR #820, opened 2022-11-28
needs_external_needs: Skip setting prefixes if no prefix was given

Rubyfi:
With the current implementation, `import_prefix_link_edit` always tries to set a prefix for external needs, even when no prefix is configured for them.
This leads to unnecessary performance problems, especially if the imported needs file is large.
This PR fixes the issue by checking if a `id_prefix` is set first.


twodrops:
@Rubyfi Does this mean even with the fix, the performance problem will reappear the moment we use `id_prefix `?

danwos:
Thanks, got merged :+1:

Issue #278, opened 2021-05-27
Latex output (PDF) no margins around need content


mihnen:
When I generate a pdf build there are large borders around the title and tags but the actual need content has no margins and looks terrible. This is the main content item that readers need to see... can this be fixed?

![2021-05-27 10_22_13 - Adobe Acrobat Reader DC (32-bit)](https://user-images.githubusercontent.com/2370211/119853271-aa55e700-bed5-11eb-97ac-4f0ed2c0c14d.png)

danwos:
I'm not a Latex/Pdf expert and sphinx-needs itself does not set any layout related information for these output formats. So the extension itself can not fix it.

But as far as I know Sphinx itself has some ways to controll the Latex/Pdf output.
I have to make some research for this....

tim-nordell-nimbelink:
@danwos This might be fixed with https://github.com/tim-nordell-nimbelink/sphinx-needs/tree/fix/extra_blanks_in_latex_needs.  The `nodes.line_block()` context, for whatever reason, was causing a lot of extra vertical whitespace within the PDF output from LaTeX.  I don't know if it breaks other things in the HTML side since I had to also drop one of the classes being emitted there for the LaTeX rendering to still work.

I'll note separately, that the built-in `complete` view is broken in the LaTeX rendering.  I don't know how this patch impacts any HTML views other than "clean".

danwos:
Thanks for the hint with `nodes.line_block()`. :+1: 

I tried to create a PR from  https://github.com/tim-nordell-nimbelink/sphinx-needs/tree/fix/extra_blanks_in_latex_needs, but it looks like it contains some other changes as well.
Can you create a PR for it, so that our CI can run the tests and we see the impact on the HTML build quite easily?

tim-nordell-nimbelink:
@danwos Sure, I'll create a PR.  That branch listed just has one change on top of tag `1.1.0`.  (Master in my fork has the 3 sets of changes I've made merged together.)

tim-nordell-nimbelink:
I guess I'll additionally note that I removed some of the nested `<span>` elements in the output that didn't have a new class in the output.  That isn't technically required as a minimal fix for the LaTeX output.  (The main thing for the LaTeX output was a class spanning multiple paragraphs, and to get rid of the `nodes.line_block()` context and related classes.)

PR #823, opened 2022-11-30
latex output: Fix LaTeX output format for needs tables

tim-nordell-nimbelink:
This adds the `colwidths-given` class to the table generated with a simple needs table.  The LaTeX output then actually checks the <colspec> blocks for each column and respects the widths specified.

Additionally, the need table was originally generating a <line-block> context and populating it with <line> elements.  This ends up adding extra spacing into the output for whatever reason and took up a lot of vertical space for each need in the rendered LaTeX PDF output.  This has been removed and instead each line is placed into a new paragraph context via nodes.paragraph().  This renders the output correctly, and the HTML side is captured via CSS to ensure that there are no margins around the paragraph.

Other attempts to solve the rendering didn't work.  If I moved the class into a nodes.inline(...) element, the LaTeX renderer doesn't like having a nested paragraph inside of a class defined via `inline(...)`.  Placing a pair of newlines, defining a new paragraph that way, also yielded the same error from the LaTeX renderer.  (I also tried using `nodes.container(...)` in lieu of `nodes.inline(...)`, but this added an extra trailing line in each section of the simple table.)

Note: Currently this breaks some of the test cases, and I haven't investigated these.  (I was more interested in figuring out how to fix the LaTeX output.)

Potentially fixes https://github.com/useblocks/sphinx-needs/issues/278.

### DIFF
--- a/sphinx_needs/css/common.css
+++ b/sphinx_needs/css/common.css
@@ -21,6 +21,10 @@ div.needstable_wrapper {
     padding: 0;
 }
 
+p.needs_table_row {
+    margin: 0 auto;
+}
+
 /* We need not line-block margin:24 px here, which is coming from sphinx theme */
 table.need td div.needs_side.line-block{
     margin-bottom: 0;
--- a/sphinx_needs/layout.py
+++ b/sphinx_needs/layout.py
@@ -187,6 +187,7 @@ class LayoutHandler:
             classes.append("needs_style_none")
 
         classes.append("needs_type_" + "".join(self.need["type"].split()))
+        classes.append("colwidths-given")
 
         self.node_table = nodes.table(classes=classes, ids=[self.need["id"]])
         self.node_tbody = nodes.tbody()
@@ -309,11 +310,10 @@ class LayoutHandler:
         if len(lines) == 0:
             return None
 
-        lines_container = nodes.line_block(classes=[f"needs_{section}"])
+        lines_container = nodes.inline()
 
         for line in lines:
-            # line_block_node = nodes.line_block()
-            line_node = nodes.line()
+            line_node = nodes.inline()
 
             line_parsed = self._parse(line)
             line_ready = self._func_replace(line_parsed)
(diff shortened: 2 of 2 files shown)


danwos:
Only 4 failing tests and the Sphinx-Needs documentation looks good.
Also our JS code for the collapse button still works.

So I would say this PR looks really promising.

Are you also able to fix the test case or do you need some support here?

For a final PR, we also need an entry in the ``docs/changelog.rst`` file, and for sure, feel free to add yourself to the ``AUTHORS`` file.

Thanks for your help :+1: 
LaTeX skills are also welcome, as it was not the favored builder in the past.

tim-nordell-nimbelink:
I did some further testing, and it breaks some of the custom layout stuff.  Specifically, multiple lines within the custom layouts are broken since I took out the outer-most `nodes.line()` that was getting created.

The LaTeX output can also recognize two newlines in a row in the text output as a paragraph.  When testing with this, the last double newline ends up creating an extra line on the last line within a given section.  I think what should ultimately happen is all individual lines should end up in a single, non-nested, `nodes.container()`, and then insert the double newlines.  This would yield a `<div>` element in HTML, and the appropriate new paragraph.  Alternatively, each could end up in a single `nodes.paragraph()` element.  The `nodes.container()` allows for nested inline classes, though, which is more ideal for the HTML side of things.  The `nodes.paragraph()` does not allow nested inline classes.

I'll leave this additional refactoring for a future patch.

danwos:
I would like to close this PR, as it seems outdated and nobody was working on it.

Especially the current implementation changes some HTML code and breaks some stuff.
As PDF-Alternative there is also[ Sphinx-SimplePDF](https://sphinx-simplepdf.readthedocs.io/en/latest/), which works quite well with Sphinx-Needs as it is using an HTML build for the PDF generation.

tim-nordell-nimbelink:
@danwos -

Sure, we can close this.

- Tim

Issue #818, opened 2022-11-28
Needextract does not work with images


lehaas:
**Describe the bug**
We want to add images to the content of `need`s and reference these `need`s with `needextract`. Unfortunately, the images are not correctly parsed when extracting the need.

I assume it is related to https://github.com/useblocks/sphinx-needs/pull/794. However, the PR only solves the problem if images are located in `source/_images`. Where the images will also be located in `build`. 

**To Reproduce**
1. Create `projectA`:
   ```bash
   mkdir projectA
   cd projectA
   sphinx-quickstart 
   ```
   I choose to separate source and build folder during the build, i.e., `Separate source and build directories (y/n) [n]: y`
2. Configure `projectA` by adding the following lines to `projectA/source/conf.py`:
   ```python
   extensions = ["sphinx_needs"]
   ```
3. Add an image to the source folder, i.e.. `projectA/source/image.png`.
4. Create a requirement by adding the following lines in `projectA/source/index.rst`:
   ```rst
   .. need:: NeedA
      :id: NEEDA
   ```
5. Link the image with `needextract`:
   ```rst
   .. needextract:: NEEDA
   ```

**Expected behavior**
The requirement `NEEDA` appears twice and contains the used image twice.

**Actual behaviour**
For the first `need`, displays the image, the second `need` displays the path in `source/index.rst`
![image](https://user-images.githubusercontent.com/19656299/204226537-cdb9c87c-13ed-46ce-a62b-a21b0dc6e1cf.png)

Before using the most recent `sphinx-needs` release, the path was `_images/image.png`, i.e., the image copy in `build`. This broken as soon as the rst file was not on the `source` level but in a subfolder. 

**Setup**
We are using python Python `3.8.13`. And the following version of sphinx/sphinx-needs:
```bash
sphinx-needs                  1.1.0
Sphinx                        5.3.0
```

I hope the description helps to reproduce the issue :) 

The issue template was copied from [labs-zola](https://github.com/NYCPlanning/labs-zola).

iSOLveIT:
Hi @lehaas,
First of all, I would like to thank you for the issue report.
I read through the description you gave on how to reproduce the issue and would like to clarify some things.

> We want to add images to the content of `needs` and reference these `needs` with `needextract`. Unfortunately, the images are not correctly parsed when extracting the need.
[...]

PR #825, opened 2022-12-06
Bugfix needextract image/download handling

danwos:
Fix for #818


tests/doc_test/doc_needextract/subfolder/file_1.rst
```diff
@@ -0,0 +1,65 @@
...
+
+.. story:: Subfolder example with image in same folder (direct relative path)
+   :id: US_SUB_004
+
+   Use a relative path for the image
+
+   .. image:: smile.png
+
+   Path: ``smile.png``
+
```
lehaas, line 39: I think a testcase using absolute paths would be good for completeness. In addition, the user story would need to be extracted in both `index.rst` and `file_1.rst`.

The intention is to test absolute paths outside of `_images`. Image could also be in a new folder, but I am not sure if that would be necessary.

```suggestion
.. story:: Subfolder example with image in same folder (indirect absolute path)
   :id: US_SUB_005

   Use a relative path for the image

   .. image:: /subfolder/smile.png

   Path: ``/subfolder/smile.png``

```

danwos: Thanks, indeed this was missing. :+1: 
Added it.

PR #833, opened 2022-12-10
use 'replace_self'

danieleades:
I *think* these two forms are equivalent. Let's see if CI agrees


danieleades:
i guess they're not equivalent...

danwos:
Maybe the reasons are the siblings, which get not replaced by self_replace.

PR #834, opened 2022-12-10
move to python 3.8+

danieleades:
sphinx 6.0 beta is out, and drops support for python <3.8.
sphinx-needs can follow suit (once the stable release is out)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,19 +1,8 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 22.12.0
     hooks:
       - id: black
-
-  - repo: https://github.com/PyCQA/flake8
-    rev: 3.9.2
-    hooks:
-      - id: flake8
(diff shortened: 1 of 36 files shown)


danwos:
I would like to close this PR, as it seems to outdated.
Any concerns?

chrisjsewell:
> I would like to close this PR, as it seems to outdated. Any concerns?

@danwos to bump to new versions of sphinx, this would still be required. Python 3.7 is also end-of-life: https://devguide.python.org/versions/

What are your thoughts on bumping python/sphinx versions? Speaking with @ubmarco he suggested that we are less "aggressive" on bumping versions, since some businesses have to deal with older versions of Python?

ubmarco:
We also have the [Compatibility with Sphinx 7.2](https://github.com/useblocks/sphinx-needs/pull/1045) PR now.
It contains new version constraints and remarks wrt API incompatibilities.

chrisjsewell:
superceded by #1056

Issue #847, opened 2023-01-01
Update jsonschema dependency


aacebedo:
Is it possible to update the jsonschema dependency ?
It is still on 3.2.0 and it prevents me to package `sphinx-needs` for NixOS.

3.2.0 is from 2019, current version is 4.17.3.

danwos:
Thanks for asking, and yes this should be possible.
I see no reason to pin it this hard to 3.2.0.

aacebedo:
great ! I made the patch temporarily in the nix package and it seems to work. I simply change =3.2.0 to >=3.2.0

danwos:
Done with PR #848

PR #848, opened 2023-01-02
Allowing newer versions of jsonschema

danwos:
Fix for #847

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -26,21 +26,22 @@ files = [
 
 [[package]]
 name = "attrs"
-version = "22.1.0"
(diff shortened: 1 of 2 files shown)


bluenote10:
Thanks for that fix!

Would it be possible to make a new release of sphinx-needs with this adjustment? The issue is that the current pinning of `jsonschema=3.2.0` is hard-coding sphinx-needs to a very old jsonschema version (from December 2019), which makes it impossible to properly use sphinx-needs in combination with other packages that have more modern minimum requirements on jsonschema.

PR #858, opened 2023-01-23
Attempt to prepare a 1.2.0 release

bluenote10:
Following up on the question in https://github.com/useblocks/sphinx-needs/pull/848#issuecomment-1400210226:

It looks like there are some nice CI pipelines for automatically releasing to PyPI, so perhaps something like this is sufficient to make small patch release?

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -5,15 +5,15 @@ on:
   workflow_dispatch:
     inputs:
       # Ask if docker deployment needed when manually triggered
-      deploy:      
+      deploy:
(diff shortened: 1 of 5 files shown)


danwos:
Thanks for your effort here.
Looks good, but I would like to raise the version to 1.2.0, because the need function "list2need" is part of it.
I know we haven't followed such a rule in the past, but I think it makes sense to start with it.

After this PR got merged, I will create a tag and the CI will release it.

bluenote10:
Great, thanks! I've bumped the version to 1.2.0 now. I will mostly be offline the next days, so feel free to take over if need be.

PR #868, opened 2023-02-07
Fixed pygls v1.0 module import error

haiyangToAI:


sphinx_needs/lsp/esbonio.py
```diff
@@ -11,15 +11,29 @@
...
-    CompletionItemKind,
-    InsertTextFormat,
-    Location,
-    Position,
-    Range,
-    TextEdit,
-)
+from pygls import __version__
+
+if __version__ < "1.0":
```
danwos, line 16: I think this will not work.
As "1.0" is a string and not something numeric, where you can use <.

Maybe the above solution from here works better:
https://stackoverflow.com/a/11887885/1096526

danwos (changes requested):

sphinx_needs/lsp/esbonio.py
haiyangToAI: It works. pygls \__version__ is a string, not numeric.  https://github.com/openlawlibrary/pygls/blob/master/pygls/__init__.py

danwos: Yes, but the string comparison checks the first character only. 
There are a lot of cases where this does not work and produces wrong results:
image

To be save, we need to parse the version-string and get a safe, comparable version-object out of it.

danwos (changes requested):
I think we need to add `packaging` package to the `pyproject.toml` file as a dependency.

In most cases, it should be already there, as it is needed by setuptools. 
But just to be sure :)

danwos (approved):
Looks good, thanks for the quick fix.

PR #869, opened 2023-02-08
Put a space after filter string in example usage

mjsir911:


danwos:
Thanks for the PR and the documentation fix :+1:

PR #887, opened 2023-03-02
Add missing commas in configuration example

holmboe:


danwos:
Thanks for the documentation fixes and the PR :+1:

Issue #893, opened 2023-03-15
Esbonio Hover feature is broken because positon in not an iterable object


aniketsalve22:
https://github.com/useblocks/sphinx-needs/blob/6d7740f0c13aa4c089a4a2747c1ae91cd305131a/sphinx_needs/lsp/esbonio.py#L245

When I hover over a directive in rst file it throws following error

![image](https://user-images.githubusercontent.com/14127049/225291335-5b2cbe5c-487b-443c-b958-c1ae75c46b2e.png)

Versions:
sphinx-needs=1.2.2
esbonio=0.16.1

Old implementation of pygls had Position class with iterable functionality, but the new Postion class from lsprotocol is not iterable.
So the  position class must be unpacked manually

Reference:
Old pygls Position class: https://github.com/openlawlibrary/pygls/blob/fb80b8c40bc1e92a15f03ab79316abd5712fbd4b/pygls/lsp/types/basic_structures.py#L114 

New lsprotocol Position class:
https://github.com/microsoft/lsprotocol/blob/5190589485854de246bf0045a9bc52533899592e/packages/python/lsprotocol/types.py#L5480

chrisjsewell:
this module is no longer part of sphinx-needs

PR #898, opened 2023-03-22
Fix esbonio hover issue

Rubyfi:
With pygls 1.0 the API for lsprotocol changed (see #893 for further details), which leads to issues with esbonio's hover feature.
This PR fixes the error.

### DIFF
--- a/sphinx_needs/lsp/esbonio.py
+++ b/sphinx_needs/lsp/esbonio.py
@@ -242,7 +242,8 @@ def get_lines(ls: NeedlsFeatures, params: Union[CompletionContext, DefinitionCon
 
 def get_word(ls: NeedlsFeatures, params: Union[CompletionContext, DefinitionContext, HoverContext]) -> str:
     """Return the word in a line of text at a character position."""
-    line_no, col = params.position
+    line_no = params.position.line
+    col = params.position.character
     lines = get_lines(ls, params)
     if line_no >= len(lines):
         return ""
--- a/sphinx_needs/directives/needimport.py
+++ b/sphinx_needs/directives/needimport.py
@@ -127,7 +127,7 @@ class NeedimportDirective(Directive):
             if filter_string is None:
                 needs_list_filtered[key] = need
             else:
-                filter_context = {key: value for key, value in need.items()}
+                filter_context = dict(need)
 
                 # Support both ways of addressing the description, as "description" is used in json file, but
                 # "content" is the sphinx internal name for this kind of information


Rubyfi:
Good point.
It should be compatible with the old and the new version of pygls.
With pygls 1.0, the Position class lost the iterable interface, which served as a wrapper to pack/unpack the line & character fields.
Since I have the check anyway, I'll take some time to test whether it works with the old version as well.

Rubyfi:
It seems something is off with the linter checks.
The error that blocked the PR was unrelated to my change.
With the new fixes, even more linter issues popped up.

Btw. I quickly checked, whether there's a difference between using `dict()` and list comprehension to copy the need.
It does not seem to alter the behavior:
```
>>> d1 = {"a": {"aa" : 1}, "b": 2}
>>> d2 = {key: value for key, value in d1.items()}
>>> d3 = dict(d1)
>>> d1
{'a': {'aa': 1}, 'b': 2}
>>> d2
{'a': {'aa': 1}, 'b': 2}
>>> d3
{'a': {'aa': 1}, 'b': 2}
>>> d1["b"] = 21
>>> d1
{'a': {'aa': 1}, 'b': 21}
>>> d2
{'a': {'aa': 1}, 'b': 2}
>>> d3
{'a': {'aa': 1}, 'b': 2}
>>> d1["a"]["aa"] = 42
>>> d1
{'a': {'aa': 42}, 'b': 21}
>>> d2
{'a': {'aa': 42}, 'b': 2}
>>> d3
{'a': {'aa': 42}, 'b': 2}
```

sphinx_needs/lsp/esbonio.py
```diff
@@ -242,7 +242,8 @@ def get_lines(ls: NeedlsFeatures, params: Union[CompletionContext, DefinitionCon
 
 def get_word(ls: NeedlsFeatures, params: Union[CompletionContext, DefinitionContext, HoverContext]) -> str:
     """Return the word in a line of text at a character position."""
-    line_no, col = params.position
+    line_no = params.position.line
```
danwos, line 245: I guess this is not backward compatible, or?

Thinking about something like:
```python

if pyglsversion < 123:
    line_no, col = params.position
else:
    line_no = params.position.line
    col = params.position.character
```

danwos (changes requested):
Thanks for the dict check  :+1: 

Regarding CI, it looks like a new flake8-version introduced some new checks. 
I hopefully find some time soon to fix this. The errors are all the same and not really problematic.

sphinx_needs/lsp/esbonio.py
Rubyfi: In the old implementation `params.position` is just a shortcut for `params.position.__iter__` which returns  `(params.position.line, params.position.character)`.
See [here](https://github.com/openlawlibrary/pygls/blob/fb80b8c40bc1e92a15f03ab79316abd5712fbd4b/pygls/lsp/types/basic_structures.py#L174-L175).
But I can make it more explicit, if that's better for you.

danwos: Yeah, lets make it more explicit.

I also thinks the CI stuff got fixed, and the PR needs to be rebased.
Then everything should be good.

twodrops: Can we merge this one already?
Would be nice to have a 1.3.0 release with this fix.

Rubyfi: Sorry, I got a little swamped lately. I'll update tomorrow.

Rubyfi:
> Can we merge this one already? Would be nice to have a 1.3.0 release with this fix. But 1.3.0 does not need pygls anymore right?

It seems you are correct, pygls as been dropped, so the hover issue no longer exists.

twodrops:
I tested with sphinx-needs 1.3.0 on master and can confirm that this issue is not present anymore.

> sphinx==6.0.1
sphinx-needs==1.3.0 (to be released)
sphinx-rtd-theme==1.2.2
esbonio==0.16.1

PR #930, opened 2023-05-25
Update need_ref.py for templating of external needs

mawieland:
update the text generation when using roles to reference for external need objects.


danwos:
Closed because of inactivity

PR #932, opened 2023-06-01
CI: Docker build arm images as well...

cpolzer:
Hello,

working on my ARM Macbook, i quickly discoverd that the sphinx-needs docker image is not usable , for example as a devcontainer in VSCode.

In context of the [sphinx docker pipeline](https://github.com/sphinx-doc/sphinx-docker-images/blob/4cd233f9c5222ab726439354bdc626248b948c1b/.github/workflows/build-ci.yml#LL40C9-L40C45), i thought this might be the correct solution, but was not able to test it as the main branch build is not working out of the box within my fork.

Anyhow, I hope we somehow get enabled to consume arm based docker images.

Regards!

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -67,6 +67,8 @@ jobs:
         with:
           push: ${{ github.event_name != 'pull_request' && steps.deploycheck.outputs.value == 'y' }}
           file: docker/Dockerfile
+          context: ci
+          platforms: linux/amd64,linux/arm64
           build-args: |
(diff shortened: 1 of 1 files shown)


danwos:
Thanks for the PR.
Have no access to an ARM system, but the change looks promising. So let's take it :)

danwos (approved):

cpolzer:
@danwos  Would we need to have some kind of dry run build (without push) for the docker images?

danwos:
Not sure, as I'm not a docker expert and don't know what a good docker-testing setup should look like.
But I'm totally open to it and don't see any arguments why not.

PR #949, opened 2023-07-18
Added config allow unsafe filter for filter_func

### LINKED ISSUE #831: Allow "dirty" filter results for filter functions
Currently, Sphinx-Needs checks the returned needs of filter functions and takes only their id.
Which is then used as a reference to the original need objects.

So it is impossible to use a filter function to add additional context or values to the filter result.

We should introduce a new option ``allow_unsafe`` or config var ``needs_allow_unsafe_filters``, which can be used to force Sphinx-Needs to use the returned filter data as it is.

### use case
I'm trying to create a ``needtable``, which shall contain data, which is not part of the shown need directly.

I have something like this:

```rst
.. feature:: My cool feature
   :id: FEATURE_1


.. usecase:: A use case 1
   :id: USECASE_1
   :requires: FEATURE_1
   :ti: 1

   .. needtable::
     :filter-func:  feature_filter

.. usecase:: A use case 2
   :id: USECASE_2
   :requires: FEATURE_1
   :ti: 2
   
   .. needtable::
      :filter-func:  feature_filter

```
``FEATURE_1`` is reused in different usecases, which have different `ti` values.
Depending on this `ti` value, a linked ``feature`` gets its own, calculated value.

So this value changes, depending on which usecase is using the feature and representing it 
in a needtable.

haiyangToAI:
fix #831

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -520,6 +520,19 @@ The defined extra filter data can also be used like:
       :style: green_border
 
 
+.. _needs_allow_unsafe_filters:
+
+needs_allow_unsafe_filters
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Allow unsafe filter for :ref:`filter_func` if set to True, which means the filtered results
+can contain manipulated needs.
+
+.. code-block:: python
+
+   needs_allow_unsafe_filters = True
+
+Default is ``False``.
 
 .. _needs_flow_show_links:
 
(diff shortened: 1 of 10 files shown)


docs/configuration.rst
```diff
@@ -520,6 +520,19 @@ The defined extra filter data can also be used like:
       :style: green_border
 
 
+.. _needs_allow_unsafe_filters:
+
+needs_allow_unsafe_filters
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Allow unsafe filter for :ref:`filter_func` if set to True, which means the filtered results
```
ubmarco, line 528: Proposal for an update:
```
Allow unsafe filter for :ref:`filter_func`.
If set to True, the filtered results will keep all fields as they are returned by the dynamic functions.
Fields can be added or existing fields can even be manipulated. Keep in mind this only affects
the filter result, original needs as displayed somewhere else are not modified.

If set to False, the filter results contains the original need fields and any manipulations of need fields are lost.
```

ubmarco (changes requested):

ubmarco (approved):
looks good to me

PR #955, opened 2023-08-03
⬆️ Drop Python 3.7

chrisjsewell:
Inline with 3.7 end-of-life https://devguide.python.org/versions/,
and also fixes issue with sphinx-immaterial.

Note, this is a minimal update to pass tests, which does not modify any code or type annotations, unlike #834.
This can be done in a follow-up PR


docs/api.rst
```diff
@@ -11,7 +11,7 @@ to keep the final configuration transparent for the Sphinx project authors.
 
 For some implementation ideas, take a look into the Sphinx extension
 `Sphinx-Test-Reports <https://sphinx-test-reports.readthedocs.io/en/latest/>`_ and its
-`source code <https://github.com/useblocks/sphinx-test-reports/blob/master/sphinxcontrib/test_reports/test_reports.py#L51>`_.
+`source code <https://github.com/useblocks/sphinx-test-reports/blob/master/sphinxcontrib/test_reports/test_reports.py>`_.
```
chrisjsewell, line 14: This link was failing to find the anchor in linkcheck

noxfile.py
```diff
@@ -56,10 +56,6 @@ def tests(session, sphinx, docutils):
 @session(python="3.10")
 def linkcheck(session):
     session.install(".")
-    # LinkCheck can handle rate limits since Sphinx 3.4, which is needed as
-    # our doc has too many links to GitHub.
-    session.run("pip", "install", "sphinx==3.5.4", silent=True)
```
chrisjsewell, line 61: This is old and overriden in the requirements.txt

PR #961, opened 2023-08-04
✨ NEW: Add `NeedsLookUpTableBuilder`

nhatnamnguyengtvthcm:
1. add builder called NeedsLookUpTableBuilder
2. Builder NeedsLookUpTableBuilder generates needs_lut.json file include {id: docs_name_path}


docs/conf.py
```diff
@@ -380,6 +380,9 @@ def custom_defined_func():
 # build needs.json to make permalinks work
 needs_build_json = True
 
+# build needs_lut.json to make permalinks work
+needs_lut_build_json = True
```
danwos, line 384: I think this is not needed for our Sphinx-Needs docs?

docs/configuration.rst
```diff
@@ -1902,6 +1902,7 @@ an absolute path (on the web server) or an URL.
 
 Default value: ``needs.json``
 
+You can choice option ``needs_lut.json`` as detail :ref:`needs_lut_build_json`
```
danwos, line 1905: Please check such sentences with Grammarly or other online tools for correctness.

docs/configuration.rst
```diff
@@ -2338,3 +2339,22 @@ If true, need options like status, tags or links are collapsed and shown only af
 Default value: True
 
 Can be overwritten for each single need by setting :ref:`need_collapse`.
+
+.. __needs_lut_build_json:
+
+needs_lut_build_json
+~~~~~~~~~~~~~~~~~~~~~~~
```
danwos, line 2346: headline too long

docs/configuration.rst
```diff
@@ -2338,3 +2339,22 @@ If true, need options like status, tags or links are collapsed and shown only af
...
+Builds a ``needs_lut.json`` file during other builds, like ``html``. Different from ``needs.json``, ``needs_lut.json`` only include list of key ``id`` and value in list [``docname``, ``external_url``].
+A helpful load data fastly when you need improve performance.
```
danwos, line 2348: Please reformulate

sphinx_needs/builder.py
```diff
@@ -158,3 +158,66 @@ def build_needumls_pumls(app: Sphinx, _exception: Exception) -> None:
         needs_builder.set_environment(env)
 
     needs_builder.finish()
+
+
+class NeedsLookUpTableBuilder(Builder):
```
danwos, line 163: A docstring please, which describes the main goal of the builder.

sphinx_needs/builder.py
```diff
@@ -158,3 +158,66 @@ def build_needumls_pumls(app: Sphinx, _exception: Exception) -> None:
...
+                needs_dict[need["id"]] = need["docname"]
+
+        try:
+            fname = os.path.join(self.outdir, "needs_lut.json")
+            with open(fname, "w", encoding="utf-8") as f:
+                json.dump(needs_dict, f, indent=4)
+        except Exception as e:
+            log.error(f"Error during writing json file: {e}")
+        else:
+            log.info("Needs doc lookup table json successfully created")
```
danwos, line 189: - doc

danwos (changes requested):
Thanks for the PR.
The idea is good and the implementation is promising.
Unfortunately, I have some findings:

I'm missing any documentation of the new builder in the ``buiders.rst`` file.
This should also contain an example of how the final file looks like.

Also a test case is missing for the javascript part.
You can find examples of it inside `/tests/js_test`:
https://github.com/useblocks/sphinx-needs/tree/master/tests/js_test

docs/builders.rst
```diff
@@ -163,3 +163,53 @@ or
...
 
     As an alternative, you can set the config option :ref:`needs_build_needumls` to export the needumls files during each build.
+
+.. _needs_lut_builder:
+
+needs_lut
+---------
+.. versionadded:: 1.4.0
+
+The **needs_lut** builder exports all found needs to a single json file, which only include list of key ``id`` and value of ``docname`` or ``external_url``.
```
danwos, line 173: We should mention what `lut` stands for. Maybe:

``**needs_lut** (Needs Lookup Table``)

docs/builders.rst
```diff
@@ -163,3 +163,53 @@ or
...
+
+The build creates file called **needs_lut.json** inside the given build-folder.
```
danwos, line 175: .. **a** file called

sphinx_needs/builder.py
```diff
@@ -157,3 +158,66 @@ def build_needumls_pumls(app: Sphinx, _exception: Exception) -> None:
...
+            if need["is_external"]:
+                needs_dict[need["id"]] = need["external_url"]
+            else:
+                needs_dict[need["id"]] = need["docname"]
+
+        try:
+            fname = os.path.join(self.outdir, "needs_lut.json")
+            with open(fname, "w", encoding="utf-8") as f:
+                json.dump(needs_dict, f, indent=4)
+        except Exception as e:
```
danwos, line 186: Can we add this part (dict creation + json.dump) also to NeedsList as  ``write_json_lut``?
It's easier to maintain if all storage functions are defined in one place.

danwos (changes requested):
This PR needs to be updated to the latest changes for 1.4.0 (master).
Especially the way of option registration has changed.

+ some other minor findings.

chrisjsewell (changes requested):
Heya @nhatnamnguyengtvthcm, as you can see in #1019 I added snapshot testing, to provide snapshots of the expected data.

Perhaps you could add this to your `test_doc_needs_lut_builder`

As an example, see: https://github.com/useblocks/sphinx-needs/blob/09277e30b404500a7c1fffd7f66949938d36f7cb/tests/test_needs_builder.py#L14

sphinx_needs/config.py
```diff
@@ -252,6 +252,7 @@ def __setattr__(self, name: str, value: Any) -> None:
     # add config for needs_id_builder
     build_json_per_id: bool = field(default=False, metadata={"rebuild": "html", "types": (bool,)})
     build_json_per_id_path: str = field(default="needs_id", metadata={"rebuild": "html", "types": (str,)})
+    lut_build_json: bool = field(default=False, metadata={"rebuild": "html", "types": (bool,)})
```
chrisjsewell, line 255:
```suggestion
    build_json_lut: bool = field(default=False, metadata={"rebuild": "html", "types": (bool,)})
```

This would be consistent with `build_json` and `build_json_per_id`

sphinx_needs/needsfile.py
```diff
@@ -131,6 +131,9 @@ def load_json(self, file: str) -> None:
 
             self.log.debug(f"needs.json file loaded: {file}")
 
+    def add_lut_need(self, version: str, needs_lut: Dict[str, Any]) -> None:
```
danwos, line 134: My idea was a little different.
Instead of adding a custom data structure and use the same writer-function, we could keep the data structure with `add_need()` and just do the output specific formating in a new writer function like `write_lut_json`.
In this case all the code is the same, till it comes to write it to the harddisk.

sphinx_needs/builder.py
danwos: I think we should stay with `add_need` (instead of `add_lut_need`) and transform the needed data in a special writer function `write_json_lut`.

danwos (changes requested):
Thankas for the changes.
I think the builder is more complex as it should be, just see my comment.

I also like the snapshot idea from Chris for the tests.

r-o-b-e-r-t-o:
LGTM. Approved for contribution.

sphinx_needs/needsfile.py
nhatnamnguyengtvthcm: I wonder if we keep formatting JSON and only map need["docname"] = need["external_url"] if need["is_external"] = True, It is easy to reuse flow generate need_lut.json like need.json. Otherwise, I will continue change formatting in *write_lut_json* like {[id] : [docname] or [external_url]}
@r-o-b-e-r-t-o @danwos

r-o-b-e-r-t-o: @nhatnamnguyengtvthcm  Not sure if I got your point. But I think we must have a mapping for internal **and** external needs because otherwise you would bot be redirected correctly by permalink.html in case an external need is referenced.

PR #962, opened 2023-08-04
✨ NEW: Add `NeedsPerPageBuilder`

nhatnamnguyengtvthcm:
1. Create a new builder in builde.py called: NeedsPerPageBuilder
2 . run: make-html-fast -> the need_per_page folder will be gened which has seperated the need with same docs_name


docs/builders.rst
```diff
@@ -163,3 +163,233 @@ or
...
++++++
+
+
+.. code-block:: bash
+
+    sphinx-build -b needs_per_page source_dir build_dir
+
+
+Format with file name: configuration.json
+++++++
```
danwos, line 187: What does this syntax do?

Please build the docs and check for errors before submitting.
A `make docs-html` is now enough, as I fixed the doc-build and made it much easier.

docs/builders.rst
```diff
@@ -163,3 +163,233 @@ or
...
+.. code-block:: python
```
danwos, line 188: Can you move this to  a json file and present it here via `.. literalinclude::`?
Much easier to maintain.

docs/configuration.rst
```diff
@@ -2289,3 +2289,34 @@ If true, need options like status, tags or links are collapsed and shown only af
...
+.. _needs_per_page:
+
+needs_per_page
+~~~~~~~~~~~~~~
+
+.. versionadded:: 1.4.0
+
+Build list of ``json`` files that contain all found needs with the same ``docname``. The name of each file should match the ``docname``.
+
+This option works like needs_build_json.
```
danwos, line 2303: Please use `:ref:` to reference other documentation objects.

docs/configuration.rst
```diff
@@ -2289,3 +2289,34 @@ If true, need options like status, tags or links are collapsed and shown only af
...
+
+
+.. _needs_per_page_build_path:
+
+needs_per_page_build_path 
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 1.4.0
+
+This option sets the location of list of ``json`` files that contain all found needs with the same ``docname``.
```
danwos, line 2315: Is this a folder or a file name?
Is it relative to the `build` folder or to the documentation root?

docs/configuration.rst
```diff
@@ -2289,3 +2289,34 @@ If true, need options like status, tags or links are collapsed and shown only af
...
+
+Default value: need_per_page
```
danwos, line 2317: Please highlight path information like `need_per_page`.
Add also a `/` if it is a folder

sphinx_needs/builder.py
```diff
@@ -157,3 +156,88 @@ def build_needumls_pumls(app: Sphinx, _exception: Exception) -> None:
...
+    def finish(self) -> None:
+        env = unwrap(self.env)
+        needs = env.needs_all_needs.values()
+        filter_string = self.app.config.needs_builder_filter
+        needs_per_page_build_path = self.app.config.needs_per_page_build_path
+        filtered_needs = filter_needs(self.app, needs, filter_string)
+        needs_per_page_dir = os.path.join(self.outdir, needs_per_page_build_path)
+        needs_per_page_data = {}
+        if not os.path.exists(needs_per_page_dir):
+            os.mkdir(needs_per_page_dir)
```
danwos, line 180: This does not work, if there are multiple new folders in the path, e.g. ``needs/pages/examples/``.

Use instead:
https://docs.python.org/3/library/os.html#os.makedirs

sphinx_needs/builder.py
```diff
@@ -157,3 +156,88 @@ def build_needumls_pumls(app: Sphinx, _exception: Exception) -> None:
...
+
+        needs_per_page_data_key = []
+        for need in filtered_needs:
```
danwos, line 183: Some more comments would be good, especially if you have such much code in a single function and haven't it broken down to some sub functions.

At least `for`-loops with multiple lines should get a comment, so that you know in advance what the goal of the for loop is before reading all code line.
Something like `sorting needs into docname-based dict` would be helpful.

sphinx_needs/builder.py
```diff
@@ -157,3 +156,88 @@ def build_needumls_pumls(app: Sphinx, _exception: Exception) -> None:
...
 
     needs_builder.finish()
+
+
+class NeedsPerPageBuilder(Builder):
+    name = "needs_per_page"
+    format = "needs"
+    file_suffix = ".txt"
+    links_suffix = None
+    LIST_KEY_EXCLUSIONS_NEEDS = ["content_node"]
```
danwos, line 166: This is not Builder specific and is shared information that each builder must follow.
So if we need to extend it, we must touch multiple locations. That's not good.

That's one of the reasons why a central ``Needsfile` class was created to deal with such common configs.
https://github.com/useblocks/sphinx-needs/blob/master/sphinx_needs/needsfile.py

tests/test_needs_per_page_builder.py
```diff
@@ -0,0 +1,40 @@
...
+        need_docname_path = os.path.join(needs_per_page_parent_path, need_docname_file)
+        # need_docname = os.path.dirname(docs_name_file)
+        assert os.path.exists(need_docname_path)
+
+        with open(need_docname_path) as needs_file:
+            needs_file_content = needs_file.read()
+            needs_list = json.loads(needs_file_content)
+            assert isinstance(needs_list["needs"], list)
+            need_of_file = needs_list["needs"]
+            all_keys = set().union(*(d.keys() for d in need_of_file))
```
danwos, line 38: Why is this needed?

would a ` need_id in needs_list.keys()` no work?
I think also `need_id in needs_list`  is enough

tests/test_needs_per_page_builder.py
```diff
@@ -0,0 +1,40 @@
...
+        need_docname_file = f"{need_docname}.json"
+        need_docname_path = os.path.join(needs_per_page_parent_path, need_docname_file)
+        # need_docname = os.path.dirname(docs_name_file)
+        assert os.path.exists(need_docname_path)
+
+        with open(need_docname_path) as needs_file:
+            needs_file_content = needs_file.read()
+            needs_list = json.loads(needs_file_content)
+            assert isinstance(needs_list["needs"], list)
+            need_of_file = needs_list["needs"]
```
danwos, line 37: `need_of_file`: Is this only one need per file?
Or better need**s**_of_file?

danwos (changes requested):
Some finding, but overall the code looks okay.

However, with ``Needsfile`` there is a generic Need-Builder class, which cares about setting everything up the right way.
And I think we should use this and just extend it by a function to store the data in a specific way.
https://github.com/useblocks/sphinx-needs/blob/master/sphinx_needs/needsfile.py

Your code doubles a lot of functions, like filtering, data preparation, and some magic configurations.
That's hard to maintain in the future.

What do you think, is updating `Needsfile` possible?

tests/test_needs_per_page_builder.py
nhatnamnguyengtvthcm: the JSON file  look like:
 {
      "needs": [
                       {
                           "R_F1B63": {
                                               "docname": "directives/list2need",
                                               ...
                                               "id": "R_F1B63",
                              }
                      },
                    {
                           "NEED-002": {
                                                     ...
                       }
        ]
}
The ID is a key of element in this list since I get key() to check ID included there. if use **need_id in needs_list** only, i will not work

docs/configuration.rst
nhatnamnguyengtvthcm: It is name of folder

nhatnamnguyengtvthcm:
I wonder do you want the format of the JSON file generated the same **needs.json**.  should  It generate by function in needsfile.

nhatnamnguyengtvthcm:
> Some finding, but overall the code looks okay.
> 
> However, with `Needsfile` there is a generic Need-Builder class, which cares about setting everything up the right way. And I think we should use this and just extend it by a function to store the data in a specific way. https://github.com/useblocks/sphinx-needs/blob/master/sphinx_needs/needsfile.py
> 
> Your code doubles a lot of functions, like filtering, data preparation, and some magic configurations. That's hard to maintain in the future.
> 
> What do you think, is updating `Needsfile` possible?

I wonder do you want the format of the JSON file generated the same **needs.json**.  should  It generate by function in needsfile.

docs/builders.rst
chrisjsewell: Maybe just remove this and use the testing snapshots.
The problem with having it "statically" in the documentation, is that it is prone to become out-of-date with the actual code

docs/conf.py
```diff
@@ -361,6 +361,9 @@ def custom_defined_func():
 # build needs.json to make permalinks work
 needs_build_json = True
 
+# build json file include needs with the same docs_name
+needs_per_page = True
```
chrisjsewell, line 365: This option name should be consistent with the other builders (`build_json`, `build_json_per_id`), i.e. `build_json_per_page`

Thinking holistically, with there now being numerous data writers, and more on the way, maybe it makes sense to have some form of unified configuration option.

For example:

```python
needs_data_writers = ["json", "json_per_id", "json_per_page"]
```

r-o-b-e-r-t-o:
LGTM. Approved for contribution.

PR #964, opened 2023-08-09
👌 Fail and emit warning on filters that do not return a boolean result

big1hc:


danwos:
Thanks for the Bugfix :+1: 

The code itself looks good, but I'm not sure if the check gets performed on the right location.
As we have 
```python
result = bool(eval(filter_string, filter_context))`
```
`result` should be always `True` or `False`.
No matter what `eval()` returns, it is typecasted to a boolean value.

I think the intention of this fix is to check if the provided filter statement is returning a boolean value or something else.
So for me, the check should be performed on the result of `eval()` before `bool()` to be sure the filter statement is working as expected.
Does this cover your use case?

Feel also free to add yourself to the `AUTHORS` file and maybe add a short bugfix-entry to `docs/changelog.rst`.

big1hc:
> Thanks for the Bugfix 👍
> 
> The code itself looks good, but I'm not sure if the check gets performed on the right location. As we have
> 
> ```python
> result = bool(eval(filter_string, filter_context))`
> ```
> 
> `result` should be always `True` or `False`. No matter what `eval()` returns, it is typecasted to a boolean value.
> 
> I think the intention of this fix is to check if the provided filter statement is returning a boolean value or something else. So for me, the check should be performed on the result of `eval()` before `bool()` to be sure the filter statement is working as expected. Does this cover your use case?
> 
> Feel also free to add yourself to the `AUTHORS` file and maybe add a short bugfix-entry to `docs/changelog.rst`.

sphinx_needs/filter_common.py
```diff
@@ -306,9 +306,14 @@ def filter_single_need(
         # Set filter_context as globals and not only locals in eval()!
         # Otherwise, the vars not be accessed in list comprehensions.
         if filter_compiled:
-            result = bool(eval(filter_compiled, filter_context))
+            result = eval(filter_compiled, filter_context)
         else:
-            result = bool(eval(filter_string, filter_context))
+            result = eval(filter_string, filter_context)
+        if isinstance(result, str):
```
danwos, line 312: It makes more sense to check here, if the result is **not** a boolean value, as this is the only allowed return type for a filter.

danwos (changes requested):
Thanks for the update, but we should change the logic for the check to be more bulletproof.

sphinx_needs/filter_common.py
big1hc: I have convert result to boolean when it is returned. I see that when we use the string filter like "search("AAA", title)" -> the `result ` type is re.Match type, this result is not a boolean but it also valid
image

danwos: Mhh then this is a bug for `search()`. And luckily `search()` returns `None`, which gets transformed to `False`, or it returns a re.Match object, which is always `True`.
So we should fix our `search()` implementation as well and I would stay with my opinion, that we should only allow Boolean returns values for `eval()`.

Checking for `string` only solves some specific problems, but is not a generic solution, which detects all faulty filter strings.

sphinx_needs/filter_common.py
```diff
@@ -306,9 +306,16 @@ def filter_single_need(
         # Set filter_context as globals and not only locals in eval()!
         # Otherwise, the vars not be accessed in list comprehensions.
         if filter_compiled:
-            result = bool(eval(filter_compiled, filter_context))
+            result = eval(filter_compiled, filter_context)
         else:
-            result = bool(eval(filter_string, filter_context))
+            result = eval(filter_string, filter_context)
+        if isinstance(result, re.Match):
```
danwos, line 312: I think we should do this not here, but in the search function:
https://github.com/useblocks/sphinx-needs/blob/master/sphinx_needs/filter_common.py#L303

I would add something like this to `filter_common.py`:
```python
def need_search(*args, **kwargs)
   return bool(re.search(*args, **kwargs))

... some other code

 filter_context["search"] = need_search
```

Then the result of `search()` gets transformed when used.
And we can save the check for re.Match here, as we should always get boolean-values back.

Sorry for making your PR a little more complex, but you found a new bug with it and for a generic solution for your problem we need to fix this as well.

danwos (changes requested):

danwos (approved):

r-o-b-e-r-t-o (approved):

danwos (changes requested):
Looks good, jsut add changelog line to the not released version.

sphinx_needs/filter_common.py
```diff
@@ -300,15 +304,20 @@ def filter_single_need(
...
-            result = bool(eval(filter_compiled, filter_context))
+            result = eval(filter_compiled, filter_context)
         else:
-            result = bool(eval(filter_string, filter_context))
+            result = eval(filter_string, filter_context)
+        if not isinstance(result, bool):
+            # Raises NeedsInvalidFilter if the result is a string type
+            raise NeedsInvalidFilter(
+                f"Error when evaluating filter: expected output to have True/False but got a string <{result}>"
+            )
```
chrisjsewell, line 350: If it is not a bool, this does not logically mean that it will be a string though?

sphinx_needs/filter_common.py
```diff
@@ -299,6 +299,10 @@ def filter_needs(
     return found_needs
 
 
+def need_search(*args, **kwargs):
+    return bool(re.search(*args, **kwargs))
```
chrisjsewell, line 303:
```suggestion
    return re.search(*args, **kwargs) is not None
```

this is maybe a bit more specific?

sphinx_needs/filter_common.py
```diff
@@ -299,6 +299,10 @@ def filter_needs(
     return found_needs
 
 
+def need_search(*args, **kwargs):
```
chrisjsewell, line 302:
```suggestion
def need_search(*args: Any, **kwargs: Any) -> bool:
```

chrisjsewell (changes requested):

sphinx_needs/filter_common.py
```diff
@@ -330,15 +334,20 @@ def filter_single_need(
...
-            result = bool(eval(filter_compiled, filter_context))
+            result = eval(filter_compiled, filter_context)
         else:
-            result = bool(eval(filter_string, filter_context))
+            result = eval(filter_string, filter_context)
+        if not isinstance(result, bool):
+            # Raises NeedsInvalidFilter if the result is a string type
+            raise NeedsInvalidFilter(
+                f"Error when evaluating filter: expected output to have True/False but got a string <{result}>"
+            )
```
chrisjsewell, line 350:
```suggestion
            raise NeedsInvalidFilter(
                f"Filter did not evaluate to a boolean, instead {type(result)}: {result}"
            )
```

chrisjsewell (approved):
Thanks!

big1hc:
> Looks good, jsut add changelog line to the not released version.

I added my change to version 1.4.0

r-o-b-e-r-t-o:
@danwos Can this PR be merged?

big1hc:
@danwos the pull request has update-to-date now

chrisjsewell:
Will fix the failing doc build, and I want to add a test, after #1128 😄

danwos (approved):

chrisjsewell:
Ok this is ready to merge, but ...

I have changed the title from a bug fix, to simply a change, since I don't think this fixes a bug, it simply changes sphinx-needs behaviour:
Before, sphinx-needs allowed any truthy/falsy values to be returned from filters, now, it only allows boolean.
I can see how this could stop users from making mistakes with their filters,
although I do fear this will be a breaking change for some users, as it was in our own internal documentation (although at least with #1128 it will now be easier to locate the issue).

@danwos can you just confirm you are happy with this breaking change, then happy to merge (but if you merge it, please do with a commit message, that mentions the breaking nature of the change 😄)

PR #968, opened 2023-08-16
👌 Performance: Memoize Inline parser for `build_need`

chrisjsewell:
As discussed in #967, `build_need` (called from `print_need_nodes`) is currently taking a large amount of build times.

A key reason for this is that the initialisation of the inline parser, which requires compiling numerous regexes, is being run on every call.
This is not required, since the initialisation is invariant.

In this PR therefore, we move the initialisation to a cached function, which will only run once per build.

As shown when using `phinx-analysis  --project needs --pyinstrument --tree --tree-filter 0.01`, compared to #967, this significantly reduces the `print_need_nodes` component of the build time; from taking ~25% to ~8%

In this example it reduces the total build time from ~1.6 seconds to ~1.2 seconds

image

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,3 +35,5 @@ Jörg Kreuzberger <j.kreuzberger@procitec.de>
 Duodu Randy <duodurandy19@gmail.com>
 
 Christian Wappler <chri.wapp@gmail.com>
+
+Chris Sewell <chrisj_sewell@hotmail.com>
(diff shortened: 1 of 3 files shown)


chrisjsewell:
> Can you add also a small line to our /docs/cahngelog.rst file?
Feel also free to add yourself to our AUTHORS file as well.

done @danwos  👍

danwos (approved):
Thanks, looks good 👍

PR #973, opened 2023-08-18
Plantuml debug mode

### LINKED ISSUE #754: Improve plantuml errors
# Current state
When there is some problem in plantuml (e.g. via `needarch`) the only output we get is
```
WARNING: error while running plantuml

b'ERROR\n989\nSyntax Error?\nSome diagram description contains errors\n'
```

# Desired state
* What is the exact error? To my knowledge plantuml provides more info than what we see above! Either as text output or as generated image showing the error message.
* Was there a puml file generated I can look at? `:debug:` doesn't help here since I don't get the generated image. With some jinja and python magic I cannot tell what the plantuml code looks like which produced the error.

danwos:
Tests all plantuml-related directives in advance and in the context of Sphinx-Needs, so that meaningful error messages get written and the problematic code is stored in a temporary file.

Tasks
* [x] Run and check plantuml
* [x] Store failed code in a temporary file
* [ ] docs (80%, file save missing)
* [ ] Test cases

Fixes #754

### DIFF
--- a/sphinx_needs/directives/needflow.py
+++ b/sphinx_needs/directives/needflow.py
@@ -1,14 +1,16 @@
 import html
 import os
 import re
+import tempfile
 from typing import Iterable, Sequence
 
 from docutils import nodes
 from docutils.parsers.rst import directives
 from jinja2 import Template
 from sphinx.application import Sphinx
+from sphinxcontrib.plantuml import _split_cmdargs  # Needed for plantuml execution path
 from sphinxcontrib.plantuml import (
-    generate_name,  # Need for plantuml filename calculation
+    generate_name,  # Needed for plantuml filename calculation
 )
 
 from sphinx_needs.debug import measure_time
@@ -489,11 +491,20 @@ def process_needflow(app: Sphinx, doctree: nodes.document, fromdocname: str, fou
             para += filter_node
             content.append(para)
 
+        # If the central PlantUML Debug option is set, we need to generate the code already once in advance and
+        # report any errors directly.
+        if app.config.needs_plantuml_debug:
+            plantuml_args = _split_cmdargs(app.config.plantuml)
+            plantuml_args.extend(["-stdrpt:1"])
+            with tempfile.NamedTemporaryFile() as tmp:
+                tmp.write(puml_node["uml"])
+                plantuml_args.extend([tmp.name])
+
         # We have to restrustructer the needflow
         # If this block should be organized differently
-        if current_needflow["debug"] and found_needs:
+        if (current_needflow["debug"] or app.config.needs_plantuml_debug) and found_needs:
             # We can only access puml_node if found_needs is set.
-            # Otherwise it was not been set, or we get outdated data
+            # Otherwise, it wss not set, or we get outdated data
             debug_container = nodes.container()
             if isinstance(puml_node, nodes.figure):
                 data = puml_node.children[0]["uml"]
--- a/sphinx_needs/directives/needflow.py
+++ b/sphinx_needs/directives/needflow.py
@@ -1,14 +1,12 @@
 import html
 import os
 import re
-import tempfile
 from typing import Iterable, Sequence
 
 from docutils import nodes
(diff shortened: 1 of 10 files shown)


danwos:
This PR should also fix all PlantUML related problems, if they are not related to Sphinx-Needs generation code. Then we will need an extra PR.

chrisjsewell:
FYI, I think there is a better way to do this, so don't merge just yet (I also fixed the current issue #984)

danwos:
I agree, that your approach is better in the midterm, when the latest sphinxcontrib-plantuml package can be used.

But I think this little change can be helpful now, and as it must be activated by config-var it should not make troubles in normal runs. Or do you see any drawbacks?

sphinx_needs/utils.py
```diff
@@ -581,3 +584,32 @@ def add_doc(env: BuildEnvironment, docname: str, category=None):
             env.needs_all_docs[category] = []
         if docname not in env.needs_all_docs[category]:
             env.needs_all_docs[category].append(docname)
+
+
+def plantuml_debug(uml_type, cmd, code, docname, lineno):
+    """Checks a given Plantuml Code and prints found errors"""
+    try:
```
chrisjsewell, line 591: Perhaps it would be better here to simply use: https://github.com/sphinx-contrib/plantuml/blob/36716425827d26502552f23f1582069878934514/sphinxcontrib/plantuml.py#L230

sphinx_needs/directives/needflow.py
```diff
@@ -489,11 +489,22 @@ def process_needflow(app: Sphinx, doctree: nodes.document, fromdocname: str, fou
             para += filter_node
             content.append(para)
 
+        # If the central PlantUML Debug option is set, we need to generate the code already once in advance and
+        # report any errors directly.
+        if app.config.needs_plantuml_debug:
```
chrisjsewell, line 494: Perhaps it would be better to run this in a transform, that goes through and finds all (sphinx-needs generated) `plantuml` nodes,
rather than having to add it to every different directive

sphinx_needs/directives/needsequence.py
```diff
@@ -211,6 +211,15 @@ def process_needsequence(app: Sphinx, doctree: nodes.document, fromdocname: str,
         if current_needsequence["show_filters"]:
             content.append(get_filter_para(current_needsequence))
 
+        if app.config.needs_plantuml_debug:
+            plantuml_debug(
+                "needflow",
+                app.config.plantuml,
+                puml_node.children[0]["uml"],
+                current_needsequence["docname"],
+                current_needsequence["lineno"],
```
chrisjsewell, line 220: I think just parse the puml_node to the function, which already contains the source information

sphinx_needs/utils.py
```diff
@@ -581,3 +584,32 @@ def add_doc(env: BuildEnvironment, docname: str, category=None):
...
+            tmp_puml.write(code)
+        # output = subprocess.check_output(plantuml_args, shell=True, errors=True, text=True).decode("utf-8")
+        output = subprocess.run(plantuml_args, capture_output=True)
+    finally:
+        status = output.returncode
+        icon = "✓" if status == 0 else "✕"
+        output_text = output.stderr.decode("utf-8").replace("\n", "")
+        print(f"{icon} {uml_type} {docname}:{lineno}: {output_text}")
+        if status != 0:
+            print(f"  uml-file: {debug_file.name}")
```
chrisjsewell, line 606: I t would be better if this was a sphinx warning (with location information)

chrisjsewell (approved):
Heya, see comments cheers

chrisjsewell (changes requested):
oops was meant to be requesting changes

sphinx_needs/utils.py
danwos: But I don't get any information back if problems have happened.

danwos:
A better solution by @chrisjsewell has already been merged.
This is not needed anymore.

PR #975, opened 2023-08-19
👌 Improve sphinx warnings

chrisjsewell:
This PR is intended to standardise and improve the sphinx warnings emitted by sphinx-needs:

1. To make it clear when a warning originates from sphinx-needs
2. Where possible, to add the location of the warning origin in the documentation

This is similar to my work in https://myst-parser.readthedocs.io/en/latest/configuration.html#build-warnings,
it prefixes all warnings with `[needs]`, and adds the `needs` type to the warning, meaning it can be specifically suppressed (see https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings).

I have also added the `location` argument to some warnings (where the location was obvious), although there are probably some more that could also have a location (but this can be done in later PRs)

Note, IMO it would be ideal if the type was automatically shown by sphinx, but this is not currently the case: https://github.com/sphinx-doc/sphinx/issues/8845

Note also, I intend to make an upstream PR to https://github.com/sphinx-contrib/plantuml, to also improve their warnings

### DIFF
--- a/sphinx_needs/api/need.py
+++ b/sphinx_needs/api/need.py
@@ -148,7 +148,8 @@ def add_need(
     if need_type not in configured_need_types:
         logger.warning(
             "Couldn't create need {}. Reason: The need-type (i.e. `{}`) is not set "
-            "in the project's 'need_types' configuration in conf.py.".format(id, need_type)
+            "in the project's 'need_types' configuration in conf.py. [needs]".format(id, need_type),
+            type="needs",
         )
 
     for ntype in types:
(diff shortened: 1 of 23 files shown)


danwos (commented):
Thanks for the PR and the hard work in changing all logger-warnings. :+1: 

Looks nearly good to me, but can you add a `changelog.rst` entry? Thanks

chrisjsewell:
> Looks nearly good to me, but can you add a changelog.rst entry? 

done 👍

danwos (approved):

PR #960, opened 2023-08-04
Namnn/need per id json

nhatnamnguyengtvthcm:
1. In the builder, I build one builder which is name build_needs_id_json to build Json file per need id
2. Using command: make docs-html-fast to run.
3. Finally, in the "_build" folder will be gen the "needs_id" folder include all file json of need id.

PR #978, opened 2023-08-21
♻️ Change `NeedsBuilder` format to `needs`

chrisjsewell:
Changing the format from `json` to `needs` will make it more specific to target the `NeedsBuilder`, and any derivatives such as in #960, #961, #962

The `format` attribute of the sphinx builder is perhaps a bit misleading, since it is mainly used to target a "class" of builders, rather than a specific output format,
see: https://www.sphinx-doc.org/en/master/usage/builders/index.html

This will allow for logic in other parts of the sphinx-needs pipeline, e.g. to skip certain build steps not required by the needs builders.


danwos (approved):
Thanks for the PR :+1:  Looks good.

PR #983, opened 2023-08-24
📚 DOCS: Always use headless plantuml

chrisjsewell:
There should be no reason why a non-headless JVM is ever required
and, without it, sphinx builds are slower and constantly switch window focus


danwos (approved):

danwos:
Thanks for the PR :+1:

PR #984, opened 2023-08-24
🐛 Fix gantt chart rendering

### LINKED ISSUE #977: needgantt generates invalid PlantUML code
Currently, our docs show no gantt-charts:
https://sphinx-needs.readthedocs.io/en/latest/directives/needgantt.html

As the generated PlantUML code is invalid by our used PlantUML version.

AS the gantt-syntax was experimental when ``needgantt`` got developed, it looks like there were some changes in the syntax.

We need to update our generation code to be in sync with the latest PlantUML version.

chrisjsewell:
In plantuml V1.2022.12 (23 Oct, 2022), it became a syntax error not to use `@startgantt`: https://plantuml.com/changes

fixes #977

### DIFF
--- a/sphinx_needs/directives/needgantt.py
+++ b/sphinx_needs/directives/needgantt.py
@@ -149,9 +149,9 @@ def process_needgantt(app, doctree, fromdocname, found_nodes):
             no_plantuml(node)
             continue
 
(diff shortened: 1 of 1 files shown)


danwos:
Thanks fo the quick fix :+1:

danwos (approved):

PR #986, opened 2023-08-25
🔧 Replace `Directive` with `SphinxDirective`

chrisjsewell:
`SphinxDirective` provides a number of additional helper methods, including access to the Sphinx environment


danwos (approved):
Ohh nice, that makes it simpler. 
Thanks for the PR . Approved :+1:

PR #987, opened 2023-08-25
🔧 Centralise sphinx `env` data access (and add schemas)

chrisjsewell:
The sphinx-needs data stored on the Sphinx `BuildEnvironment` is currently difficult to interpret and work with,
since the dynamic nature of `BuildEnvironment` provides no type safety or static introspection.

This PR intends to centralise access to sphinx-needs data, via `sphinx_needs.data.SphinxNeedsData`, which is a thin wrapper around the sphinx `BuildEnvironment`, to define and provide type safe access to the sphinx-need specific data.

[TypedDict](https://docs.python.org/3/library/typing.html#typing.TypedDict) are used to type annotate the dictionary keys for the different data types,
and thus `sphinx_needs.data` essentially provides a schema for the data that sphinx-needs stores.

This is a non-breaking change, since all data can still be accessed via the sphinx `BuildEnvironment`, (although this is now discouraged).

---

Note, this is intended as a step toward making the package type checked, since at present, although mypy is used, almost every module is set to be ignored (see `pyproject.toml`)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,6 +93,7 @@ warn_unused_ignores = false
 module = [
   'sphinx_needs.api.need',
   'sphinx_needs.builder',
+  'sphinx_needs.data',
   'sphinx_needs.diagrams_common',
   'sphinx_needs.directives.need',
   'sphinx_needs.directives.needbar',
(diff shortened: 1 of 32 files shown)


danwos (approved):
Wow, that is not a small change.
Thanks for it. Looks good for me. :+1:

PR #988, opened 2023-08-25
🧪 Add headless java to test builds

chrisjsewell:
This makes it easier to run the test suite,
since it inhibits window focus switching


danwos (approved):
Not much to say, looks good :) Approved!

PR #989, opened 2023-08-25
🧪 Remove all make files in test build folders

chrisjsewell:
These are unnecessary for the running of the tests,
and this makes them more minimal


danwos (approved):
299 Files changed, respect! :)

PR #995, opened 2023-08-26
🐛 Fix adding sections to hidden needs

chrisjsewell:
Previously hidden needs would not be added to the doctree. However, this meant that the `add_sections` transform did not add section related data fields to the need data. (this caused errors in the docs builds when filters required these fields)

The needs node is not created in the doctree, but marked with a `hidden` attribute, and these nodes are skipped in the HTML/LaTeX rendering.


danwos:
I'm not a big supporter of the visitor functions, as this is a builder-specific implementation. So other builders are not supported out of the box. Sphinx-Needs tries to avoid this; in my experience, this can become a maintenance hell.

For instance, I know that these kinds of builders are used with Sphinx-Needs as well:
json, Confluence, text

Is there any chance to get it done without visitor-specific functions?

chrisjsewell:
> I'm not a big supporter of the visitor functions, as this is a builder-specific implementation. 
> Is there any chance to get it done without visitor-specific functions?

Good point @danwos, I've moved it to a `doctree-resolved` event, does this look better

danwos (approved):
Looks good and approved.
The only thing is the additional ``findall()`` function call from docutils, as I have the feeling that its implementation is quite slow and walking the whole doctree is somehow inefficient.
But that's a general topic for a different repo ;)

chrisjsewell:
> as I have the feeling that its implementation is quite slow and walking the whole doctree is somehow inefficient

Now skipping doctree walks with no needs: https://github.com/useblocks/sphinx-needs/pull/995/commits/cc6d3c1287ff6dfe9d58dad296bb426be8975920

But for sure could be improved later

PR #998, opened 2023-08-27
🔧 Centralise access to sphinx-needs config

chrisjsewell:
similar to #987, accessing sphinx-needs configuration via the sphinx `Config` object is difficult to work with and ensure type safety, due to its dynamic nature.

This PR moves all configuration specification to the `NeedsSphinxConfig` class, which is a thin wrapper around the sphinx `Config`, to define and provide type safe access to the sphinx-need specific configuration.

This is non-breaking change, since all configuration can still be accessed via the sphinx `Config`, (although this is discouraged).


danwos (approved):
Ok, that's smart and a great example of the dataclasses benefits.
I like it :+1:

PR #1001, opened 2023-08-28
🔧 Make `NEEDS_CONFIG` type safe

chrisjsewell:
This commit makes `NEEDS_CONFIG` specific to the actual data that it holds, rather than just a generic data store.

---

Note, I have a feeling that this should be replaced entirely, since I'm not a big fan of mutable global variables.
But, for now, this does not change in any way the logic of sphinx-needs


danwos (approved):
You are too fast for my review skills ;)

Looks good to me :+1:

PR #1002, opened 2023-08-28
🔀🔧 MERGE: Add strict typing for most `sphinx_needs` modules

chrisjsewell:
This PR adds strict typing to most of the package, building on #987 and #998 to try to confine use of dynamic types to a small surface area.
This hopefully makes it easier for developers to navigate the code base, and lessens the potential for bugs arising from type mismatches.

---

There are still some open questions, particularly around `NeedsInfoType`, which

1. has some dynamic fields (such as user defined links/options) and,
2. has some fields which are only set after post-transforms (such `process_need_nodes`)


chrisjsewell:
Will be fixed by #1003

danwos (approved):

PR #1005, opened 2023-08-31
🧪 CI: Make documentation builds fail on warnings

### LINKED ISSUE #1004: Build official docs in own action workflow
I think the official docs get currently build inside a test case via pytest.
This makes it hard to see quickly any warnings.

We have a "Docs CI" pipeline, but this only contains the linkcheck. We should also add a doc-build there.

chrisjsewell:
closes #1004

Adds https://docs.readthedocs.io/en/stable/pull-requests.html

image

@danwos I believe this would be the best solution, however, it is currently failing due to https://github.com/jbms/sphinx-immaterial/issues/283 😒 (you can see this warning in all our docs builds)


chrisjsewell:
I added similar in the myst-parser docs @2bndy5  lol https://github.com/executablebooks/MyST-Parser/blob/350c6331da4738057556d966b0bfd9d98d570352/myst_parser/_docs.py#L275

2bndy5:
Yeah, I tried using that when adding a copy of MyST typography page to our docs, but it wasn't as bulletproof (I forget why and how). So I added the ability to highlight and parse the content as MD syntax if the doc source was MD.

2bndy5:
Ours isn't meant to be internal only. It does use theme-specific assets (CSS) like our `.. md-tab-set` (modeled after sphinx-design 😉 but with different JS for tab syncing and horizontally scroll-able overflow), so getting publicly exposed/packaged with the theme was a good fit. I kinda regret not documenting it now...

chrisjsewell:
> I kinda regret not documenting it now...

Never too late 😉

In the same file I have a decent myst syntax renderer, should probably make that public at some point

chrisjsewell:
Good to go @danwos

danwos (approved):
Thanks for the PR :+1: 

I guess we still need to configure github to allow passed RTD builds only, right?

chrisjsewell:
> I guess we still need to configure github to allow passed RTD builds only, right?

You can indeed add it as required in the branch protection rules

PR #1006, opened 2023-09-01
🐛 FIX: `NeedImport` logic

chrisjsewell:
@danwos unless I am missing something (?) I think this is broken, since `getattr` is used on a dictionary, instead of `dict.get`, so they will always return the default `None`.

Also improved efficiency, by using `json.load` instead of `json.loads`, and not recreating the `known_options` list within a loop


danwos (approved):
Great finding and thanks for the fix and the optimization. 
Looks all good to me.

chrisjsewell:
Cheers, just have to work out the test failure

PR #1007, opened 2023-09-02
Generic Sphinx-Needs JS test framework

iSOLveIT:
#### Setup Cypress Locally
- Install Node JS on your computer and ensure it can be accessed through the CMD.
- Install Cypress using the `npm` package manager by running `npm install cypress`. Read more from this link https://docs.cypress.io/guides/getting-started/installing-cypress#npm-install.
- Verify if Cypress is installed correctly and is executable by running: `npx cypress verify`. Read more from this link https://docs.cypress.io/guides/guides/command-line.
- If everything is successful then we can use Cypress.

#### Enable Cypress Test in PyTest
- First, create a test folder to store your Cypress JS test files (files should end with: `*.cy.js`). For each Cypress JS test file, you will need to write the Cypress JS test cases in the file. You can read more from the [Cypress Docs](https://docs.cypress.io/). You can also read the `tests/js_test/js-test-sn-collapse-button.cy.js` file as a reference.
- In your Python test files, you must mark every JS-related test case with the marker - `jstest` and you also need to pass the `spec_pattern` key-value pair as part of the `test_app` fixture parameter. For example, your test case could look like the below:
```python

# test_case.py

import pytest

@pytest.mark.parametrize(
    "test_app",
    [
        {
            "buildername": "html",
            "srcdir": "doc_test/variant_doc",
            "tags": ["tag_a"],
            "spec_pattern": "js_test/js-test-sn-collapse-button.cy.js",
        }
    ],
    indirect=True,
)
def test_case(test_app):
    ...
```
**NB:** The `spec_pattern` key-value pair is required to ensure Cypress locates your test files or folder. Visit this link for more info on how to set the [`specPattern`](https://docs.cypress.io/guides/guides/command-line#cypress-run-spec-lt-spec-gt).

- After you have set the `spec_pattern` key-value pair as part of the `test_app` fixture parameter, you can call the `app.test_js()` in your Python test case to run the JS test for the `spec_pattern` you provided. For example, you can use it like below:
```python
# tests_case.py

import pytest


@pytest.mark.jstest
@pytest.mark.parametrize(
    "test_app",
    [
        {
            "buildername": "html",
            "srcdir": "doc_test/variant_doc",
            "tags": ["tag_a"],
            "spec_pattern": "js_test/js-test-sn-collapse-button.cy.js"
        }
    ],
    indirect=True,
)
def test_collapse_button_in_docs(test_app):
    """Check if the Sphinx-Needs collapse button works in the provided documentation source."""
    app = test_app
    app.build()

    # Call `app.test_js()` to run the JS test for a particular specPattern
    js_test_result = app.test_js()

    # Check the return code and stdout
    assert js_test_result["returncode"] == 0
    assert "All specs passed!" in js_test_result["stdout"].decode("utf-8")
```
**NB:** `app.test_js()` will return a dictionary object containing the `returncode`, `stdout`, and `stderr`. Example:
```python
{
    "returncode": 0,
    "stdout": "Test passed string",
    "stderr": "Errors encountered,
}
```


chrisjsewell (changes requested):
Thanks @iSOLveIT, a few things to do initially:

1. Revert the changes to poetry.lock: no dependencies have changed, so this should not be changed.
2. The description of how to run the tests should be part of the contributing documentation
3. Please move the test to a seperate file and it should also probably me marked (see https://pytest.org/en/7.4.x/example/markers.html) so that contributors can "opt-in" to running this test, since it brings more complexity and overhead than is required for most changes

FYI, I feel there might be a better way to integrate these tests into pytest, and also the vendoring of JS in sphinx-needs could be improved. But that doesn't need to be addressed here

iSOLveIT:
Thanks for the comments @chrisjsewel. The PR is actually a draft for @danwos to see if we would like to adopt this mechanism.

danwos (changes requested):
Good work, and awesome documentation in the PR. 

I agree with all points from @chrisjsewell, especially the PR description should be part of the `contribute` page.

Also, the pytest-server handling should be done a little bit easier, see my comments.

But it all goes in the right direction and I'm happy to have this feature soon :+1:

tests/test_variants.py
```diff
@@ -56,3 +56,63 @@ def test_empty_variant_options_html(test_app):
...
+
+    app = test_app
+    app.build()
+
+    # Start the HTTP server using subprocess
+    server_process = subprocess.Popen(["python", "-m", "http.server", "-d", f"{app.outdir}", "65323"])
+
+    try:
+        # Run the Cypress test command
+        completed_process = subprocess.run(
```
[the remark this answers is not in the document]
danwos: Argh, also the selected test-file is not related to the js-test.
But maybe this is because the PR is just a proof of concept :)

tests/conftest.py
```diff
@@ -15,6 +21,82 @@ def copy_srcdir_to_tmpdir(srcdir, tmp):
...
+            "returncode": 1,
+            "stdout": None,
+            "stderr": f"The spec_pattern you provided cannot be found. (spec_pattern: {self.spec_pattern})",
+        }
+
+    js_test_config = {
+        "specPattern": cypress_testpath,
+        "supportFile": get_abspath("js_test/cypress/support/e2e.js"),
+        "fixturesFolder": False,
+        "baseUrl": "http://localhost:65323",
```
danwos, line 45: Does a hardcoded port work with parallel test execution?

If we have 20 js tests, they all would initiate their own cypress-server and block port 65323.
Can we make this somehow a little bit more safe?
E.g. use a range of 5 ports, and select the next one, if the current one is blocked by an already running cypress server?

Or do I see here a problem, where no problem is?

tests/conftest.py
```diff
@@ -43,6 +125,10 @@ def test_app(make_app, request):
         docutilsconf=builder_params.get("docutilsconf"),
         parallel=builder_params.get("parallel", 0),
     )
+    # Add the spec_pattern as an attribute to the Sphinx app object
+    app.spec_pattern = builder_params.get("spec_pattern", "")
+    # Add the test_js() function as an attribute to the Sphinx app object
+    app.test_js = test_js.__get__(app, Sphinx)
```
danwos, line 131: Ok, the Python gods have left me.
You need to explain this dark voodoo in a short call. Looks interesting :)

tests/js_test/js-test-sn-collapse-button.cy.js
```diff
@@ -0,0 +1,86 @@
+describe('Test Sphinx Needs Collapse', () => {
```
danwos, line 1: is `js-test-...` as prefix for the js-test file needed?
The folder is already called like this, so in my understanding we could shorten the file name a little bit. Just to `sn-collapse-button.cy.js`?

tests/test_js_code.py
```diff
@@ -0,0 +1,52 @@
+import pytest
```
danwos, line 1: I think we should not have a generic `test_jc_code.py` file, as it does not tell you what gets tested.
Let's keep the use-case-specific naming, as it doesn't matter if the test uses cypress or pytest only to test a feature.

danwos (changes requested):
Thanks for the changes. It getting better and better :)

Form the functional point of view, I only have some concerns with the parallel execution of tests and the hard-coded port number for cypress. Maybe there we can do some smarter port selection, if a port is already blocked.

The rest is only minor stuff.

tests/conftest.py
iSOLveIT: Sure.

iSOLveIT:
Hi @everyone,
I will be closing this PR due to some commit and rebase issues. 
You can find a resolved version of this PR in PR #1051.

PR #1008, opened 2023-09-03
🐛 Fix creation of need title nodes

chrisjsewell:
Previously the logic was mistakenly looping over every character of the string, and so creating a `Text` node per character, rather than a single `Text` node for the full title string.


danwos (approved):
Ohh, good finding and thanks for the fix :+1: 

Maybe this brings us also a little performance boost, as the Text()-initialization is used quite often (especially with the bug).

PR #1015, opened 2023-09-06
👌 Improve `process_constraints`

chrisjsewell:
- Add deprecation warning for `needs_constraints_failed_color` config option, which is no longer used
- Improve exception messages for missing keys in config options
- Add source location to failing constraints warnings
- Improve docstrings and type annotations for data fields/config options


danwos (approved):
LGTM :+1:

PR #1017, opened 2023-09-09
🔧 Remove `unwrap` function

chrisjsewell:
This run-time null check is effectively now replaced by the static type-checking.


danwos (approved):
Thanks for the PR and cleaning the code :+1:

arwedus:
This unwrap function was used by our permalink_ext_needs extension, do we have permalink feature in sphinx-needs itself now? Or do we need to cope with this some other way?

chrisjsewell:
>  I basically copied the unwrap function into our extension now. 

Yeh exactly just do this 👍, although I'd suggest maybe better to handle the root cause of why it's null, rather than raising assertion errors.
In the cases here, they can never be null, so there was no point checking

PR #1018, opened 2023-09-11
♻️ Refactor need data post-processing

chrisjsewell:
This PR moves the post-processing of needs data into a single event, which takes place before the write phase of Sphinx

In principle, this means that `needs` builders can bypass the entire write phase of the sphinx build.

However, there are currently two issues:

1. If `export_id` is set on a directive, this requires filter data to also be output in the `needs.json` data, and this is only computed in the write stage
2. contraints are currently not processed for "newly created" needs data (just tracking down where this happens, I guess for `needextract`)
3. For `needsuml` builder, the data is only added in `process_needuml`


chrisjsewell:
Closing as superceded by #1054

PR #1019, opened 2023-09-11
🧪 Add snapshot testing for needs data

chrisjsewell:
Previously, the validity of needs data was tested by asserting for only specific keys and values.
This PR adds snapshot testing, with https://github.com/tophat/syrupy, to more fully capture the generated data.
This should help to better identify any regressions caused by future changes (such as in #1018).


danwos (approved):
That's a great idea, using snapshot for this.

Is the snapshot-file taken automatically based on the test-file name?
Can the exact name of the snapshot be configured in the pytest?
I see use cases, where the snapshot may differ for specific tests in the same test-file.
Maybe this can also be documented in `contributing` in 1-2 lines.

However, approved :+1:

chrisjsewell:
> Is the snapshot-file taken automatically based on the test-file name?

Yes

> Can the exact name of the snapshot be configured in the pytest?

I don't believe so

> I see use cases, where the snapshot may differ for specific tests in the same test-file.

I think maybe you are misunderstanding?
A snapshot file can hold multiple snapshots, for a single test file; for each test and even multiple per test (https://github.com/tophat/syrupy#name)

PR #1029, opened 2023-09-19
Adds cache for meta._parse() function

danwos:


chrisjsewell (changes requested):
Heya, I would suggest using https://docs.python.org/3/library/functools.html#functools.lru_cache for this. Global variables are generally not a great coding practice 😅

danwos:
I know, should be also just a test and the change makes no sense. I close this PR

PR #1030, opened 2023-09-19
👌 Optimize needextend filter_needs usage

danwos:
Instead of using `filter_needs()` for a single need (`id == ""XY`) , the need is taken directly from `all_needs`, so that we do not need to perform any filtering.


chrisjsewell (approved):
LGTM

PR #1031, opened 2023-09-19
👌 From string fix

danwos:


sphinx_needs/external_needs.py
```diff
@@ -17,6 +19,16 @@
 log = get_logger(__name__)
 
 
+@lru_cache(maxsize=20)
+def get_target_template(target_url: str) -> Any:
```
chrisjsewell, line 23: Please try to avoid `Any` 😅 this returns a `jinja2.Template`

chrisjsewell (changes requested):

sphinx_needs/external_needs.py
danwos: Thanks for the hint. Was confused as `from_string()` is defined like this:
```
def from_string(
        self,
        source: t.Union[str, nodes.Template],
        globals: t.Optional[t.MutableMapping[str, t.Any]] = None,
        template_class: t.Optional[t.Type["Template"]] = None,
    ) -> "Template":
```

Where do I define things like `"Template"` for mypy?

chrisjsewell: > Where do I define things like "Template" for mypy?

I'm not sure what you mean by this?
You just import it, or if you really don't want to import it at runtime (because of import time or a circular dependency), then you can import it under an `if typing.TYPE_CHECKING:` clause, and refer to it as a string (as they do), or use deferred evaluation with `from __future__ import annotations` and then you don't have to use a string

danwos: Thanks, was not aware of the `if typing.TYPE_CHECKING` solution.

chrisjsewell (approved):
All good cheers

PR #1033, opened 2023-09-20
 👌 Removing deepcopy

danwos:
Tasks to do:
* [x] Remove deepcopy calls for filtering functions
* [ ] ~~Find alternative to secure need-data from manipulations~~ (Done by #1018)

Looks like this little change brings up to 80% better performance in a build.
Huge project got down to 5min from 28min.
Also memory consumption is cut down to 50-60%.


sphinx_needs/directives/needpie.py
```diff
@@ -156,7 +155,8 @@ def process_needpie(app: Sphinx, doctree: nodes.document, fromdocname: str, foun
                 # execute filter_func code
                 # Provides only a copy of needs to avoid data manipulations.
                 context = {
-                    "needs": copy.deepcopy(need_list),
+                    # "needs": copy.deepcopy(need_list),
```
chrisjsewell, line 158: you can remove the commented out code, we have the git history for that 😄 
```suggestion
```

sphinx_needs/filter_common.py
```diff
@@ -167,7 +166,8 @@ def process_filters(
     else:
         # Provides only a copy of needs to avoid data manipulations.
         context = {
-            "needs": copy.deepcopy(all_needs_incl_parts),
+            # "needs": copy.deepcopy(all_needs_incl_parts),
```
chrisjsewell, line 169:
```suggestion
```

chrisjsewell (changes requested):

danwos:
Right, also forget about the old code comments

chrisjsewell (approved):
Cheers

PR #1036, opened 2023-09-25
✨ Add error messages for constraint failures

### LINKED ISSUE #1011: Feature: Constraint help for error cases
Constraints shall get a help value, which is set to `constraints_help` inside a need if constraints are not passed.
The help text of all not passed constraints shall be joined to a single string.

Jinja shall be supported inside the help-text to inject values from the current needed, which has not passed the constraint.
Rendering rst is not supported, as the meta-data fields themselves do not support it.

``constrains_help`` shall be also available inside layout definitions.

Example: 
```python
needs_constraints = {
    "critical": {
        "check_0": "'critical' in tags",
        "check_1": "'security_req' in links",
        "severity": "CRITICAL",
        "help": "need {{id}} does not fulfill CRITICAL constraint, because tags are {{tags}}"
    },
}
```

chrisjsewell:
closes #1011 

@danwos I felt the name `help` was not very descriptive of its purpose, and having the config key as `error_message` and the needs field `constraints_error` was more understandable. What do you think?

Also perhaps:

- [x] Add documentation
- [ ] A try/catch for errors in the jinja templating
- [x]  Including the message in the sphinx warning

---

I'd note that, in principle, I feel it would be more "correct" to lay out the configuration like this:

```
needs_constraints = {
   "name": {
       "checks": {
            "0": "'critical' in tags",
		},
		"severity": "CRITICAL",
        "error_message": "need {{id}} does not fulfill CRITICAL constraint, because tags are {{tags}}",
   }
}
```

i.e. having the checks under a namespace, rather than having "special" keys


sphinx_needs/need_constraints.py
```diff
@@ -45,6 +46,11 @@ def process_constraints(app: Sphinx, need: NeedsInfoType) -> None:
                 need["constraints_results"].setdefault(constraint, {})[name] = False
                 need["constraints_passed"] = False
 
+                if "error_message" in executable_constraints:
+                    need["constraints_error"] = jinja2.Template(str(executable_constraints["error_message"])).render(
```
danwos, line 50: Initiating a jinja template takes some time, can we do it only once and execute only the rendering per need?

danwos (commented):
Thanks for the PR. Have only one suggesting regarding performance optimization.

chrisjsewell:
> Initiating a jinja template takes some time, can we do it only once and execute only the rendering per need?

Thanks @danwos, this pushed me to do something I was already planning to do in #1018: process all constraints in a single pass, rather than per document.

Do you know any reason why it was per document in the first place?
In fact, from the snapshots, this seems to fix some issues where some constraints were not getting processed 😄

danwos:
No, can't say (anymore9 why it was implemented like this. 
@mawieland As the initial implementer of this feature, do you remember or see any constraints?

danwos (approved):

PR #1037, opened 2023-09-25
🐛 Improve logic for `process_needextend`

### LINKED ISSUE #976: needextend: new link to not existing needs crashes
If I set an extra-link via `needextend` like this:
```rst
.. needextend:: NEED_1
   :own_link_type: NOT_EXISTING_NEED_ID
```

`needextend.py` throws an error:
```python
  File "[...]/python3.8/site-packages/sphinx_needs/directives/need.py", line 393, in process_need_nodes
    process_needextend(app, doctree, fromdocname)
  File "[...]/python3.8/site-packages/sphinx_needs/directives/needextend.py", line 142, in process_needextend
    if found_need["id"] not in env.needs_all_needs[ref_need][f"{option_name}_back"]:
KeyError: 'NOT_EXISTING_NEED_ID'
```

The reason is that this code:
`found_need["id"] not in env.needs_all_needs[ref_need][f"{option_name}_back"]`
is using ` env.needs_all_needs[ref_need]` and `ref_need` is referencing a not existing need.

**Solution**: Check if the link target exists in advance.

chrisjsewell:
Do not add non-existing links, and warn if found

closes #976


chrisjsewell:
@danwos I'd note that this did not, and still does not cover sub-IDs (a.k.a parts)

danwos (approved):
Thanks for the PR.
No support for need-parts is okay.
Not sure if there is really a use case for it, so let's wait for it before we support it  :)

PR #1045, opened 2023-10-11
🔧 Compatibility with Sphinx 7.2

ubmarco:
Changed compatibility matrix to
```
PYTHON_VERSIONS = ["3.9", "3.10", "3.11"]
SPHINX_VERSIONS = ["7.2.0", "7.2.6"]
DOCUTILS_VERSIONS = ["0.19", "0.20", "0.20.1"]
```
Test cases are green, however there are still warnings, mainly due to usage of deprecated Sphinx APIs:
`==== 129 passed, 561 warnings in 108.77s (0:01:48) ====`

This PR removes compatibility with Python <=3.8 and Sphinx <=7.2.
Background:
- Sphinx 7.2.0 dropped Python 3.8
- Sphinx 6.0.0 dropped Python 3.6 and Python 3.7
- Supporting Sphinx >7,<7.2 is difficult due to the deprecation of
  ```
  sphinx.builders.html.StandaloneHTMLBuilder.css_files
  sphinx.builders.html.StandaloneHTMLBuilder.script_files
  ```
  Using only non-deprecated APIs in one version of Sphinx-Needs for multiple Sphinx versions would require code branches.
  Also the new css/js builder properties fix the issue of stale asset files in consecutive builds, as tested by [tests/test_styles/test_style_css_js_registration.py](https://github.com/useblocks/sphinx-needs/blob/mh-sphinx-7.2.5/tests/test_styles/test_style_css_js_registration.py#L7). The custom code to reset the state is not necessary anymore and is removed by this PR.
- Supporting Sphinx <7.2 is difficult due to the deprecation of `sphinx.testing.path` in favor of `pathlib.Path`. This is used all over the place. Supporting both variants does not seem feasible.


danwos:
This PR drops support for too many Sphinx versions.
We created issue #1049, which defines our support strategy and cares about the correct implementation.
So the work from this PR will be reused for Sphinx 7.x support.

Thanks for the work :+1:

PR #1046, opened 2023-10-13
Fix usage of reST syntax in prefix parameter of meta

sleiner:
The usage of the `prefix` parameter of the `meta` layout function (as per the docs) does not work. There are two problems with this:

1. The documentation for `meta` does not mention that backslashes need to be escaped and shows unescaped backslashes instead. The docs for `meta_all` do not make this mistake.
2. [Since version 0.16](https://docutils.sourceforge.io/HISTORY.html#release-0-16-2020-01-16), docutils inserts null-characters when parsing escaped backslashes, which in turn is not accepted by [`ast.parse`](https://docs.python.org/3/library/ast.html#ast.parse). The solution is described in the [release notes for docutils 0.16](https://docutils.sourceforge.io/HISTORY.html#release-0-16-2020-01-16): Using `node.astext()` instead of `str(node)`

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -37,3 +37,5 @@ Duodu Randy <duodurandy19@gmail.com>
 Christian Wappler <chri.wapp@gmail.com>
 
 Chris Sewell <chrisj_sewell@hotmail.com>
+
+Simon Leiner <simon@leiner.me>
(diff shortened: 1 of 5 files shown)


danwos:
Thanks for the PR and the updated docs :+1: 

If you like, you can also add yourself to the AUTHORS list.
I will merge this PR tomorrow, so hurry up ;) (Or create a second PR for the AUTHORS update later)

danwos (approved):

Issue #957, opened 2023-08-03
needs.json builder performance


danwos:
Some actions are happening also in the needs.json-builder, which are not needed for the json-output itself.

For instance in the function `process_need_nodes` the `print_need_nodes()`-func gets executed which prepares the docutil-table representation of a need.
https://github.com/useblocks/sphinx-needs/blob/01f0e5d4e8a877fa92d3074facb9051b555cc2cc/sphinx_needs/directives/need.py#L397

There may be other locations and actions, which calculate some unneeded stuff. Some ideas:
* Representation of `needtable`
* Representation of `needflow` and `needuml`
* Copying css/js data to `_static` (may be already deactivated)

We need a way to skip such actions if a builder is used, which is not transforming the doctree into a result-format.

A simple `if builder == "needs"` is not working, as it does not allow to integrated other builders in future.
For instance, a PR is getting prepared, which creates a `needs.json` per file or even per need-object.
So checking the builder-type may be a better approach, not sure....

danwos:
Also `smartquotes` handling should be deactivated by default, as our jsin-files only contain the unrendered rst-code.

PR #1053, opened 2023-10-28
👌 Remove some unnecessary keys from `needs.json`

chrisjsewell:
Remove additional keys from the `needs.json`, that are not intended to be exposed to the user:
- back links for user defined link types (standard back links are already ommited),
- `id_parent` and `id_complete`, which are only relevant to parts expansion.

(Note, this is relevant for #957)


danwos (approved):
LGMT :+1:

PR #1054, opened 2023-10-28
👌 Improve performance of needs builders

### LINKED ISSUE #957: needs.json builder performance

chrisjsewell:
closes #957

This PR refactors the `needs` and `needs_id` builders, so that they can completely skip the "write" phase of the Sphinx build.

The main aim of these builders is to output the final needs data, in a particular format.
The needs data is collected during the "read" phase of the Sphinx build, i.e. when all documents are read, then after all needs data has been collected, a number of "post-processing" functions are applied to the needs data.
This post-processing is independant of (a) the individual documents, (b) the output format, and so does not require the Sphinx "write" phase to run.

Here we combine all of the post-processing into a single `post_process_needs_data` function that can be called directly by these builders.
Note, ideally this would be called in a [Sphinx event](https://www.sphinx-doc.org/en/master/extdev/appapi.html#emitting-events) occuring between the "read" and "write" events, however, this does not currently exist (the closest is `env-check-consistency`, but this is not triggered on "write-only" rebuilds).

The only corner case for skipping the "write" phase, is when the `needs` builder is required to also export filters, due to the presence of `export_id` options on certain directives.
These filters are currently computed during the "write" phase, so here we do run that, but emit a warning, to let the user know of this degradation in performance.


danwos (approved):
Great work and could catch with the `export_id` corner case. :+1:

PR #1056, opened 2023-11-02
⬆️ Update to Sphinx 7

chrisjsewell:
The principle goal of this PR is to update code/tests to allow for use with sphinx 7 and upgrade to Python 3.8-3.11.
To do this a number of other changes are made:

- The development dependency configuratio, in `pyproject.toml`, is modified to use separate [extras](https://python-poetry.org/docs/pyproject/#extras) for tests, benchmarks and docs
  - This allows for the `docs/requirements.txt` to be removed
- The noxfile now directly calls pytest/sphinx, rather than call `make`. This removes a confusing indirection, and allows for passing arguments to sessions: https://nox.thea.codes/en/stable/config.html#passing-arguments-into-sessions
- The CI now also directly calls pytest, again removing confusing indirections
  - The tests are split into core and JS tests (JS were previously not actually run)
- The `typing_extensions` backport for python<=3.7 is removed
- The [`sphinxcontrib-jquery`](https://github.com/sphinx-contrib/jquery) dependency is added, which re-adds the jquery JavaScript removed in sphinx>=6
  - This fixes the JS tests for sphinx 7


danwos (approved):
Thanks for all the work getting it running for 7.x :+1:

PR #1066, opened 2023-11-07
♻️ Default to warning for missing needextend ID

### LINKED ISSUE #1025: Sphinx-needs throws an exception when extending unknown need
When extending a need that does not exist (maybe because it was not imported by another sphinx plug-in like our EA needs importer), sphinx crashes due to an uncaught exception of sphinx-needs.

Console output:

```
...
Copying static files for sphinx-needs custom style support...[100%] /home/.../workspace/.sphinx-venv/lib/python3.8/site-packages/sphinx_needs/css/modern/modern.css

looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output in parallel...
Needs doc lookup table json successfully created
Needs successfully exported

Extension error (sphinx_needs.directives.need):
Handler <function process_need_nodes at 0x7fdf5ceac700> for event 'doctree-resolved' threw an exception (exception: 'SYSFCT_GenerateDynObjectsRepresentation')
```

Example document content causing the error:

```rst
.. req:: 
  :id: REQ_SYS_MJS_EnvRep_OBJ_Types

  example requirement

.. req:: 
  :id: REQ_SYS_MJS_EnvRep_OBJ_001

  example requirement
  
.. req:: 
  :id: REQ_SYS_MJS_EnvRep_OBJ_002

  example requirement

.. req:: 
  :id: REQ_SYS_MJS_EnvRep_OBJ_003

  example requirement
  
.. req:: 
  :id: REQ_SYS_MJS_EnvRep_OBJ_004

  example requirement
  
.. req:: 
  :id: REQ_SYS_MJS_EnvRep_OBJ_005

  example requirement
  
.. req:: 
  :id: REQ_SYS_MJS_EnvRep_OBJ_006

  example requirement

.. needextend:: SYSFCT_GenerateDynObjectsRepresentation
  :+implements: REQ_SYS_MJS_EnvRep_OBJ_Types,
                REQ_SYS_MJS_EnvRep_OBJ_001,
                REQ_SYS_MJS_EnvRep_OBJ_002,
                REQ_SYS_MJS_EnvRep_OBJ_003,
                REQ_SYS_MJS_EnvRep_OBJ_004,
                REQ_SYS_MJS_EnvRep_OBJ_005,
                REQ_SYS_MJS_EnvRep_OBJ_006
```

NOTE: "implements" is a custom relationship here, that represents a link from "req" to whatever "SYSFCT_GenerateDynObjectsRepresentation" ois.

This behavior is [as specified](https://sphinx-needs.readthedocs.io/en/rtd_jaav/directives/needextend.html#strict), because we have set the following configuration option:

```
needs_needextend_strict = True
```

However, what we actually need is that sphinx-needs emits a **Warning** if needextend is referencing a non-existing need. We can then filter this warning via our sphinx-warnings-filter extension, or configure if we treat warnings as error via sphinx' own `-W` option.

A warning message is also more readable than an exception.

chrisjsewell:
Rather than defaulting to "strict" mode for unknown IDs in `needextend` directives, whereby the entire build is excepted,
the default is now changed to "non-strict" mode, and this emits a warning rather than an info log.

I would suggest that actually this deprecates the use of the strict config/directive option, since the user can already halt builds on warnings using the standard `sphinx-build -W` flag.

Closes #1025

@arwedus this would mean you could simply remove `needs_needextend_strict = True` and rely on the new default warning behaviour. Make sense?


danwos:
I'm afraid this change destroys 1-2 use cases.

There are situations, where `-W` is already used, but not found ID in `needextend` is okay, because the Need to extend is part of another Doc project or get imported via `needimport`, and this kind of separated data is not available during a local build or shall not be used, as it is too big and slows down the build too much.

So during a local build, not found `needetend`-IDs is fine. But during a CI build, which combines all sources, a missing ID must throw an error. And `-W` is used in all cases to identify problems quite quickly.

chrisjsewell:
> And `-W` is used in all cases to identify problems quite quickly.

I would note, with the warning added here, it has the `needs.extend` type/subtype, allowing you to "fine-grain" the use of `-W` with https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings

I'll have a think about your use case though

arwedus:
@danwos: We cover that use case over our sphinx-warnings-filter extension instead of yet another config option for sphinx-needs. A [needs.extend] warning that can be configured in the sphinx warnings_filter list is imho indeed the cleanest solution.

danwos (approved):

PR #1069, opened 2023-11-13
🧪 Test README builds in pre-commit

chrisjsewell:
The v2.0.0 failed to publish 🙈, see https://github.com/useblocks/sphinx-needs/actions/runs/6851896295/job/18629213955

I believe this should be the fix.
Not sure if there is any better way to pre-check the deployment?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,3 +42,15 @@ repos:
     rev: 1.7.0
     hooks:
       - id: poetry-check
+
+  - repo: local
+    hooks:
+      - id: check-readme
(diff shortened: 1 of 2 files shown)


danwos (approved):
I love that you always create a CI update to test stuff automatically, instead of fixing only the problem :)

The solution is totally okay for me, as I hope that changes and problems are not happening too often in the README file.

chrisjsewell:
> I love that you always create a CI update to test stuff automatically, instead of fixing only the problem :)

Aha, have to protect against my human error 😅

PR #1070, opened 2023-11-13
🔧 Fix Docker deployment

chrisjsewell:
Fix the Docker build, which failed on the last deployment

Also 

- Add the new `plotting` extra, to install `matplotlib`
- Makes the docker build workflow run (without deployment) in PRs that modify the Docker files,
- fixes deprecation warning in workflow: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -12,8 +12,12 @@ on:
   push:
     tags:
       - '*.*.*'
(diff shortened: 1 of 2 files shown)


danwos (approved):
Looks good, thanks :+1:

PR #1080, opened 2023-12-01
🔧 Replace black/isort/pyupgrade/flake8 with ruff

chrisjsewell:


ubmarco:
Thanks for the efforts, that simplifies our life

danwos (approved):

PR #1081, opened 2023-12-05
Update permalink.html with a frienly text

csIcaK:
### Changed
- added text to display while loading the content.

### Reason
Currently on bigger projects it takes considerable amount of time to load the location of the selected ID. During that time currently there is a blank page shown to the user.

This change will display a static text: `Searching, Please Wait...`


danwos (approved):
Small change, a big improvement in the usability. Thanks for the PR :+1:

PR #1100, opened 2024-02-07
🔧 Remove unnecesary mypy ignore

chrisjsewell:
This disable does not appear to be needed


chrisjsewell:
ok it seems there is different behaviour for mypy on different python versions, so closing this for now

PR #1102, opened 2024-02-07
👌 Make needtable titles more permissive

### LINKED ISSUE #1089: Custom column titles in needtable does not work
Custom column titles in needtable does not work if the `needs_extra_options` has special characters like "-".

The following, for example, prints an empty "Description" column.

```rst
.. needtable::
   :filter: id.startswith("REQ_ID")
   :columns: id;title;status-desc as "Description";status
```

The following works:

```rst
.. needtable::
   :filter: id.startswith("REQ_ID")
   :columns: id;title;status-desc;status
```

chrisjsewell:
For `key as "title"` columns, the regex has been changed, to allow for the key to contain anything except whitespace characters, and the title to contain anything except the `"` character

closes #1089

### DIFF
--- a/sphinx_needs/defaults.py
+++ b/sphinx_needs/defaults.py
@@ -199,7 +199,7 @@ NEEDFLOW_CONFIG_DEFAULTS = {
     """,
 }
 
-TITLE_REGEX = r'([\w]+) as "([\w ]+)"'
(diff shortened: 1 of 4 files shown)


danwos (approved):
Looks good, thanks for the fix :+1:

PR #1104, opened 2024-02-08
🐛👌 Centralise need missing link reporting

### LINKED ISSUE #1038: Add sphinx warning for dead link
Currently, when links are checked in `check_links`, if a non-existent link ID is found then `has_forbidden_dead_links` is set on the need, but no warning is issued.

I think it would make sense to also emit a sphinx warning, @danwos ?

Also, I noticed this because if you change a link to a non-existent ID in our docs, then it will except the whole build, because this excepts due to an index error: https://github.com/useblocks/sphinx-needs/blob/4e604d0d187911f38b0c1a764c492b4bb521d0a1/sphinx_needs/filter_common.py#L175

so maybe this should also fail more gracefully

### LINKED ISSUE #1095: needs.json builder seems to not check links
The needs-builder seems to skip/not perform some checks, which are performed during the normal HTML build.

Reported from users are:
* links from need `A` to need `B`. If `B` does not exist, this seems to not be detected. 

We must be sure, that all validation gets executed independently from the used builder.

chrisjsewell:
Currently, warnings for need outgoing links that reference a non-existent ID, are only generated "indirectly" in the render phase; where  `NeedOutgoing` nodes are converted to nodes that are understood by sphinx builders (HTML, PDF, ...).

Since this render phase is no longer needed/run for simply creating the `needs.json`, using the `needs` builder, these warnings are not generated.

Additionally, any needs that are not explicitly rendered in the documentation, like externally imported needs, are also skipped.

The reporting of unknown links is now moved to the `check_links` function, meaning it is now run for all needs and all builders. The warnings have also been given a subtype `link_outgoing` or `external_link_outgoing`, e.g.

```
srcdir/index.rst:12: WARNING: Need 'SP_TOO_002' has unknown outgoing link 'NOT_WORKING_LINK' in field 'links' [needs.link_outgoing]
WARNING: http://my_company.com/docs/v1/index.html#TEST_01: Need 'EXT_TEST_01' has unknown outgoing link 'SPEC_1' in field 'links' [needs.external_link_outgoing]
```

This means they can be suppressed using the standard Sphinx config: 

```python
suppress_warnings = ["needs.link_outgoing", "needs.external_link_outgoing"]` 
```

This deprecates the need for the `needs_report_dead_links` configuration, which now emits a deprecation warning if set by the user.

@danwos, do you agree that `needs_report_dead_links` can be deprecated, and that warnings should be emitted for unknown ids in external need links?

closes #1038 
closes #1095

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -90,7 +90,7 @@ DEFAULT_DIAGRAM_TEMPLATE = (
 # Absolute path to the needs_report_template_file based on the conf.py directory
 # needs_report_template = "/needs_templates/report_template.need"   # Use custom report template
 
-needs_report_dead_links = False
+suppress_warnings = ["needs.link_outgoing"]
 
 needs_types = [
     # Architecture types
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -294,7 +294,13 @@ In this cases, you can provide a list of tuples.
 needs_report_dead_links
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Deactivate/activate log messages of outgoing dead links. If set to ``False``, then deactivate.
+.. deprecated:: 2.1.0
+
+    Instead add ``needs.link_outgoing`` to the `suppress_warnings <https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings>`__ list::
+
+        suppress_warnings = ["needs.link_outgoing"]
+
+Deactivate/activate log messages of disallowed outgoing dead links. If set to ``False``, then deactivate.
 
 Default value is ``True``.
 
--- a/sphinx_needs/config.py
+++ b/sphinx_needs/config.py
@@ -191,6 +191,7 @@ class NeedsSphinxConfig:
     Example: [{"name": "blocks, "incoming": "is blocked by", "copy_link": True, "color": "#ffcc00"}]
     """
     report_dead_links: bool = field(default=True, metadata={"rebuild": "html", "types": (bool,)})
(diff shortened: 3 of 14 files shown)


danwos (approved):
Thanks for the fix. Looks good :+1:  And I also agree `needs_report_dead_links` is deprecated with this change-

arwedus:
Hi @danwos  and @chrisjsewell , I've noticed in a local test that this does not raise warnings in case I've changed a need ID but forgot to update the links to it in other needs, when  I'm building with the needs builder (or inventories builder).

Is something else needed to raise link warnings with the needs/inventories (no-html-write-phase) builders?

chrisjsewell:
> Nevermind, it pulled in sphinx-needs 2.0.0 instead of master. Works now. Any chance of a new minor release soon?

Ah good 😄 

yes I hope soon, I'd just like to push to try and finalise some of my outstanding PRs first 😅 
probably another major, since some of the changes are "technically" a little breaking, but if you are already trying out `master`, then thats a good sign

arwedus:
> > Nevermind, it pulled in sphinx-needs 2.0.0 instead of master. Works now. Any chance of a new minor release soon?
> 
> Ah good 😄
> 
> yes I hope soon, I'd just like to push to try and finalise some of my outstanding PRs first 😅 probably another major, since some of the changes are "technically" a little breaking, but if you are already trying out `master`, then thats a good sign

@chrisjsewell and @danwos: Looking forward to it. Even if it's a new major, please don't drop support for Python 3.8 yet, if any way possible.

PR #1105, opened 2024-02-09
🧪 Add test for `needreport` directive

chrisjsewell:
Currently there is no test for this directive, this PR adds one.

This PR also fixes the directive:

- Make the options flags
- Change errors in the directive to emit warnings, rather than excepting the whole build
- Allow for `template` to be specified as a directive option
- Allow the the `dropdown` directive used in the default template, which requires an external sphinx extension, to be overriden using `needs_render_context = {"report_directive": "admonition"}` (I left the default as `dropdown`, so as not to introduce a breaking change)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -82,43 +82,11 @@ NOTE_TEMPLATE = """
    {% endif %}
 """
 
-EXTRA_CONTENT_TEMPLATE_COLLAPSE = """
-.. _{{id}}:
-
-{% if hide == false -%}
-.. role:: needs_tag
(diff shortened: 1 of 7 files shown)


chrisjsewell:
I've also just realised there is a bug in this directive:
The `usage` flag creates a count of user defined need types.
However, it does this during the processing of the directive, rather than in a post-processing step.
Therefore, the count will usually be incorrect.

danwos (approved):
THanks for the PR and the fixes :+1:

Issue #1103, opened 2024-02-08
Needs list shall take into account "only" directive 


David-Le-Nir:
When using the [only directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-only) some needs may require to be excluded from the needs list, i.e. from needs.json, needtable, needpie ...

For example:

```restructuredtext
.. only:: tag1

  .. req:: requirement for tag1
       :id: req-001

       This one shall only appear when building tag1

.. req:: requirement for all tags
     :id: req-002

       This one shall always appear
```

in this case when I don't build tag1, I don't want req-001 to appear in needs.json or in needtable 

Do you agree about the expected behaviour ?
Do you know if while processing the need level we are able to know that the element is under an "only" block ?

just for information, we found this interesting extension which can be inspiring: https://github.com/pfalcon/sphinx_selective_exclude

chrisjsewell:
Heya, oo tricky one

So, setting aside the logic of how to actually use this data for now, I think we may be able to capture the `only` tags on the needs data item, within the `analyse_need_locations` function.

Here you see we already capture what parent sections the need is under, so we may be able to do the same thing to capture what parent `only` the need is under:

https://github.com/useblocks/sphinx-needs/blob/4faedec27c24e314b7da9c298dc9aa67e97e9ff6/sphinx_needs/directives/need.py#L288

David-Le-Nir:
would removing the nodes as it is done for hidden needs be sufficient ? 
or it is only about the display ?

chrisjsewell:
> would removing the nodes as it is done for hidden needs be sufficient ? or it is only about the display ?

Only about display
 #1106 misunderstands hidden needs: they don't remove the need data, merely its represenataion in the output (e.g. in the HTML)

Remove data at this stage is not viable, because it would not work for multiple (cached) builds, using different tags / builders, i.e. subsequent builds would always have this data removed, even if you change the tags

The cached data (from the read phase) needs to be agnostic of builder / tags,
hence why I said you need to store the only expressions in the need data item.
Then, later in the process, during the builder/tag specific write phase, you decide what needs to omit

David-Le-Nir:
Thanks for the explanation.
I started a PR but, I don't know why, I do not manage to run tests.

Anyway, adding a field to the needs is ok. I propose to add boolean field named "excluded_by_only".
I'm not very happy with this name but could not find a better one. Please advise for another name if you have a better one.

[...]

PR #1106, opened 2024-02-09
exclude needs under only directive if required

David-Le-Nir:
fix for https://github.com/useblocks/sphinx-needs/issues/1103

### DIFF
--- a/sphinx_needs/directives/need.py
+++ b/sphinx_needs/directives/need.py
@@ -340,6 +340,13 @@ def analyse_need_locations(app: Sphinx, doctree: nodes.document) -> None:
         if need_node.get("hidden"):
             hidden_needs.append(need_node)
 
+        if need_node_excluded_by_only_directive(env, need_node):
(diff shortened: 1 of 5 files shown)


sphinx_needs/directives/need.py
```diff
@@ -356,6 +359,16 @@ def previous_sibling(node: nodes.Node) -> Optional[nodes.Node]:
     return node.parent[i - 1] if i > 0 else None  # type: ignore
 
 
+def need_node_excluded_by_only_directive(env: BuildEnvironment, node: nodes.Node) -> bool:
+    """Return True if the node is under an "only" directive that shall be excluded given to the current tags"""
+    if hasattr(node.parent, "tagname") and node.parent.tagname == "only":
```
danwos, line 364: I guess this would not support nested needs or needs in other directives like `container`, `list-table` and co.

danwos (changes requested):
Please see also my comment in the related issue: https://github.com/useblocks/sphinx-needs/issues/1103#issuecomment-1938228802

Issue #1099, opened 2024-02-05
needs_id_regex should not apply on needs_external_needs


David-Le-Nir:
When using _needs_id_regex_ and when referencing external documents via _needs_external_needs_ the regex is checked against all "remote" needs.

Would it be possible to only check the regex for the "local" document ?

For my users, each document should have its own regex for their own needs. Then when they refer to other documents, the regex check fails.

I agree with my users that the expected behaviour of the _needs_id_regex_ should to only apply locally because else every documents have to use the same regex.

I guess including _is_external_ value in the [regex check](https://github.com/useblocks/sphinx-needs/blob/73b961e29583bc0ac8892ef9c4e2bfb75f7f46bb/sphinx_needs/api/need.py#L193) would not be too hard. The questions are:

- how to make this behaviour optional, like a new option _need_id_regex_local_only_ 
- and what would be its default value

If you agree on this, I can make the PR if you want.

danwos:
Ohh, that's an interesting bug, and thanks for reporting it.

I agree that the external_needs should not be checked by the regex.
So I don't even need an option to configure this, I would just deactivate it completely for external_needs.

I see no use case, why the needs of project A should be checked by rules in project B.

Will be happy to merge an upcoming PR.

PR #1108, opened 2024-02-15
👌 Exclude external needs from `needs_id_regex` check

David-Le-Nir:
fix for [#1099](https://github.com/useblocks/sphinx-needs/issues/1099)


danwos (changes requested):
Thanks for the quick PR :+1: 

The change in `need.py` looks good, as well as the updated test case.

However, I think we need to clean this PR a little bit. :)

Version changes are introduced by a separate PR. Makes it easier to rewind single commits. So please remove this.

`.DS_Store` and `sphinxcontrib-needs.iml` are also not needed.
It's okay to add them to `.gitignore` if it simplifies your life.

And it would be also good to add a small line to `docs/changelog.rst` for a `2.1.0` release.

danwos:
Ohh and please ignore the failing pre-commit action.
Looks like a new mypy version finds much more now. I deal with this in a different PR and will merge this PR by hand.

David-Le-Nir:
> Version changes are introduced by a separate PR. Makes it easier to rewind single commits. So please remove this.
> 
> .DS_Store and sphinxcontrib-needs.iml are also not needed.
> It's okay to add them to .gitignore if it simplifies your life.
> 
> And it would be also good to add a small line to docs/changelog.rst for a 2.1.0 release.

yes, sorry, my CLI shortcuts are a bit too "shortcutting" ... I'll fix that
yes, for the changelog, I didn't know which version to set. Also should I update the pyproject.toml to set 2.1.0 for version ?

danwos:
No, we need to raise or better add the version just in the changelog.
All the other locations (pyproject.toml, needs.py, ...) will be done by a separate PR.

danwos:
I was asked by @chrisjsewell to hold this PR back for some time.
He may have some concerns but is sitting on a train right now.

docs/changelog.rst
```diff
@@ -9,6 +9,15 @@ License
 .. include:: ../LICENSE
 
 
+2.1.0
+-----
```
chrisjsewell, line 13:
```suggestion
Unreleased
----------
```

sphinx_needs/api/need.py
```diff
@@ -190,7 +190,7 @@ def run():
     else:
         need_id = id
 
-    if needs_config.id_regex and not re.match(needs_config.id_regex, need_id):
+    if needs_config.id_regex and not is_external and not re.match(needs_config.id_regex, need_id):
         raise NeedsInvalidException(f"Given ID '{need_id}' does not match configured regex '{needs_config.id_regex}'")
```
chrisjsewell, line 194:
```suggestion
    if (
        needs_config.id_regex
        and not is_external
        and not re.match(needs_config.id_regex, need_id)
    ):
        raise NeedsInvalidException(
            f"Given ID '{need_id}' does not match configured regex '{needs_config.id_regex}'"
        )
```

chrisjsewell (changes requested):
This is all good thanks, if you can apply the changes and fix the merge conflicts

sphinx_needs/api/need.py
chrisjsewell: or this will be done by running pre-commit

chrisjsewell (approved):
cheers

danwos (approved):

PR #1109, opened 2024-02-15
🔧 Update pre-commit hooks

chrisjsewell:


danwos (approved):
Thanks for fixing this. :+1: 
I'm pretty sure that wasn't so easy.

chrisjsewell:
> Thanks for fixing this. 👍 I'm pretty sure that wasn't so easy.

image

PR #1111, opened 2024-02-15
🔧 Use future annotations in all modules

chrisjsewell:
Just to keep things consistent across the code-base


danwos (approved):
Like the changes, as they look much more intuitive.
Thanks :pray:

PR #1112, opened 2024-02-15
👌 Capture `only` expressions for each need

### LINKED ISSUE #1103: Needs list shall take into account "only" directive

chrisjsewell:
@David-Le-Nir and @danwos, as I explained in https://github.com/useblocks/sphinx-needs/issues/1103#issuecomment-1936305902, I think this is a better solution for handling need defined  within `only` directives.

In this first "read" phase, we simply just note all the parent `only` expressions of the need, storing them on an (optional) `only_expressions` field of the need data item.

If desired, in a subsequent "build" post-processing phase, called from the cached data (once per build), you could then evaluate the `only_expressions` and remove needs as a necessary (or do whatever).
This logic would go here: https://github.com/useblocks/sphinx-needs/blob/84a5f72f2e72ab1471ab2d1bb5c570d6115ef199/sphinx_needs/directives/need.py#L385

this is more in-line with the logic of the `only` directive, where everything is cached and parts of the doctree are only removed during the build phase.

would supercede #1106 

closes #1103

### DIFF
--- a/sphinx_needs/directives/need.py
+++ b/sphinx_needs/directives/need.py
@@ -7,6 +7,7 @@ from typing import Any, Sequence
 from docutils import nodes
 from docutils.parsers.rst.states import RSTState, RSTStateMachine
 from docutils.statemachine import StringList
+from sphinx import addnodes
 from sphinx.addnodes import desc_name, desc_signature
 from sphinx.application import Sphinx
 from sphinx.environment import BuildEnvironment
@@ -309,7 +310,8 @@ def purge_needs(app: Sphinx, env: BuildEnvironment, docname: str) -> None:
 
 def analyse_need_locations(app: Sphinx, doctree: nodes.document) -> None:
     """Determine the location of each need in the doctree,
-    relative to its parent section(s) and need(s).
+    relative to its parent section(s) and need(s),
+    and also note any parent ``only`` expressions.
 
     This data is added to the need's data stored in the Sphinx environment,
     so that it can be used in tables and filters.
@@ -360,6 +362,16 @@ def analyse_need_locations(app: Sphinx, doctree: nodes.document) -> None:
         if parent_needs:
             need_info["parent_needs"] = parent_needs
             need_info["parent_need"] = parent_needs[0]
+
+        # find any parent only expressions, and note them on the need data item
+        expressions = []
+        parent: nodes.Element = need_node
+        while parent := getattr(parent, "parent", None):  # type: ignore
+            if isinstance(parent, addnodes.only):  # noqa: SIM102
+                if expr := parent.get("expr"):
+                    expressions.append(expr)
+        if expressions:
+            need_info["only_expressions"] = expressions[::-1]
 
         if need_node.get("hidden"):
             hidden_needs.append(need_node)
--- a/sphinx_needs/data.py
+++ b/sphinx_needs/data.py
@@ -200,6 +200,11 @@ class NeedsInfoType(NeedsBaseDataType):
     query: str
     url: str
 
+    only_expressions: list[str]
+    """List of parent only expressions, from outer to inner.
+    Note this key is only present if there are any only expressions.
+    """
+
     # Note there are also:
     # - dynamic default options that can be set by needs_extra_options config
     # - dynamic global options that can be set by needs_global_options config
--- a/tests/doc_test/doc_build_only/conf.py
+++ b/tests/doc_test/doc_build_only/conf.py
@@ -1,0 +1,4 @@
+version = "1"
+extensions = ["sphinx_needs"]
+
+needs_build_json = True
--- a/tests/doc_test/doc_build_only/index.rst
+++ b/tests/doc_test/doc_build_only/index.rst
@@ -1,0 +1,20 @@
+Title
+=====
+
+.. req:: Requirement 1
+   :id: REQ_1
+
+.. only:: html
+
+   .. req:: Requirement 2
+      :id: REQ_2
+
+   .. req:: Requirement 3
+      :id: REQ_3
+
+.. only:: not something
+
+   .. only:: other
+
+      .. req:: Requirement 4
+         :id: REQ_4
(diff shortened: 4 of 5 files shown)


chrisjsewell:
I see that the `only_expressions` is shown in the rendered admontion, I guess this should be hidden:

image

David-Le-Nir:
Hello,

I understand your solution is cleaner and offers more flexibility to choose what to do latter in the process. However when I tried that way, I was annoyed that I'd have to take this new field into account in each directive (e.g. needtable) to exclude the need automatically or that users would have to manually exclude the needs when using the directives.

Is the post-processing the solution for that ?

If so do you mean I have to implement my own logic in my own document ? Or do you suggest we add to this PR an exclusion of the needs in post_process_needs_data ?

sphinx_needs/directives/need.py
```diff
@@ -361,6 +363,16 @@ def analyse_need_locations(app: Sphinx, doctree: nodes.document) -> None:
             need_info["parent_needs"] = parent_needs
             need_info["parent_need"] = parent_needs[0]
 
+        # find any parent only expressions, and note them on the need data item
+        expressions = []
+        parent: nodes.Element = need_node
+        while parent := getattr(parent, "parent", None):  # type: ignore
+            if isinstance(parent, addnodes.only):  # noqa: SIM102
+                if expr := parent.get("expr"):
+                    expressions.append(expr)
```
danwos, line 372: Shouldn't we add a `break` here, so that we don't "look up" anymore after `only` was found?

chrisjsewell: no; we want to capture ALL parent `only` expressions, for example (this is in the test case):

```restructuredtext
.. only:: not something

   .. only:: other

      .. req:: Requirement 4
         :id: REQ_4
```

here the requirement should only be kept if BOTH `not something` and `other` evaluate to true,
so we capture and store `{"only_expressions": ["not something", "other"]}`

danwos:
I like the PR from a technical point of view, but I'm not a fan of it from a usability point of view.

I agree with @David-Le-Nir, that adding this to filter-strings and Co, to get a correct result, is not the best way to go.
95% of the users will likely not be aware of this.

Especially in bigger projects a user may not know that `only` is used somewhere.

In most of the projects, we decided to always use a *full* build to get a reliable result for a Sphinx-Needs project, as there can be so much external data and dynamic  functions, which do not work well with an *incremental* build.
Maybe using `only` is also such a case, so the user needs to be aware to create a *full*-build, if the final docs shall be used somehow official.

Not carrying correctly about `only_expressions` in filter strings could result every time in an invalid result.
Imagine having 2 Sphinx-Needs objects, one for **html** and one for **pdf**. If the filter string does not contain `only_expressions`, you will always get 2 items as a result.

Removing `only`-needs from the needs-dict may not be perfect, but at least we have one setup, which gives us always a valid result, by running simply a *full* build.

sphinx_needs/directives/need.py
danwos: good point and fully agree :+1:

chrisjsewell:
> we decided to always use a full build to get a reliable result for a Sphinx-Needs project

is this in the documentation?

If it doesn't, then really thats a quite a big limitation for sphinx-needs

> as there can be so much external data and dynamic functions, which do not work well with an incremental build.

I feel that sphinx-needs in general works quite well with incremental.
Certainly all the changes I have made, have been with that in mind, for example, dynamic functions are run after the needs-dict is cached, so these do work for incremental builds
Where it doesn't then we should work to fix that.

I really quite against introducing things like #1106, that inherently break incremental builds

> Not carrying correctly about only_expressions in filter strings could result every time in an invalid result.
Imagine having 2 Sphinx-Needs objects, one for html and one for pdf. If the filter string does not contain only_expressions, you will always get 2 items as a result.

This is not correct, you would remove the needs items, that have negatively-evaluated `only_expressions`, before any filter strings are are processed

danwos:
> I don't think this is correct; you would remove the needs items (that have negatively-evaluated only_expressions) before any filter strings are processed.

Ahh okay, then I have misunderstood the concept. But the current PR does not include this removement from the needs-list is not (yet?) part of the PR right?

> I feel that sphinx-needs in general works quite well with incremental.
Yes, it technically works well and incremental builds are running smoothly.

But if you change or remove a Need object in **File A**,  `needtables` and co do not get updated in **File B**. So the shown data is not valid.
So the representation of changed need objects in tables, flow charts, and extracts may not be correct, if all the files containing such need-representing directives get not rebuild as well. So in the end, a **full*-build is needed.

chrisjsewell:
> Or do you suggest we add to this PR an exclusion of the needs in post_process_needs_data ?

> Ahh okay, then I have misunderstood the concept. But the current PR does not include this removal from the needs-list, right?

yes and yes; I haven't implemented the removal yet, just wanted to intially show how the "first" phase would work

chrisjsewell:
> But if you change or remove a Need object in File A, needtables and co do not get updated in File B. 

if doesn't already, there is no reason why it cannot be; the needtable is rendered in the build (a.k.a post-processing) phase, so will use the updated need data
I think is just because FIle B is not being re-built, but I feel you could get this to trigger

chrisjsewell:
> So in the end, a full-build is needed.

but yeh obviously this is a more broad topic that I want to look into

PR #1113, opened 2024-02-16
👌 Improve and test github `needservice` directive

chrisjsewell:
By codecov, this directive was completely untested in the current test suite, so I added a test and also improved some other things I noticed:

- remove duplication of request mocking solutions in the test-suite ([responses](https://github.com/getsentry/responses) is more popular than [requests-mock](https://github.com/jamielennox/requests-mock))
- remove superfluous requests mocking in basic docs tests (that do not make any requests)
- add test for github `needservice` directives
- replace raising`NeedGithubServiceException` with emitting `needs.github` warnings
- pass the directive to the `service.request` method (in a non-breaking manner) so that it can be used to add the directive source location to the warnings
- Replace remaining tests that used `Sphinx` with `SphinxTestApp`, which can be "cleaned" after running (otherwise later test were emitting warnings caused by global variables previously mutated by these earlier tests)

### DIFF
--- a/tests/test_basic_doc.py
+++ b/tests/test_basic_doc.py
@@ -1,70 +1,22 @@
 import json
 import os.path
-import re
 import sys
 import tempfile
-import uuid
 from pathlib import Path
-from random import randrange
 
 import pytest
-import responses
 from sphinx import version_info
 from sphinx.application import Sphinx
 from sphinx.testing.util import SphinxTestApp
 from syrupy.filters import props
 
 from sphinx_needs.api.need import NeedsNoIdException
-from tests.data.service_github import (
-    GITHUB_ISSUE_SEARCH_ANSWER,
-    GITHUB_SEARCH_COMMIT_ANSWER,
-    GITHUB_SPECIFIC_COMMIT_ANSWER,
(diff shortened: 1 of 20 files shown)


tests/test_basic_doc.py
```diff
@@ -1,70 +1,22 @@
...
-
-#  OFFICIAL DOCUMENTATION BUILDS
-
-
-@responses.activate
 @pytest.mark.parametrize(
     "test_app", [{"buildername": "html", "srcdir": "doc_test/doc_basic"}], indirect=True
 )
 def test_build_html(test_app):
-    responses.add_callback(
```
danwos, line 58: Does removing this mean, our tests are firing now real requests to the GitHub API.
Or have I missed the locations, where this is still mocked?

chrisjsewell: `doc_test/doc_basic` has no `needservice` directives in, i.e. there is no requests to make 😄

danwos: argh sure, thought these removals had some functional reason.
But now it makes all sense...

danwos (approved):

Issue #1122, opened 2024-02-21
Internal handling of `extra_options` seems inconsistent


chrisjsewell:
This issue arises from when I was looking into the code considering #1082 and #1119
(@danwos as we discussed this morning)

The use of the config variable `needs_extra_options` and of the global variable `NEEDS_CONFIG.extra_options` seem inconsistent, in that:

1. Sometimes `needs_extra_options` is used, where it seems that `NEEDS_CONFIG.extra_options` should probably be used instead
2. `NEEDS_CONFIG.extra_options` is difficult to track, because at different stages it has new items added, which are obviously not taken into account in previous stages.

Below is a summary of their use in the code base:

---

Extra options are created by either:

1. A user adds them via the `needs_extra_options` configuration, as a `list[str]`
3. A user / extension uses `api.configuration.add_extra_option` to "programmatically" add a `name` to `NEEDS_CONFIG.extra_options` (raises `NeedsApiConfigWarning` if already added). For example `sphinx-test-reports` uses this in a `config-inited` event

They are used in the following places (in this order):

- In `load_config` (`config-inited` event) 
  - `needs_extra_options` are added to `NEEDS_CONFIG.extra_options` object (a warning is raised if one already exists)
  - `needs_duration_option` (default `duration`) + `needs_completion_option` (default `completion`) are also added if not present (as of #1127)
  - each `NEEDS_CONFIG.extra_options` is added as an option on:
    - `NeedDirective`
    - `NeedserviceDirective`
    - `NeedextendDirective` (and also as `+<name> -> unchanged` and `-<name> -> flag`)

- In `check_configuration` (`config-inited` event):
  - `needs_filter_data` keys are checked against `needs_extra_options` names; matching keys cause a `NeedsConfigException`
  - `needs_extra_options` names are checked against `INTERNALS` list; matching keys cause a `NeedsConfigException`
  - link type names (and their corresponding `<name>_back` key) are checked against `needs_extra_options` names; matching keys cause a `NeedsConfigException`
  - ensure all `needs_variant_options` keys match a `needs_extra_options` name, or a link type, or a default option; if not, a `NeedsConfigException` is raised

- In `prepare_env` (`env-before-read-docs` event):

    1. the `ServiceManager` is initialised and the `GithubService` and `OpenNeedsService` classes are registered via `ServiceManager.register`. In `ServiceManager.register`, for all `options` on the service class, the option is added to `NEEDS_CONFIG.extra_options` (if not already present)

[...]

Issue #1082, opened 2023-12-06
More efficient file format for needs database than needs.json


arwedus:
## Problem description

sphinx-needs stores all needs with all fields in needs.json. If you have a large meta-model with lots of custom attributes, every single needs adds several kb worth of data. All fields get written to the need dictionary, even those with empty default value.

For a large project, this file ends up being huge, even more so with lots of imported needs from other sources which are imported as needs directives via custom extension. We see 130 MB for a project that is still in "incubation phase". This leads to a noticeable wait in the "writing needs.json phase", and to an impressive disk space usage if you think about caching different needs.json versions.

Please re-write the needs export (and import) to use a more space-efficient storage format, without sacrificing much run-time in the read/write phase.

I have some ideas for open discussion, nothing I've yet investigated:

- only store non-default values for each need. This would already be a huge space saver, and the default values are easy to get back. All you would need is another entry with a meta-dictionary or maybe a schema that you save along with the project's needs.json. [edit CJS: done, see <https://sphinx-needs.readthedocs.io/en/latest/changelog.html#needs-json-improvements>]
- Use a different file format that supports optional fields, like google protobuf or maybe [msgspec](https://jcristharif.com/msgspec/)
- Internally change the way needs are kept in memory to organizing needs into containers per need-type. I.e. if you have the need types "req-sys", "req-sw", "test", and "spec": 4 containers. Change the configuration of needs options to define allowed extra needs options **per need type**. When writing needs.json, store needs grouped by need type. This way, you optimize for RAM and disk space memory. I understand that this may require a major rewrite of central parts of sphinx-needs, but I just wanted to put it in the discussion here for completeness, as this idea has also been discussed already since 2020.

arwedus:
@danwos: Well, you know our main use case (multi-project build and build caching), yet I don't find it to be represented in your list. For our use case, I'd definitely stay with one file and I'd prefer to have it either human readable or binary (like parquet or protobuf), but not first json, than zip file.

However, I feel that you have not adressed the obvious yet - it's simply unnecessary to store parts of the needs that do not add information. In any of those scenarios. And that's actually the optimization I'm requesting here. We will get already a factor of 5 - 10 size reduction in needs.json (depending on the complexity of the metamodel used in a project), while retaining compatibility.

danwos:
> ... it's simply unnecessary to store parts of the needs that do not add information

I can't entirely agree in all cases, but to be backward-compatible we can also not touch the `needs.json` export.

[...]

PR #1125, opened 2024-02-22
♻️ Store unset extra options as `Undefined` and remove from `needs.json`

chrisjsewell:
As discussed in more detail at https://github.com/useblocks/sphinx-needs/issues/1082#issuecomment-1948796243,
at present every stored need (dict) item in a project is required to have every key present.
In particular this is required so that filter strings (which are evaluated as Python code) can be evaluated for all needs; for example `"xxx == 1 "` would raise a `NameError` if `xxx` was not a key of the need.

As discussed also in https://github.com/useblocks/sphinx-needs/issues/1122#issue-2147120566,
this has meant that currently all fields added to `NEEDS_CONFIG.extra_options`, are added to every need item, with a default of `""`.
Storing as a `""` makes it difficult to tell if this was specifically set by the user or not, and consequently whether it can be removed from the `needs.json`

By contrast, jinja2 has the concept of https://jinja.palletsprojects.com/en/3.0.x/api/#undefined-types, by default assigned to any variable not provided as part of the context.

This PR changes the default to an `Undefined` instance and, when writing to a `needs.json`, removes and keys with this value.
As can be seen from the tests, for a standard need item, this removes 18 fields.

This solution should be "minmally breaking", in that mainly `Undefined` acts the same as `""`.
Although there may some rare exceptions, for example, the `calc_sum` filter function expected `float(value)` to raise a `ValueError`, but it now raises an `UndefinedError`.


chrisjsewell:
superceded by #1232

PR #1129, opened 2024-02-25
🔧 Centralise need parts creation and strongly type needs

chrisjsewell:
Need "parts" (a.k.a. sub-needs)  are expanded to full needs by copying the parent node then overding certain fields like the ID and content.
This PR replaces the duplicated logic by a single `iter_need_parts` function.

It removes the `is_need`, `is_part` keys from the `parts` dict, since this is handled in `iter_need_parts`,
plus it removes the `document` key, since this appears to be an erroneous duplication of the needs `docname` field (the part is always in the same document as the parent need)

Since the `NeedsPartsInfoType` is now identical to `NeedsInfoType` we remove it (these have to be identical, to allow filters to work for both)

Lastly the https://docs.python.org/3/library/typing.html#typing.Required is added to core fields in `NeedsInfoType`,
to allow for `add_need` to be properly typed checked, and so should ensure future code changes do not break the need schema.


danwos (approved):
The for the cleanup.
Looks all good :) :+1:

PR #1130, opened 2024-02-25
🔧 Remove usage of `hide_status` and `hide_tags`

chrisjsewell:
These needs fields were removed in v0.5.0


danwos (approved):
Thanks for finally getting this out from the code. :+1:

PR #1135, opened 2024-02-28
🐛 FIX: Image layout function

danwos:
Removes the relative path calculation from the "image" function in layout.py,as it was broken.
Now all set url-paths must be relative to the documentation root.


tests/conftest.py
```diff
@@ -283,6 +284,16 @@ def test_app(make_app, sphinx_test_tempdir, request):
         docutilsconf=builder_params.get("docutilsconf"),
         parallel=builder_params.get("parallel", 0),
     )
+    # Add the Sphinx warning as list to the app
+    # Somehow "app._warning" seems to be just a boolean, if the builder is "latex" or "singlehtml".
```
chrisjsewell, line 288: Oh thats weird 🤔

chrisjsewell (approved):
Great thanks 👌

PR #1151, opened 2024-03-23
make windows valid ci job

PhilipPartsch:


danwos:
I'm not 100% sure why this restriction is or was needed...
Maybe because we would create 2 reports for codecov, one for ubuntu and one for windows.

@chrisjsewell  Do you have an opinion here?

PhilipPartsch:
I tried to get a failed windows ci build, as some tests are failing there. I did not get it done. Any ideas?

chrisjsewell:
> I'm not 100% sure why this restriction is or was needed...

@danwos it was moreso, for when coverage CI did not use API tokens, and could be a little flaky to upload.
Although it doesn't really matter for our code base, since there's no code that only runs on windows, so it would not change the code coverage

> I tried to get a failed windows ci build, as some tests are failing there. I did not get it done. Any ideas?

@PhilipPartsch I'm not sure what you mean here by "make windows valid ci job";
the CI jobs still run the full pytest suite for windows (and nothing is failing),
your change just affects whether coverge is uploaded to codecov

When you say the tests are failing, I assume this is hen you run it locally; and so this seems to be something caused by your local setup.

Perhaps you could share, exactly what tests are failing, and also how you run the test suite locally?

PhilipPartsch:
> > I'm not 100% sure why this restriction is or was needed...
> 
> @danwos it was moreso, for when coverage CI did not use API tokens, and could be a little flaky to upload. Although it doesn't really matter for our code base, since there's no code that only runs on windows, so it would not change the code coverage
> 
> > I tried to get a failed windows ci build, as some tests are failing there. I did not get it done. Any ideas?
> 
> @PhilipPartsch I'm not sure what you mean here by "make windows valid ci job"; the CI jobs still run the full pytest suite for windows (and nothing is failing), your change just affects whether coverge is uploaded to codecov
> 
> When you say the tests are failing, I assume this is hen you run it locally; and so this seems to be something caused by your local setup.
> 
> Perhaps you could share, exactly what tests are failing, and also how you run the test suite locally?

As far as I can see, the job for windows in this PR is failing but marked green:
[build result](https://github.com/useblocks/sphinx-needs/actions/runs/8412086522/job/23032502282?pr=1150#step:6:1922)

chrisjsewell:
> As far as I can see, the job for windows in this PR is failing but marked green:

Good spot! see #1153

PR #1152, opened 2024-03-24
font fix check

danwos:


danwos:
Sets `font=False` for the documentation, to avoid usage of google fonts.
See https://jbms.github.io/sphinx-immaterial/customization.html#themeconf-font

PR #1153, opened 2024-03-25
🧪 Fix Windows pytest CI runs

chrisjsewell:
In #1067 I added coverage uploads.
@PhilipPartsch has uncovered an issue with this; it appears at least in Windows if `pytest` fails it still progresses to the next command `coverage xml` and then if this runs ok it doesn't mark the job step as failing 🤦

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,9 +44,9 @@ jobs:
           python -m pip install sphinx~=${{ matrix.sphinx-version }} -e .[test]
           python -m pip freeze
       - name: Run pytest
-        run: |
-          python -m pytest -v --ignore=tests/benchmarks -m "not jstest" --cov=sphinx_needs --cov-report=xml --cov-report=term-missing tests
-          coverage xml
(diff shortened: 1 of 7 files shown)


chrisjsewell:
Now I just need to fix the failures 😅

PhilipPartsch:
> Now I just need to fix the failures 😅

I have the fixes available at my privet PC at home. I can commit them in 12 hours.

chrisjsewell:
> I have the fixes available at my privet PC at home. I can commit them in 12 hours.

thanks!

PhilipPartsch:
> > I have the fixes available at my privet PC at home. I can commit them in 12 hours.
> 
> thanks!

My fixes are [here](https://github.com/useblocks/sphinx-needs/pull/1154)

chrisjsewell:
@PhilipPartsch for future knowledge, this is the problem: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

> `cmd`: There doesn't seem to be a way to fully opt into fail-fast behavior other than writing your script to check each error code and respond accordingly. Because we can't actually provide that behavior by default, you need to write this behavior into your script.

Windows being window 🤷‍♂️

danwos (approved):

PR #1168, opened 2024-04-25
Improve `Need` node creation and content parsing

chrisjsewell:
This PR improves the creation of need nodes and the `api.need.add_need()` function:

- It now accepts the `content` argument as a `docutils.StringList`, which can be passed directly from a directive.
- It now accepts a new `lineno_content` argument, which denotes the starting line number of the content, within the directive.
-  `lineno_content` is stored within the internal need data index (although for now it is not output in the `needs.json`)
- These arguments are used to improve the reporting of warnings, at the correct line number.

Note, it also deals with fixing some issues in the test suite arising from sphinx v7.3

----

original comment:

This is a WIP based on @PhilipPartsch original PR (#1150), where the main goal is to have sphinx warnings point to a sensible location in the source documentation (file-path and line-number)

The main difference being that here, where possible we simply pass through the `StringList` object (a "source-mapped" representation of the content) from the actual directive and pass that.
This works best for needs directives that are specified directly on the page, e.g.

```restructuredtext
.. req: title
   :id: my_id

   This is the content, which can have **nested syntax** that we want to parse
```
 
If the content is not directly on the page, then this is where I have found things becoming tricky.
This could either be because, the need contains jinja content, e.g.

```restructuredtext
.. req: title
   :id: my_id
   :jinja_content: true

   {% for i in range(11) %}
   This is the content.
   After it is "pre-rendered" by jinja, then the content might not align with the original source.
   {% endfor %}
```

or because it is an external need, loaded from a JSON file, e.g.:

```python
needs_external_needs = [
      {
        'base_url': 'http://mydocs/my_project',
        'json_url':  'http://mydocs/my_project/needs.json',
        'version': '1.0',
        'id_prefix': 'ext_',
        'css_class': 'external_link',
      },
 ]
```

For both of these, ideally we want to construct our own source-mapping object to pass to docutils/sphinx: https://github.com/live-clones/docutils/blob/582f8c946f776da7db652c06e0b49521f9dc3fdf/docutils/docutils/statemachine.py#L1323

For example, we might want to simply point every line to the start of the directive:

```python
sourcemap = StringList(content.splitlines(), items=[(source, lineno) for _ in content.splitlines()])
state.nested_parse(sourcemap, ...)
```

The appears to me, to be a minor technical annoyance here, in that docutils ignores this source-mapping, when it comes to issue warnings, for example here: https://github.com/live-clones/docutils/blob/582f8c946f776da7db652c06e0b49521f9dc3fdf/docutils/docutils/parsers/rst/states.py#L429

instead of using `StateMachine.get_source_and_line(lineno)`, which returns what we set in `sourcemap`, to get the correct place to report about, it actually uses `abs_line_number`, which just report the line offset in.

This is probably a bit technical for people reading this, but I write it here asI think I will eventually upstream the question as a bug to docutils.


PhilipPartsch:
Cool.
Hint: `needs_external_needs` are not an issue to been parsed, as external needs are been skipped before parsing. See current [api/need.py](https://github.com/useblocks/sphinx-needs/blob/90bae4a7664d8c7d899f034671374007a6b792f4/sphinx_needs/api/need.py#L459).

I believe we have to have a look about [needimport](https://github.com/useblocks/sphinx-needs/blob/90bae4a7664d8c7d899f034671374007a6b792f4/sphinx_needs/directives/needimport.py#L254C58-L254C62), especially as they are from a external source but not marked as `is_external`.

danwos (approved):

Issue #1172, opened 2024-05-07
Improve compatibility with common sphinx HTML themes


chrisjsewell:
On looking into improving our documentation, I was seeing if we could move to the furo theme https://github.com/pradyunsg/furo, since these days it is probably the most widely used sphinx theme.

I have found this quite difficult to get looking nice with the current HTML/CSS of sphinx-needs.
As exampled by the screenshot below, you can see that our current documentation relies on quite a "delicate" mixture of CSS across multiple sources; certainly non-trivial for users to get correct

Basically, we should aim to support out-the-box some of the common sphinx themes (or at least have documentation on the CSS required).

![image](https://github.com/useblocks/sphinx-needs/assets/2997570/e7ecd671-7951-4485-b3ab-3e819df4a5be)

Note, in sphinx-design I actually build the documentation for multiple themes (in synced repo branches). It may be overkill here, but it is a possibility:
![image](https://github.com/useblocks/sphinx-needs/assets/2997570/bbe1aed3-4ccb-47f5-8821-37933749c313)

PR #1178, opened 2024-05-08
👌 Streamline internal CSS

chrisjsewell:
As one might expect from CSS themes, the only thing that should really change are colors, and that was essentially the case for `sphinx_needs/css`

This PR moves nearly all the duplication of CSS within the theme specific files to `common.css`, 
and moves to simply using [color variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)

This makes it easier to customise these colors, have them dynamically change in [light/dark mode](https://css-tricks.com/dark-modes-with-css/) and, most importantly, integrate better with other sphinx themes (like e.g. https://pradyunsg.me/furo/customisation/#theme-options)

(this is a step towards #1172)


sphinx_needs/css/dark/data_tables.css
```diff
@@ -0,0 +1,101 @@
+/* DataTable layout*/
```
chrisjsewell, line 1: I would like to also remove this, but there wasn't obvious duplication here, so not as trivial

chrisjsewell:
a  few tests to fix, but yeh this is still ready for review

danwos (approved):
Looks much cleaner now. Thanks for this PR :+1:

PR #1181, opened 2024-05-15
👌 Improve cross-theme compatibility of need collapse button

chrisjsewell:
Currently, the need collapse toggle icons are added to the page like:

image

Wrapping the SVG icons in an `` is not great, because it does not allow it to be properly styled.
Indeed for our own docs, we have to use the horrible hack of `filter: brightness(0) invert(1)`, to get these icons to have the correct color for light/dark mode 😬 

This PR instead places the entire SVG onto the page, and adds some core CSS to set it at the correct relative size (and float right)

image

the SVG then matches the local text color, without any need for additional CSS:

image

Also, the initial visibility of the icons is now set correctly, before any JS is run,
so that you don't get a bad render experience, whereby both icons are initial visible (until the JS is triggered)

Note, it might be ideal to de-duplicate these SVGs on the page (see e.g. [this stackoverflow](https://stackoverflow.com/questions/64997454/how-to-eliminate-html-code-duplication-when-using-the-same-svg-icon-in-multiple)), but given these icons are small I don't think this is worth the added complexity for now

---

part of addressing #1172

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -66,7 +66,7 @@ h3 {
     --img-filter: brightness(0) invert(1);
 }
 
-img.needs-logo-big, img.sn_collapse_img {
+img.needs-logo-big {
     filter: var(--img-filter);
 }
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -1112,13 +1112,13 @@ python-dateutil = ">=2.7"
 
 [[package]]
 name = "mdit-py-plugins"
-version = "0.4.0"
+version = "0.4.1"
 description = "Collection of plugins for markdown-it-py"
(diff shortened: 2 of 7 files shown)


chrisjsewell:
Another note, https://sphinx-togglebutton.readthedocs.io also directly adds the SVG to the page, they also wrap it in a `button` rather than a `span`, which may be better for accessibility, but I won't do that in this PR
https://github.com/executablebooks/sphinx-togglebutton/blob/f468ca6c5fd4caf0085a97f84794d733794f3614/sphinx_togglebutton/_static/togglebutton.js#L28

danwos (approved):

PR #1186, opened 2024-05-31
✨ Allow creating a `needflow` from a `root_id`

chrisjsewell:
A classic use-case for visualising a graph, is to select a root node and visualise the sub-tree of all ancestor / decendant nodes.
Here we add the capability to the `needflow` directive, via new `root_id`, `root_direction` and `root_depth` options

### DIFF
--- a/docs/directives/needflow.rst
+++ b/docs/directives/needflow.rst
@@ -45,6 +45,71 @@ Options
    **needflow** supports the full filtering possibilities of **Sphinx-Needs**.
    Please see :ref:`filter` for more information.
 
+.. _needflow_root_id:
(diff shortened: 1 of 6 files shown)


danwos (changes requested):
Great idea and implementation, but I guess we need a little more docs for user :)

chrisjsewell:
> Great idea and implementation, but I guess we need a little more docs for user :)

done 👍 see https://sphinx-needs--1186.org.readthedocs.build/en/1186/directives/needflow.html#root-id

danwos (approved):
Thanks for the great docs

PR #1188, opened 2024-05-31
👌 Allow collapse/delete/jinja_content directive options to be flags

chrisjsewell:
Currently, one must use e.g.:

```restructuredtext
.. req:: title
   :collapse: true
```

but it feels more consistent/expected to allow for:

```restructuredtext
.. req:: title
   :collapse:
```


sphinx_needs/directives/need.py
```diff
@@ -86,32 +86,29 @@ def run(self) -> Sequence[nodes.Node]:
...
+            option_value = self.options.get(option_name)
+            if isinstance(option_value, str):
+                if option_value.upper() in ["", "TRUE", "YES"]:
+                    return True
+                if option_value.upper() in ["FALSE", "NO"]:
+                    return False
+                raise ValueError(
+                    f"{option_name!r} option must be true/false/yes/no, found: {option_value!r}"
+                )
+            return None
```
danwos, line 99: is `none` correct here?
Shouldn't it be `True` in cases where no string was provided, but the option was set?

chrisjsewell, line 99:
```suggestion
            return option_value
```

chrisjsewell: no, if the option is set then the value will be the empty string, which is now converted to `True` above,
I just changed it to return the `option_value` otherwise, to be the same as previous, although this should never not be `None` at this point

sphinx_needs/directives/need.py
```diff
@@ -86,32 +86,29 @@ def run(self) -> Sequence[nodes.Node]:
...
+            option_value = self.options.get(option_name)
+            if isinstance(option_value, str):
+                if option_value.upper() in ["", "TRUE", "YES"]:
+                    return True
+                if option_value.upper() in ["FALSE", "NO"]:
+                    return False
+                raise ValueError(
+                    f"{option_name!r} option must be true/false/yes/no, found: {option_value!r}"
+                )
+            return option_value
```
chrisjsewell, line 99:
```suggestion
            return option_value  # type: ignore[no-any-return]
```

docs/directives/need.rst
```diff
@@ -173,7 +173,7 @@ Below is an implementation of variants for need options:
       :id: VA_004
       :status: ['variants' in tags and not collapse]:enabled, disabled
       :tags: variants;support
-      :collapse: true
+      :collapse: True
```
chrisjsewell, line 176:
```suggestion
      :collapse: true
```

docs/directives/need.rst
```diff
@@ -183,7 +183,7 @@ Below is an implementation of variants for need options:
    :id: VA_004
    :status: ['variants' in tags and not collapse]:enabled, disabled
    :tags: variants;support
-   :collapse: true
+   :collapse: True
```
chrisjsewell, line 186:
```suggestion
   :collapse: true
```

sphinx_needs/directives/need.py
chrisjsewell: I removed the `true` from here: https://sphinx-needs--1188.org.readthedocs.build/en/1188/directives/need.html#collapse, and you can see it does work correctly

danwos (approved):

PR #1189, opened 2024-05-31
👌 Improve need link tooltips

chrisjsewell:
Currently, the tooltips in needs, for links to need IDs is a bit odd, it just says the current need ID with no context:
 
need-hover-now

This seems not useful, so here I have changed them to provide a tooltip of the title of the linked need:

need-hover-after


chrisjsewell:
Will need to fix the tests for this, but would be good to get feedback first

danwos:
This change is fine and makes total sense. :+1:

PR #1190, opened 2024-05-31
📚 Use new need-example directive for all examples

chrisjsewell:
This PR removes the duplication of examples (as literal and rendered), by adding a `need-example` directive.

This removes issues with inconsistencies, and also allows for consistent/centralised formatting of the examples.


chrisjsewell:
Don't know why RTD is not reporting the status, but it built: https://sphinx-needs--1190.org.readthedocs.build/en/1190/index.html

danwos (approved):
Love it. Thanks :+1:

PR #1191, opened 2024-06-02
🔧 Improve `match_variants` function

chrisjsewell:
Simplify the logic, add better docstring, and add unit tests, to make it clearer how it works.

Change to return `None` when no match, rather than just return the original value, to make the result clearer.

Allow the source location to be passed to the function, for reporting in warnings.

---

Note, this was actually done in order to enable an upcoming PR, to add a `border_color` option to `needflow`, e.g. to color borders based on their status:

```restructuredtext
.. needflow:: Engineering plan to develop a car
    :root_id: T_CAR
    :config: lefttoright,tutorial
    :show_link_names:
    :border_color: [status == 'open']:FF0000, [status == 'in progress']:0000FF, [status == 'done']:00FF00
```

### DIFF
--- a/sphinx_needs/utils.py
+++ b/sphinx_needs/utils.py
@@ -6,7 +6,6 @@ import operator
 import os
 import re
 from functools import lru_cache, reduce, wraps
-from re import Pattern
 from typing import TYPE_CHECKING, Any, Callable, TypeVar
 from urllib.parse import urlparse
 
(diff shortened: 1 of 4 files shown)


sphinx_needs/utils.py
```diff
@@ -542,110 +541,85 @@ def match_string_link(
...
-
-    # Handling multiple variant definitions
-    if isinstance(option_value, str):
-        multiple_variants: list[str] = variant_splitting.split(rf"""{option_value}""")
-        multiple_variants = [
+    options_list: list[str]
+    if isinstance(options, str):
+        options_list = re.split(
+            r"([\[\]]{1}[\w=:'. \-\"]+[\[\(\{]{1}[\w=,.': \-\"]*[\]\)\}]{1}[\[\]]{1}:[\w.\- ]+)|([\[\]]{1}[\w=:.'\-\[\] \"]+[\[\]]{1}:[\w.\- ]+)|([\w.: ]+[,;]{1})",
+            rf"""{options}""",
```
danwos, line 577: Should this regex be precompiled?

Otherwise, I guess it gets recompiled with every need-option?!

sphinx_needs/utils.py
```diff
@@ -542,110 +541,85 @@ def match_string_link(
...
-        return new_option_value
+    elif isinstance(options, (list, set, tuple)):
+        options_list = [str(opt) for opt in options]
     else:
-        return option_value
+        raise TypeError(
+            f"Option value must be a string or iterable of strings. {type(options)} found."
+        )
+
+    variant_regex = re.compile(r"^([\w'=,:.\-\"\[\] ]+):([\w'=:.\-\"\[\] ]+)$")
```
danwos, line 591: Does compiling these regexes on the module level only once make sense?

danwos (changes requested):
1-2 questions. But thanks for this enhancement.

sphinx_needs/utils.py
chrisjsewell: Nope, Python caches every compiled regexes: https://github.com/python/cpython/blob/8e6321efd72d12263398994e59c5216edcada3c0/Lib/re/__init__.py#L330

so in fact it was inefficient to be re-compiling on every call to this function, and this should improve perfomance!

danwos (approved):

PR #1198, opened 2024-06-18
🐛 Fix needextract

chrisjsewell:
- remove content_id
- rename content_node -> node_copy
- move needextract specific code to needextract.py
- use warnings instead of raising


chrisjsewell:
Superseded by #1249

PR #1201, opened 2024-06-26
📚 Make each example's sentence the need's own content in roles.rst

Peter-Baer:
Both examples on the roles page — the `:need:` one and the `need_outgoing` one — become the content of the need they describe, rather than paragraphs after it.

Until #1919 this could not build: a `[[field]]` inside a need's content was intercepted as the deprecated dynamic-function syntax (4.0.0, #1269) before the `:need:` role that carries it was resolved, so the first example's `[[value]] [[unit]] …` line failed the `-W` docs build with ten warnings once it sat inside the need. #1919 removed that scan for 9.0.0 (#1906), and the `:need:` role's own template syntax now renders inside a need — that build was the acceptance test for #1919, with this branch's diff applied.

The change itself is the original four lines; the branch was rebased after the documentation moved under `packages/sphinx-needs/docs/`, and brought up to date with `master` once #1919 landed.


chrisjsewell:
Rebased onto `master` with the full change restored. It cannot merge yet: the `[[…]]` inside the first example's `:need:` role is caught by the 4.0 deprecation of dynamic functions in need content and fails the `-W` docs build. #1906 removes that deprecation for 9.0.0; once it lands this branch builds as is and can go in.

chrisjsewell:
Blocked-on #1906 is now #1919; once that lands this can be brought up to date and merged.

chrisjsewell (approved):
Builds clean on master now that #1919 removed the content scan; the indented `:need:` example renders its resolved template inside the need.

PR #1202, opened 2024-07-01
FIX: Use `re.fullmatch` for `needs_id_regex` in need creation

chrisjsewell:
The regex supplied should be assumed to match the entire string, rather than be a partial match (as is the case currently using `re.match`).

This is then consistent with the usage of `needs_id_regex` in:

- https://github.com/useblocks/sphinx-needs/blob/1de1a7a047bfc7912ea0aa779ffb187def7a5254/sphinx_needs/directives/needextend.py#L111
- - https://github.com/useblocks/sphinx-needs/blob/1de1a7a047bfc7912ea0aa779ffb187def7a5254/sphinx_needs/directives/needextract.py#L103

### DIFF
--- a/sphinx_needs/api/need.py
+++ b/sphinx_needs/api/need.py
@@ -216,7 +216,7 @@ def add_need(
     if (
         needs_config.id_regex
         and not is_external
-        and not re.match(needs_config.id_regex, need_id)
+        and not re.fullmatch(needs_config.id_regex, need_id)
     ):
         raise NeedsInvalidException(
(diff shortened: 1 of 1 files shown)


chrisjsewell:
@danwos this now fails our docs build and some tests, because there it appears that regex is assumed to only be partial (only checking the start of the string).

So the question would be, do we instead change the code in `needextend` and `needextract` to use `re.match`, rather than `re.fullmatch`

danwos:
Good point, I guess the user can reach `fullmatch` by using in the configured regex an ending `$` right?
So yes, we should change it then also to `match` for needextend and co.

Issue #1035, opened 2023-09-21
Error when building multiple latex documents


marcel-22:
When I configure latex_documents[] in conf.py to output multiple documents, the following warning is generated and finally an exception occurs:

```
resolving references...
done
writing... WARNING: unknown node type: <NeedIncoming: <inline...>>
failed
Needs successfully exported

Exception occurred:
  File "C:\...\.venv\Lib\site-packages\docutils\nodes.py", line 2068, in unknown_departure
    raise NotImplementedError(
NotImplementedError: <class 'sphinx.writers.latex.LaTeXTranslator'> departing unknown node type: NeedIncoming
```

The latex_documents[] is setup to output two documents: 1) a document with the full output including all Sphinx pages, 2) a document with only a subset of the Sphinx pages.
```
latex_documents = [
    ('index', 'project.tex', 'documentation', author, 'manual', True),
    ('subset', 'project_subset.tex',  'subset', author, 'manual', True),
]

```
Both documents contain a page with the following sphinx-needs construct:

```
.. req:: Demo requirement
   :id: REQ_77bed70d
   :status: closed
   :collapse: False
   :tags: example

   Demo requirement.

.. req:: Link example
   :id: REQ_ae3d9643
   :status: closed
   :links: REQ_77bed70d
   :tags: example

   An example need with a link set to REQ_77bed70d.

```
 When I delete above statements from the .rst file, the file being completely empty, the build succeeds without errors.
Also when I remove the second document with the partial output from latex_documents[], the build succeeds.

```
Traceback (most recent call last):
  File "C:\...\.venv\Lib\site-packages\sphinx\cmd\build.py", line 285, in build_main
    app.build(args.force_all, args.filenames)
  File "C:\...\.venv\Lib\site-packages\sphinx\application.py", line 353, in build
    self.builder.build_update()
  File "C:\...\.venv\Lib\site-packages\sphinx\builders\__init__.py", line 308, in build_update
    self.build(['__all__'], to_build)
  File "C:\...\.venv\Lib\site-packages\sphinx\builders\__init__.py", line 378, in build
    self.write(docnames, list(updated_docnames), method)
  File "C:\...\.venv\Lib\site-packages\sphinx\builders\latex\__init__.py", line 306, in write
    docwriter.write(doctree, destination)
  File "C:\...\.venv\Lib\site-packages\docutils\writers\__init__.py", line 78, in write
    self.translate()
[...]

PR #1208, opened 2024-07-23
Bug fix: Generate several latex documents

fournier-p:
Hello,

This PR is an attempt to make it possible to generate several latex documents from the same doc project, fixing #1035.

I'm not sure of the possible side effects of the modifications I made, especially the one on col_row_maker, so any help is appreciated.

With the changes I try to detect when we want to reference a need from another document in need_outgoing.py and need_incoming.py and simply replace it by an emphasis to still show the need.

There may be a need to also change need_ref.py and other files.


MaxZignGroup:
Would it be possible to get this pull request merged? This correctly solves the issue that I am encountering at the following issue report: https://github.com/useblocks/sphinx-needs/issues/1035

PR #1215, opened 2024-08-08
chore(docs): fix typo

frank-fzi:
see https://www.scribbr.com/commonly-confused-words/a-vs-an/#u


danwos:
Thanks for the fix.
As our docs got refactored in the meantime (yeah I'm too slow in reviewing PRs) this fix is no longer needed for the `index.rst` file.

PR #1223, opened 2024-08-20
🐛 fix warnings for duplicate needs in parallel builds

chrisjsewell:
Change duplicate need feedback from raising exceptions to emitting Sphinx warnings, e.g.

```
path/to/page.rst:4: WARNING: A need with ID STORY_PAGE_1 already exists, title: 'duplicate'. [needs.duplicate_id]
```

and ensure warning is also emitted during parallel builds

Note, although there was already a parallel build test in the test suite, 
actually it was not running parallel because there were not enough documents in the test project (see https://github.com/sphinx-doc/sphinx/pull/12796),
so one more is added and now the warning is correct.

### DIFF
--- a/sphinx_needs/data.py
+++ b/sphinx_needs/data.py
@@ -678,7 +678,6 @@ def merge_data(
 
     _merge("needs_all_docs")
     _merge("need_all_needbar")
-    _merge("need_all_needbar")
     _merge("need_all_needextend")
     _merge("need_all_needextracts")
     _merge("need_all_needfilters")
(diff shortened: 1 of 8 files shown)


sphinx_needs/data.py
```diff
@@ -655,7 +676,8 @@ def _merge(name: str, is_complex_dict: bool = False) -> None:
                     f"not {type(other_objects)} and {type(objects)}"
                 )
 
-    _merge("needs_all_docs", is_complex_dict=True)
+    _merge("needs_all_docs")
+    _merge("need_all_needbar")
```
danwos, line 680: "needbar" is merged twice now

danwos (changes requested):
Looking good, just a small copy & paste issue.

danwos (approved):

PR #1132, opened 2024-02-26
👌 Auto-generate `needs` JSON schema

chrisjsewell:
(currently based on top of #1129)

description to come ...

PR #1230, opened 2024-08-21
✨ Add `needs_schema` to `needs.json`

chrisjsewell:
This PR builds and adds a schema for needs in the `needs.json` This will allow for people/tools to understand what each need field actually is, and is especially useful, for example, to allow tools to identify which fields are links.

Supersedes #1132


danwos (changes requested):
Looks good and looking forwar to use the new information in needs.json

However, a small hint in the builder section of the docs would be good :)

chrisjsewell:
> However, a small hint in the builder section of the docs would be good :)

done!

PR #1232, opened 2024-08-21
👌 Add `needs_json_remove_defaults` configuration

chrisjsewell:
Setting `needs_json_remove_defaults=True`,
will now create a `needs.json` with all need fields that have default values removed, and the defaults are specified in the `needs_schema` (recently added in #1230)

In the external/import needs loaders, these are then restored from the `needs_schema`.

In our internal documentation (of 224 needs), this reduces the generated `needs.json`size from 19692 lines / 760kB to 5894 lines / 248kB

Note: Currently, the default is `needs_json_remove_defaults=False`, so as not to be back-breaking yet,
but I would look to eventually change this to `True` before the next "major" release.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1725,8 +1725,8 @@ needs_json_remove_defaults
 
 .. versionadded:: 2.1.0
 
-Setting ``needs_json_remove_defaults = True`` will remove all need fields with default values, greatly reducing the size of the JSON file.
-The defaults are can be retrieved from the ``needs_schema`` now also output in the JSON file.
+Setting ``needs_json_remove_defaults = True`` will remove all need fields with default from ``needs.json``, greatly reducing its size.
+The defaults can be retrieved from the ``needs_schema`` now also output in the JSON file (see :ref:`this section <needs_builder_format>` for the format).
 
 .. _needs_build_json_per_id:
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1689,7 +1689,9 @@ Builds a ``needs.json`` file during other builds, like ``html``.
 
 This allows to have one single Sphinx-Build for two output formats, which may save some time.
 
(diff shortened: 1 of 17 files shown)


danwos:
Looks already good.
But one question, does the user need to specify the default now while registering an extra option? Or is there a common default for all extra options?

chrisjsewell:
> Or is there a common default for all extra options?

yep it's always the empty string

docs/configuration.rst
```diff
@@ -1709,6 +1709,10 @@ Example:
     Setting ``needs_reproducible_json = True`` will ensure the JSON output is reproducible,
     e.g. by removing timestamps from the output.
 
+.. versionadded:: 2.1.0
+
+    Setting ``needs_json_remove_defaults = True`` will remove all need fields with default values, greatly reducing the size of the JSON file.
+    The defaults are can be retrieved from the ``needs_schema`` now also output in the JSON file.
```
danwos, line 1715: The defaults ~~are~~ can be...

docs/configuration.rst
```diff
@@ -1709,6 +1709,10 @@ Example:
     Setting ``needs_reproducible_json = True`` will ensure the JSON output is reproducible,
     e.g. by removing timestamps from the output.
 
+.. versionadded:: 2.1.0
```
danwos, line 1712: Shouldn't `needs_json_remove_defaults` get its own headline?

Currently, we have for (nearly) all config options an extra section, so that it shows up in the toc.
This feature can also be used with ``make needs``. So, it's not only related to the containing section of `needs_build_json`

danwos (changes requested):
Looks good from the function point of view.

Only minor stuff in the docs.

docs/configuration.rst
chrisjsewell: done

docs/configuration.rst
chrisjsewell: done

danwos (approved):

PR #1235, opened 2024-08-23
✨ Allow for use of graphviz as the "engine" for `needflow`

chrisjsewell:
This PR allows for the use of [`graphviz`](https://graphviz.org) as the underlying engine for `needflow` diagrams, in addition to the default [`plantuml`](https://plantuml.com).

This is fully detailed in <https://sphinx-needs--1235.org.readthedocs.build/en/1235/directives/needflow.html>, with examples.

It introduces a new configuration option; `needs_flow_engine`, which can be set to either `plantuml` or `graphviz`, and defaults to `plantuml`,
and a new directive option; `engine`, which can be set to either `plantuml` or `graphviz`, and defaults to `plantuml`.
Thus the `graphivz` engine can be activated on a per-diagram basis, or globally.

------

The intention of this is to simplify and improve performance of graph builds:

- calls to plantuml requires installing it, then initialising an entire Java virtual machine to render the graph, whereas graphviz just requires a CLI call to an executable, making it a lot simpler, and likely faster than
  - note, potentially in the future, there is even a full-blown python FFI wrapping the Graphviz C library, which could make things even quicker, although they are still working on wheel distributions: https://github.com/pygraphviz/pygraphviz/issues/167
  - integration with sphinx for plantuml requires a third-party library (https://github.com/sphinx-contrib/plantuml), whereas graphviz integration is built-in (https://www.sphinx-doc.org/en/master/usage/extensions/graphviz.html)

Indeed, setting `needs_flow_engine = "graphviz"` for the documentation of this repo reduces the local build time from 81 seconds to 56 seconds.

------

The `graphviz` engine supports all the existing `needflow` options,
although one key complication is the translation of existing "style related" configurations, which are hard-coded to plantuml syntax:

- fields in `needs_extra_links`: `style`, `style_part`, `style_start`, `style_end`
- `needs_flow_configs`
- `needs_diagram_template`

The `needs_extra_links` fields are translated to graphviz syntax, where possible, and warnings emitted for those that could not be. Two issues that probably cannot be resolved (it is unclear exactly how plantuml achieves them):
1. Link styles with a direction, such as `-up->`
2. Applying correct shapes for needs that contain children/parts; in graphviz nodes cannot contain other nodes, and instead we must use "subgraph clusters" for such nodes, which do not allow a `shape` attribute

A separate `needs_graphviz_styles` configuration is introduced, which is similar to `needs_flow_configs`, except allowing graphviz attributes to be set (and used by the `needflow` directive's `config` option).

`needs_diagram_template` is not used; currently the graphviz engine hard-codes the format of node labels, but this could potentially be added in the future.

Additionally an `alt` option is added to the `needflow` directive, and supported by both engines.

------

Note, currently the rendering with the `graphviz` engine is restricted to HTML output only, this could fairly easily be extended to latex/man/text/texinfo, which `sphinx.ext.graphviz` already supports, the main reason it is not already, is that we override the visitor node currently, to make some improvements.

Then one extra thing I want to note here, that shouldn't block this PR but would be nice to resolve eventually,
is what HTML to output and how to make it allow for transparent backgrounds and clickable nodes (linking to their need description):

- if you use `<object data="path/to/file.svg" type="image/svg+xml">` it allows directly for clickable links, but I can't seem to get the background of the image to be transparent (its always white, which is a bit horrible for dark themes)
- if you use `` the clickable links do not work directly (you first need to click on the whole image, to go to a page of the SVG), but transparent backgrounds do work


PhilipPartsch:
@chrisjsewell Really cool implementation! I have worked on this even, so cool to see how fare the implementation already is. 
I even tried to implement a possibility to use graphviz graphs directly in a `need` or document like `needarch` or `needuml` for plantuml.
For this I used `needgraph`. So I want to suggest to hide the implementation of `needflow` with graphviz from the user and would make plantuml / graphviz as a configuration option in conf.py or the dedicated `needflow` directive. 
So we can use `needgraph` to be used like `needuml` for jinja templated need processing.

What do you think?

chrisjsewell:
> What do you think?

all done

PR #1241, opened 2024-08-26
🔧 Remove `content_node` field from need items

chrisjsewell:
`content_node` is different to other need fields,
in that it is a non-jsonable type, and is not expected to be used by users in need filters.
Its only use is for `needextract` directives.

Therefore, we move it to its own separate "storage solution". For now, they are still stored on the `BuildEnvironment`, but ultimately we may want to pickle them to disk (similar to doctrees), since they can add a significant memory size to the `BuildEnvironment` that can have effects on performance.


danwos (approved):
I agree and also find it a little dangerous, as the content_node must not have any connection to the related doctree element, otherwise the doctree gets not freed after pickling and at the end multiple doctress are kept in memory. We had such bug in the past.

But the realization of that is something for another PR.

PR #1249, opened 2024-08-27
🐛 Fix rendering of `needextract` needs

chrisjsewell:
Rendering of `needextract` nodes had two key issues:

1. The call to `BuildEnvironment.resolve_references` was already converting the need node to its final representation (since the calls `process_need_nodes`), and so the later call to `build_need` didn't do anything.
   This is why the requested layout was not being created.

2. A `needextract` can generate multiple need representation, but all of them were being given the same `id` of the original `NeedExtract` node.
   All nodes are now wrapped in a containing node, that is set with this `id`

Additionally, warnings are now emitted for unfound needs, rather than raise exceptions

Also, `build_need` is renamed to `build_need_repr`, to better describe its purpose, and its signature is slightly changed, to allow for the new logic.

For examples of now correct rendering, see: https://sphinx-needs--1249.org.readthedocs.build/en/1249/directives/needextract.html

### DIFF
--- a/docs/directives/needextract.rst
+++ b/docs/directives/needextract.rst
@@ -23,12 +23,17 @@ For instance, a supplier could get a copy of requirements but would not see all 
       :style: green_border
 
 
-.. note:: 
-   
-   **needextract** supports the full filtering possibilities of **Sphinx-Needs**.
-   Please read :ref:`filter` for more information.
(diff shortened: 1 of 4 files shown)


danwos (approved):
Thanks for this fix :+1:

PR #1259, opened 2024-08-30
🐛 Coerce dynamic function return to str (when replacing content)

chrisjsewell:
As found by @ubmarco, the return type of a dynamic function may not be a string, as so we need to make sure it is coerced to one, otherwise `str.replace` will except

I guess for your exception @ubmarco perhaps a field value was changed from an empty string to `None` (going from v2.1.0 to v3.0.0) 🤔

### DIFF
--- a/sphinx_needs/functions/functions.py
+++ b/sphinx_needs/functions/functions.py
@@ -155,7 +155,10 @@ def find_and_replace_node_content(
             if isinstance(func_return, list):
                 func_return = ", ".join([str(v) for v in func_return])
 
-            new_text = new_text.replace(f"[[{func_string_org}]]", str(func_return))
+            new_text = new_text.replace(
+                f"[[{func_string_org}]]",
+                "" if func_return is None else str(func_return),
+            )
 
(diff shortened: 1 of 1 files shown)


ubmarco:
This is the problematic code on my side
```
needs_global_options = {
    "style": [
        (
            "status_[[copy('status')]] needs_style_release_[[link_copy('release')]]",
            "type in ['story', 'release', 'specification', 'phase', 'problem']",
        ),
    ],
}
```
The call `"copy('status')"` returns None, so it's not a self-defined dynamic function.
In my case I need to make sure every affected need has a valid `status` field.
Coercing `None` to an empty string fixes the problem immediately, but might have hard-to-debug side effects.
E.g. in my case the style field would have the value `status_` set which is only visible when looking at the output.
A warning would make sense. I don't see a case in which having a dynamic function return None is not an exception.

ubmarco (changes requested):
Rephrasing my comment in form of a review: Please add a warning in case functions do not return a string.

chrisjsewell:
Superseded by #1262

PR #1262, opened 2024-09-01
🔧 Improve typing/docs of `DynamicFunction`

chrisjsewell:
- strictly type functions
- fix issue with wrong type to `str.replace`
- log warning, if incorrect function or return type, rather than excepting (with correct source location)
- add test for warnings
- add `DynamicFunction` protocol to API documentation

### DIFF
--- a/sphinx_needs/functions/functions.py
+++ b/sphinx_needs/functions/functions.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 import ast
 import re
-from typing import Any, Callable, List, Union
+from typing import Any, Protocol
 
 from docutils import nodes
 from sphinx.application import Sphinx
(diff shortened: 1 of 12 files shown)


sphinx_needs/functions/functions.py
```diff
@@ -93,18 +119,24 @@ def execute_func(app: Sphinx, need: NeedsInfoType, func_string: str) -> Any:
...
     if isinstance(func_return, list):
         for element in func_return:
-            if not isinstance(element, (str, int, float, unicode)):
-                raise SphinxError(
-                    f"Element of return list of function {func_name} is of type {type(func_return)}. "
-                    "Allowed are str, int, float"
+            if not isinstance(element, (str, int, float)):
+                log_warning(
+                    logger,
+                    f"Return value of function {func_name!r} is of type {type(func_return)}. Allowed are str, int, float, list",
```
ubmarco, line 135: Can you please include the list index to the message? The type is already a list, just one value has a wrong type.

ubmarco (changes requested):
Very nice improvement, just one remark

ubmarco (approved):
Thanks for the quick fix. Approved!

PR #1269, opened 2024-09-03
✨ Add `ndf` role, deprecate `need_func` & `[[...]]` in need content

chrisjsewell:
There are currently two ways to use dynamic functions within need directive's content:

1. ``[[copy("id")]]``; is problematic, because it does not adhere to the rst/myst syntax specification, and has already show to cause issues and be surprising to users (#1263)

2. `` :need_func:`[[copy("id")]]` ``; is better but overly verbose

In this PR, I introduce a new role, which I believe should supersede both: `` :ndf:`copy("id")` `` (open to other suggestions for the name)
Here we take the entire content to be the function, as so do not require the `[[]]`, reducing verbosity and processing

Note, one thing that the role approach cannot do, is to allow placeholders in URLs, e.g. `` `link <http://www.[[copy('id')]]>`_ ``
However, if this is truly required, then I believe we should introduce a new role specifically for that.

---

if this is accepted, I would then move to emit warnings if users already use `[[...]]`, explaining how to replace them,
then eventually remove `[[...]]` after some deprecation period

---

things to clarify:

- [x] do we agree this role should be added
- [x] do we agree on its name
- [x] do we agree this should deprecate `[[...]]` in content
  - [ ] do we need an additional role for URL links with placeholders
- [x] do we agree this should deprecate `need_func`

### DIFF
--- a/tests/doc_test/doc_dynamic_functions/index.rst
+++ b/tests/doc_test/doc_dynamic_functions/index.rst
@@ -10,6 +10,8 @@ DYNAMIC FUNCTIONS
 
     This is also id :need_func:`[[copy("id")]]`
 
+    This is the best id :ndf:`copy("id")`
+
 .. spec:: TEST_2
    :id: TEST_2
    :tags: my_tag; [[copy("tags", "SP_TOO_001")]]
@@ -37,4 +39,6 @@ DYNAMIC FUNCTIONS
 
         nested id also :need_func:`[[copy("id")]]`
 
-This should warn since it has no associated need: :need_func:`[[copy("id")]]`
+        nested id best :ndf:`copy("id")`
+
+These should warn since they have no associated need: :need_func:`[[copy("id")]]`, :ndf:`copy("id")`
--- a/docs/directives/needextend.rst
+++ b/docs/directives/needextend.rst
@@ -35,8 +35,8 @@ Also, you can add links or delete tags.
 
       This requirement got modified.
 
-      | Status was **open**, now it is **[[copy('status')]]**.
-      | Also author got changed from **Foo** to **[[copy('author')]]**.
+      | Status was **open**, now it is :ndf:`copy('status')`.
+      | Also author got changed from **Foo** to :ndf:`copy('author')`.
       | And a tag was added.
       | Finally all links got removed.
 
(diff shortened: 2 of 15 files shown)


ubmarco:
Hi @chrisjsewell, thanks for this proposal.
As you said and as the past showed, the `[[ ]]` construct is brittle and breaks RST.
I like the role approach and also the name. Here are my answers:

- [x] do we agree this role should be added
- [x] do we agree on its name
- [x] do we agree this should deprecate `[[...]]` in content
  - [ ] do we need an additional role for URL links with placeholders
- [x] do we agree this should deprecate `need_func`

For the URL links I'm hesitant to add it without reported use case.
For `ndf` - can we somehow make sure that the name does not interfere with other extensions providing a role of the same name? `need_func` was certainly more scoped to SN.

chrisjsewell:
Yeh cheers

> can we somehow make sure that the name does not interfere

Eurgh, I would be very surprised if someone else was specifically using `ndf` 😅
Currently it will just override any existing role.
Happy to think of an even more specific name, but the idea was to make it not too verbose, similar to `np`

tests/doc_test/doc_dynamic_functions/index.rst
```diff
@@ -37,4 +39,6 @@ DYNAMIC FUNCTIONS
 
         nested id also :need_func:`[[copy("id")]]`
```
ubmarco, line 40: if `need_func` is deprecated, please add a check for the warning in the test case

ubmarco (approved):
LGTM in general, docs and still deprecation missing, but can be done in an upcoming PR

danwos (changes requested):
I'm okay with the changes. Good work.

But as this is a not backward-compatible change, we should add the docs and a depreciation warning for need_fuc with this PR.

chrisjsewell:
> But as this is a not backward-compatible change

yep I was already about to add docs and deprecations 👍 , but just to note, this PR on its own is backwards compatible; it is just adding a new role

danwos (changes requested):
Looks great. Thanks for the doc updates.

danwos (approved):

ubmarco (approved):
Nice work and good addition to SN.

Issue #1264, opened 2024-09-02
Better abstraction for accessing needs data (to improve scaling)


chrisjsewell:
Currently, all needs are stored and accessed from a simple in-memory `dict[str, dict]` on the sphinx build environment, mapping the need ID to the need data item.

However, when considering large projects with ~100,000 needs or more this starts to become problematic for:

1. the memory consumption of sphinx; especially when considering multi-processing, if every process requires a copy of this dict
2. filtering of needs; having to run a Python `for` loop over all needs, then check on each data item, will be noticeably slow

Ideally we want an abstraction for accessing the needs, particularly in the write/analysis phase, such that:

1. the abstraction hides to the user where/how the actual data is stored
2. the abstraction offers additional methods for performant filtering on common queries (e.g. finding all needs with a certain status)

Some work was started on this in #1255, however, one big pain point is that currently the needs are "exposed" to users in a number of different ways (see below), principally:

1. Directly, the stored `dict[str, dict]` (after processing)
2. the stored `dict[str, dict]` converted to a `list[dict]`
3. the stored `dict[str, dict]` converted to a `list[dict]`, but also with all need parts "expanded" and appended to the list

so somehow, we need to account for these representations also, without back-breaking changes

This also relates to: #1238 and #1217, asking to expose needs in more places, and #1219 requesting better scaling filtering

----

## Needs access life-cycle

- reading/collection phase
   - retrieve cached (unprocessed) needs, or instantiate
   - add any external needs
   - before document read: remove all needs for that document
   - during document read: add needs from directives
   - after document read: update all needs for that document (with data such as `sections`)
   - cache (unprocessed) needs

- after caching (unprocessed) needs and before write/analysis phase
   -  process needs (resolving dynamic values, computing backlinks etc)

- write/analysis phase
   - retrieve views of the processed needs (to be filtered, read, etc) 

----

## Places where needs are exposed to the user

In configuration:

- `needs_warnings` (when string): list of needs, no parts (to `process_warnings`)
- `needs_builder_filter`: list of needs, no parts (to `filter_needs`)

In roles:

- `need_count`:
  - content: list of needs + parts (to `filter_needs`)
- `need_func`:
  - replacement of dynamic functions: map of id to need, no parts

In directives:

- need directives:
  - resolution of dynamic values containing dynamic functions: map of id to unprocessed need, no parts
  - replacement of dynamic functions in content and `style` option: map of id to need, no parts
- `needbar`:
[...]

Issue #1219, opened 2024-08-15
Performance improvement: Optimize queries for needflow, filters etc.


arwedus:
In our project with ~3800 pages, ~20.000 internal needs and ~20.000 imported needs, ~10.000 external needs, and ~700 needflow diagrams, we saw a dramatic increase in sphinx build html time (with external needs resolved), from former 20-30 minutes to anything between 2-5 hours,when adding another 10000 imported needs. A performance analysis showed that build progress stalls in processing of the needflow directive, even if the plantuml binary is configured to just echo "plantuml disabled".

Also, filter queries for needextract are very slow (taking seconds, each) in this large project, with most of them being like:

```
:filter_func: id = ["REQ_SW_12345"]
```

My assumption is that the internal needs data structure is a list, and thus has O(N) cyclomatic complexity for ID search operations.
Would it be possible to change sphinx-needs to use a dict instead, allowing O(1) access to needs by their ID? See https://wiki.python.org/moin/TimeComplexity

chrisjsewell:
> Would it be possible to change sphinx-needs to use a dict instead

Heya @arwedus it is already a dictionary,
the `filter_func` you show though is obviously a very "special case", i.e. selecting a single need via the ID field,
which would be the only time when one could think to just use the dictionary lookup.

It would make more sense, just to have an additional option on `needextract`, something like `select`, which allows for a list of IDs to initially select (maybe before applying `filter_func` if also present)

chrisjsewell:
Closed in https://github.com/useblocks/sphinx-needs/pull/1281

PR #1276, opened 2024-09-04
🔧 split `filter_needs` func by needs type

chrisjsewell:
This makes it clearer what representation of needs we are trying to filter (see #1264).

The idea is that eventually we might be able to analyse the filter string and perform "short-circuit" filters,
for example, if it is `id == "abc"` and `filter_needs_view`, then all we actually need to perform is an `O(1)` lookup of the mapping, rather than perform an `O(N)` iteration over all the needs.

This then begins to address #1219.

### DIFF
--- a/sphinx_needs/filter_common.py
+++ b/sphinx_needs/filter_common.py
@@ -20,6 +20,7 @@ from sphinx_needs.config import NeedsSphinxConfig
 from sphinx_needs.data import (
     NeedsFilteredBaseType,
     NeedsInfoType,
+    NeedsMutable,
     NeedsPartsView,
     NeedsView,
     SphinxNeedsData,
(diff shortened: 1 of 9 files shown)


sphinx_needs/filter_common.py
```diff
@@ -297,8 +298,65 @@ def intersection_of_need_results(list_a: list[T], list_b: list[T]) -> list[T]:
...
+        filter_string,
+        current_need,
+        location=location,
+        append_warning=append_warning,
+    )
+
+
 @measure_time("filtering")
-def filter_needs(
+def _filter_needs(
```
ubmarco, line 359: need to deprecate the old name?

ubmarco (approved):
Approved with a comment

sphinx_needs/filter_common.py
chrisjsewell: I think no, because this is not part of the public api (i.e. https://sphinx-needs.readthedocs.io/en/latest/api.html), and anyhow it will force people to stop using it if they were 😅

chrisjsewell: (in principle we should change almost every module to have an `_`, but eurgh that would be annoying 😅

PR #1277, opened 2024-09-05
✨ Add `get_needs_view` function to public API

chrisjsewell:
This allows users to access to the (immutable) resolved needs during the write/analysis phase (for example https://github.com/useblocks/sphinx-modeling, cc @ubmarco, also see #1264 for more info)

To ensure it is not misused, checks are now put in-place to disallow modification of needs after the needs have been resolved.
This actually surfaced a few issues where the resolved needs were mistakenly accessed, instead of the unresolved (mutable) ones.

The other reason for doing this, is that if we "know" we are only accessing immutable needs, then we will be able to improve filtering performance, by performing lazy indexing etc (see #1281)

### DIFF
--- a/sphinx_needs/data.py
+++ b/sphinx_needs/data.py
@@ -730,6 +730,8 @@ class SphinxNeedsData:
         .. important:: this should only be called within the read phase,
             before the needs have been fully collected and resolved.
         """
+        if self.needs_is_post_processed:
+            raise RuntimeError("Needs have already been post-processed.")
         self._env_needs[need["id"]] = need
 
     def remove_need(self, need_id: str) -> None:
@@ -738,6 +740,8 @@ class SphinxNeedsData:
         .. important:: this should only be called within the read phase,
             before the needs have been fully collected and resolved.
         """
+        if self.needs_is_post_processed:
+            raise RuntimeError("Needs have already been post-processed.")
         if need_id in self._env_needs:
             del self._env_needs[need_id]
         self.remove_need_node(need_id)
@@ -748,6 +752,8 @@ class SphinxNeedsData:
         .. important:: this should only be called within the read phase,
             before the needs have been fully collected and resolved.
         """
+        if self.needs_is_post_processed:
+            raise RuntimeError("Needs have already been post-processed.")
         for need_id in list(self._env_needs):
             if self._env_needs[need_id]["docname"] == docname:
                 del self._env_needs[need_id]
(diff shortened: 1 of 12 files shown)


tests/test_needs_external_needs_build.py
```diff
@@ -36,7 +36,10 @@ def test_doc_build_html(test_app, sphinx_test_tempdir):
         ["sphinx-build", "-b", "html", "-D", rf"plantuml={plantuml}", src_dir, out_dir],
         capture_output=True,
     )
-    assert not output_second.stderr
+    assert output_second.stderr.decode("utf-8").splitlines() == [
```
chrisjsewell, line 39: it seems that previously this was incorrect; one would assume if you get these warnings on the first build, then there is no reason you should not get them on a cached re-build

sphinx_needs/functions/common.py
```diff
@@ -170,14 +170,15 @@ def copy(
         need = needs[need_id]
 
     if filter:
+        location = (
+            (need["docname"], need["lineno"]) if need and need["docname"] else None
+        )
         result = filter_needs_view(
-            needs,
+            needs,  # type: ignore[arg-type]
```
chrisjsewell, line 177:
```suggestion
            needs,
```

sphinx_needs/functions/common.py
```diff
@@ -170,14 +170,15 @@ def copy(
         need = needs[need_id]
 
     if filter:
+        location = (
+            (need["docname"], need["lineno"]) if need and need["docname"] else None
+        )
         result = filter_needs_view(
             needs,
```
chrisjsewell, line 177: this is problematic, and in fact any use of `needs` in dynamic functions, because we do not know if we really need to account on whether we are calling it from the "resolution phase", i.e. dynamic functions in need fields, or the analysis phase, e.g. dynamic functions in need content

sphinx_needs/data.py
```diff
@@ -762,15 +768,23 @@ def get_needs_mutable(self) -> NeedsMutable:
...
         .. important:: this should only be called within the write phase,
-            after the needs have been fully collected
-            and resolved (e.g. back links have been computed etc)
+            after the needs have been fully collected.
+            If not already done, this will ensure all needs are resolved
+            (e.g. back links have been computed etc),
+            and then lock the data to prevent further modification.
         """
+        if not self.needs_is_post_processed:
+            from sphinx_needs.directives.need import post_process_needs_data
```
ubmarco, line 785: Can you raise an exception if not in the right phase?

chrisjsewell:
This has been subsumed by #1281

PR #1281, opened 2024-09-05
 ♻️ Replace need dicts/lists with views (with fast filtering)

chrisjsewell:
This PR addresses issues in performance scalability of needs filtering:

- In a known user project, with ~40,000 needs, there are 5316 individual filter calls, across directives such as `needarch`, `needtable`, `needpie`, etc. Each takes on average 0.33 seconds, cumulative totalling **1751 seconds**
- Applying this PR reduces the average time to 0.033 seconds, cumulatively totalling **181 seconds**

----

The main issue with filtering (a.k.a. querying) needs, is that it requires looping over every need and executing a Python evaluation of the filter string (e.g. `id == "xxx"`), which scales with `O(N)` time.

To reduce this to `O(1)` time, for common patterns, we look to do two things:

1. More tightly control access to needs data across its lifecycle (see #1264): 
   By making needs data strictly immutable/read-only during the write/analysis phase (after it has been fully collected and post-processed), and moving access behind an abstract interface (as opposed to a standard dictionary),
   we can perform indexing on standard need fields, making value lookups `O(1)`

2. Analysing the filter string for common patterns (by parsing/analysing the Python AST), we can then utilise *index lookups* rather than *row scans* in these cases (akin to [SQL query plans](https://sqlite.org/eqp.html)).

---

One complication in creating the indexes and abstract interfaces, is that within the code base there are essentially two ways of accessing needs data:

- As a mapping of need ID to need data (now `NeedsView`)
- As a list of both needs and "expanded" parts (now `NeedsAndPartsListView`)

Particularly when it comes to e.g. `id == "xxx"`, this then has different meanings, as the former is only filtering for need IDs and the latter is filtering for both Need IDs and part IDs

---

It is of note that this will be breaking for any projects that currently attempts to mutate needs data within the write phase.
This has been mitigated by the addition of two sphinx events, which give more *precise* control for this use case:

- `needs-before-post-processing`: callbacks `func(app, needs)` are called just before the needs are post-processed (e.g. processing dynamic functions and back links)
- `needs-before-sealing`: callbacks `func(app, needs)` just after post-processing, before the needs are changed to read-only

Additionally, `env.needs_all_needs` has been replaced with `env._needs_all_needs`, since this "raw" data-structure should not be accessed directly.

The `get_needs_view` function has been added to the `sphinx_needs.api` to mitigate this, and it should perhaps be made clearer that users should NOT be accessing any `sphinx_needs` API outside this module.

---

Because filter strings are (Turing complete) Python code, it is fundamentally impossible to fully process all such strings via AST parsing.
The new analysis code could be improved to recognise more patterns in the future, but really it would be ideal to move to a more "well-defined" filter syntax, such as the [SQLite boolean expressions](https://www.sqlite.org/lang_expr.html#booleanexpr).

Additionally, it may be ultimately beneficial to move the storage of needs data from an in-memory Python structure, to something like an sqlite database.

---

Note, a follow-up PR will look to add configuration for warning about particularly long-running filtering code, so that it may be accessed for improvements.

### DIFF
--- a/sphinx_needs/views.py
+++ b/sphinx_needs/views.py
@@ -1,0 +1,225 @@
+from __future__ import annotations
+
+from itertools import chain
+from typing import TYPE_CHECKING, Iterable, Iterator, Mapping
+
+if TYPE_CHECKING:
+    from sphinx_needs.data import NeedsInfoType
+
+
+class _Indexes:
+    """Indexes of common fields for fast filtering of needs."""
+
+    __slots__ = ("is_external", "status", "tags", "types", "type_names")
+
+    def __init__(
+        self,
+        is_external: dict[bool, list[str]],
+        types: dict[str, list[str]],
+        type_names: dict[str, list[str]],
+        status: dict[str | None, list[str]],
+        tags: dict[str, list[str]],
+    ) -> None:
+        self.is_external = is_external
+        self.types = types
+        self.type_names = type_names
+        self.status = status
+        self.tags = tags
+
+
+class NeedsView(Mapping[str, "NeedsInfoType"]):
+    """A read-only view of needs, mapping need ids to need data,
+    with "fast" filtering methods.
+
+    The needs are read-only and fully resolved
+    (e.g. dynamic values and back links have been computed etc)
+    """
+
+    __slots__ = ("_needs", "_maybe_indexes", "_maybe_len")
+
+    def __init__(
+        self,
+        *,
+        _needs: Mapping[str, NeedsInfoType],
+        _indexes: _Indexes | None = None,
+    ):
+        """Create a new view of needs,
+
+        .. important:: This class is not meant to be instantiated by users,
+            a singleton instance is managed by sphinx-needs.
+
+            Instead, use the filter methods to create new views.
+        """
+        self._needs = _needs
+        self._maybe_len: int | None = (
+            None  # Cache the length of the needs when first requested
+        )
+        self._maybe_indexes = _indexes
+
+    def _get_indexes(self) -> _Indexes:
+        """Lazily compute the indexes for the needs, when first requested."""
+        if self._maybe_indexes is None:
+            _idx_is_external: dict[bool, list[str]] = {}
+            _idx_types: dict[str, list[str]] = {}
+            _idx_type_names: dict[str, list[str]] = {}
+            _idx_status: dict[str | None, list[str]] = {}
+            _idx_tags: dict[str, list[str]] = {}
+            for id, need in self._needs.items():
+                _idx_is_external.setdefault(need["is_external"], []).append(id)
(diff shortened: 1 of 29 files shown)


sphinx_needs/views.py
```diff
@@ -0,0 +1,225 @@
...
+    after resolution (e.g. back links have been computed etc)
+
+    The parts are created by creating a copy of the need for each item in ``parts``,
+    and then overwriting a subset of fields with the values from the part.
+    """
+
+    __slots__ = ("_needs", "_maybe_len")
+
+    def __init__(self, *, _needs: NeedsView):
+        """ "Initialize a new view of needs.
```
ubmarco, line 197:
```suggestion
        """Initialize a new view of needs.
```

sphinx_needs/views.py
```diff
@@ -0,0 +1,225 @@
...
+        return iter(self._needs.values())
+
+    def __len__(self) -> int:
+        return len(self._needs)
+
+    def to_map_view(self) -> NeedsView:
+        return self._needs
+
+
+class NeedsAndPartsListView:
```
ubmarco, line 186: How would I get all parts of a certain need?

ubmarco (commented):
I pretty much like the idea of this. Users need to be informed about these capabilities in the docs, so they don't end up looping over `dict(SphinxNeedsData(app.env).get_needs_view())` recursively to filter things.
`needs_all_needs` has to be deprecated and then removed.

sphinx_needs/views.py
chrisjsewell: well it would be via; https://github.com/useblocks/sphinx-needs/blob/dd79aa3f18b3f8ee7a3a7ab8d7bb30b77dc5232b/sphinx_needs/roles/need_part.py#L44

although to note, this is not currently part of the "public API" (and never has been)

to note also, this `NeedsAndPartsListView` view is added separately so as not to be back-breaking for current user code (see https://github.com/useblocks/sphinx-needs/issues/1264),
if it was not for this, I would probably have just added an iter method to `NeedsView`

chrisjsewell:
>  Users need to be informed about these capabilities in the docs

yep, I've already done this is a little in this PR, although obviously want to clarify the PR/API before bothering to add more docs.

> `needs_all_needs` has to be deprecated and then removed.

I would note that `sphinx.BuildEnvironment.needs_all_needs` was never part of the "public API" 😅,
i.e. not in the `sphinx_needs.api` module,
so I would question how much it really needs to be "deprecated" (PACE/bosch have already agreed that it is "their problem" if they are using anything like this)

The needs, at least fo now, will still be stored on the environment, but perhaps changing it to `sphinx.BuildEnvironment._needs_all_needs`, with the `_` will make it clearer that this is not for public access

ubmarco (approved):
Just an awesome PR, approved!

PR #1284, opened 2024-09-06
🐛 Fix `need_part` with multi-line content

chrisjsewell:
fixes #1283


ubmarco (approved):
Nice, I was missing this in the past, too :)

PR #1285, opened 2024-09-06
🐛 Fix `needreport` `usage` numbers

### LINKED ISSUE #1110: needreport usage should count needs in post-process
I've also just realised there is a bug in this directive:

The `usage` flag creates a count of user defined need types.
However, it does this during the processing of the directive, rather than in a post-processing step.
Therefore, the count will usually be incorrect.

_Originally posted by @chrisjsewell in https://github.com/useblocks/sphinx-needs/issues/1105#issuecomment-1937172837_

chrisjsewell:
these need to be post-processed

fixes #1110


ubmarco (approved):
LGTM

PR #1286, opened 2024-09-06
🐛 Fix `parent_need` propagation from external/import

### LINKED ISSUE #979: Bug: nested needs in needimport do not render correctly in needflow
When using `parent_needs` inside a needs.json and importing it with needimport, they are not nested properly in needflow.
If you look at the :debug: planuml code, you can see, that the nodes are additionaly added as root noodes.
This is due to the failing check here:
https://github.com/useblocks/sphinx-needs/blob/377e6855cb582e57eafded26ca9b0937e596a73a/sphinx_needs/directives/needflow.py#L246C27-L246C27
There, only parent_need is checked.
To supply parent_need in needs.json additionally is not possible because it is deleted here:
https://github.com/useblocks/sphinx-needs/blob/78413f6ca72c176abc5fc8250e0883faa5c92644/sphinx_needs/directives/needimport.py#L216

A possible solution could be to add `parent_need `to the list here:
https://github.com/useblocks/sphinx-needs/blob/78413f6ca72c176abc5fc8250e0883faa5c92644/sphinx_needs/needs.py#L583

But that might conflict with this check:
https://github.com/useblocks/sphinx-needs/blob/78413f6ca72c176abc5fc8250e0883faa5c92644/sphinx_needs/needs.py#L677
which uses this List https://github.com/useblocks/sphinx-needs/blob/78413f6ca72c176abc5fc8250e0883faa5c92644/sphinx_needs/utils.py#L55 

So, the topic seems a bit complicated and maybe with other side-effects.

Maybe this could be revisited and cleaned up.

As an end user i would prefer to only pass either parent_needs or parent_need in the needs.json file.

chrisjsewell:
fixes #979


ubmarco (approved):
LGTM

PR #1292, opened 2024-09-10
✨ Add `ids` option for `needimport`

chrisjsewell:
A more performant alternative to the `filter` option.

This PR also adds some other performance optimisations;
caching the schema load and not reading the import file twice (during validation)

Also, add to and improve the `needimport` tests


sphinx_needs/directives/needimport.py
```diff
@@ -115,21 +112,21 @@ def run(self) -> Sequence[nodes.Node]:
                     f"Could not load needs import file {correct_need_import_path}"
                 )
 
-            errors = check_needs_file(correct_need_import_path)
+            try:
+                with open(correct_need_import_path) as needs_file:
+                    needs_import_list = json.load(needs_file)
+            except json.JSONDecodeError as e:
```
ubmarco, line 118: `OSError` should also be caught, but there is already a TODO

sphinx_needs/directives/needimport.py
```diff
@@ -146,6 +143,13 @@ def run(self) -> Sequence[nodes.Node]:
 
         needs_config = NeedsSphinxConfig(self.config)
         data = needs_import_list["versions"][version]
+
+        if ids := self.options.get("ids"):
+            id_list = [i.strip() for i in ids.split(",") if i.strip()]
```
ubmarco, line 148: Can IDs contain a dynamic function, so you would have to use the 'split with dyn' function?

sphinx_needs/directives/needimport.py
```diff
@@ -184,8 +188,13 @@ def run(self) -> Sequence[nodes.Node]:
         needs_list = needs_list_filtered
 
         # tags update
-        for need in needs_list.values():
-            need["tags"] = need["tags"] + tags
+        if tags := [
+            tag.strip()
+            for tag in re.split("[;,]", self.options.get("tags", ""))
```
ubmarco, line 193: same here, the new function should be used (or is that in a different open PR?)

ubmarco (changes requested):

sphinx_needs/directives/needimport.py
chrisjsewell: no definitely not, no dynamic functions here; this is intended for fast filtering

sphinx_needs/directives/needimport.py
chrisjsewell: this was already existing, but again no dynamic functions.
I feel you are confusing this directive with a `need` declaration directive, that is the only place we use dynamic functions in the directive options

sphinx_needs/directives/needimport.py
chrisjsewell, line 118:
```suggestion
            except (OSError, json.JSONDecodeError) as e:
```

chrisjsewell: done

chrisjsewell: (knowing what exceptions to handle is always a pain, I was reading this with interest recently https://discuss.python.org/t/extend-type-hints-to-cover-exceptions/23788/51)

ubmarco (approved):
Thanks for the PR

PR #1293, opened 2024-09-10
🔧  change type of need fields with `bool | None` to just `bool`

chrisjsewell:
It is unnecessary for these field to be nullable,
and this further reduces the size of `needs.json`, when `needs_json_remove_defaults = True`, since the default is clearer.


sphinx_needs/api/need.py
```diff
@@ -306,21 +306,21 @@ def run():
         "status": status,
         "tags": tags,
         "constraints": constraints,
-        "constraints_passed": None,
+        "constraints_passed": False,
```
ubmarco, line 309: Is `False` ok here or must it be `True` and the constraints checking pulls it to `False` if it fails? I ask because the schema defaults to `True`.

chrisjsewell: yeh I guess we can change this to `True`; in the actual analysis it changes it to `True` before assessing: https://github.com/useblocks/sphinx-needs/blob/a0d7abd5c92003f5c4fef1a04496913310afb591/sphinx_needs/need_constraints.py#L31

chrisjsewell, line 309:
```suggestion
        "constraints_passed": True,
```

chrisjsewell: changed

ubmarco (approved):
Thx for the update, approved

Issue #1148, opened 2024-03-16
Feature: needimport supports caching and configured needs.json sources


arwedus:
We've used needextract to show a few needs out of many, many needs imported from some other tool (via needs.json and needimport) in the right place in the documentation.

What bugs me is that the links never go to the place where the needextract is rendered, but always to the location where the original need is defined. This is a problem for the actual use case needextract was made for: If I want to generate a customer documentation (in HTML!) and want to show only the relevant parts of needs to the customer, I would do something like this:

```rst
Doc1: Needimport Page
=====================

.. just importing my architecture model or whatever...

.. needimport:: //build/model/architecture_needs.json
   :hide:
```

And in the actual docu:

```rst
Doc2: Customer docu
===================

Component Foo documentation
---------------------------

.. needextract::
   :filter: id == "COMP_FOO"
   :layout: complete

.. needextend:: COMP_FOO
   :+implements: REQ_SW_FOO_123495823

The Foo component does X and Y. It is implemented in ...
```

What I get with this approach is an error message / extension error at the end of the build process:

```
Extension error (sphinx_needs.needs):
Handler <function process_creator.<locals>.process_caller at 0x7f1eef8d5160> for event 'doctree-resolved' threw an exception (exception: Need COMP_FOO has no content node.)
```

## Alternatives considered

An alternative would be to "needimport" the same needs.json many times. But I fear that this would not be so performant for my use case (7000 needs). Maybe, if needimport would cache all need and not re-read the same json file twice, this might not be a big build time problem, but I also don't want to have people provide the same needs.json path in 1000 individual documents.

Also, one needs to be super careful with the filters to avoid importing the same needs twice, while at the same time not forgetting to import those needs which are linked but have no dedicated documentation page.

## Expectations

In order to not break the current behavior, maybe add an option to set whether a needextract "overrules" the location of the original link.

Also, add a user friendly error message instead of the extension error above.

Needextract "overruling" locations of original definition may bring it's own bunch of non-determinism problems. Unfortunately, attempts to implement it have also not been successful so far (see #689 ).

So let's rephrase the expectations more generally:

I want a directive to include selected imported needs exactly where I need them, with links going to this page. I want to define the path to the data source (needs.json) only once. I want the data source to be read only once.

## Solution idea

I have a suggestion to improve the yaaa-needimport:

- Define needimport sources in conf.py centrally, just like needs_external_needs, and reference them by key in the needimport directive.
[...]

PR #1297, opened 2024-09-12
✨ Add `needimport` caching and `needs_import_cache_size` configuration

chrisjsewell:
This PR introduces "lazy and size-bounded" caching for the reading of `needs.json` in the `needimport` directive.

This reads/writes to an in-memory cache, keyed on the path and mtime,
and is bounded by the `needs_import_cache_size` (configurable by the user),
which sets the maximum number of needs allowed in the cache,
to ensure we do not have large increases in build memory usage.

---

Note, in #1148 there was discussion of "centralised, pre-caching",
however, that is problematic because:

1. It means all import sources have to be read in for every build/re-build, irrespective of whether they actually may be used
2. this can introduce a noticeable increase in memory usage and time for re-builds
3. for parallel builds, all of this data will be copied to every process, irrespective of whether that processes actually uses it, again meaning potentially large multipliers of memory usage

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -1645,7 +1645,18 @@ keys:
             The related CSS class definition must be done by the user, e.g. by :ref:`own_css`.
             (*optional*) (*default*: ``external_link``)
 
+.. _needs_import_cache_size:
 
+needs_import_cache_size
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 3.1.0
+
+Sets the maximum number of needs cached by the :ref:`needimport` directive,
+which is used to avoid multiple reads of the same file.
+Note, setting this value too high may lead to high memory usage during the sphinx build.
+
+Default: :need_config_default:`import_cache_size`
 
 .. _needs_needextend_strict:
 
--- a/sphinx_needs/directives/needimport.py
+++ b/sphinx_needs/directives/needimport.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import json
 import os
 import re
-from typing import Sequence
+import threading
+from copy import deepcopy
+from typing import Any, OrderedDict, Sequence
 from urllib.parse import urlparse
 
 import requests
@@ -52,7 +54,8 @@ class NeedimportDirective(SphinxDirective):
 
     @measure_time("needimport")
     def run(self) -> Sequence[nodes.Node]:
-        # needs_list = {}
+        needs_config = NeedsSphinxConfig(self.config)
+
         version = self.options.get("version")
         filter_string = self.options.get("filter")
         id_prefix = self.options.get("id_prefix", "")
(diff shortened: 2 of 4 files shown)


arwedus:
Hi @chrisjsewell , first of all thanks for tackling this topic, even before I had conducted the performance measurement :-)

> Note, in #1148 there was discussion of "centralised, pre-caching", however, that is problematic because:
> 
>     1. It means all import sources have to be read in for every build/re-build, irrespective of whether they actually may be used
> 

How do we handle that the needs.json import source has changed? Sphinx will not consider the importing document changed, so in this case do we need to trigger a full re-build? Or does sphinx-needs have a way to handle this?

Is this maybe a general problem of needimport?

chrisjsewell:
> How do we handle that the needs.json import source has changed? Sphinx will not consider the importing document changed

It informs sphinx of the documents dependency on the file: https://github.com/useblocks/sphinx-needs/blob/fcf40f8efdd87faab9e1ba00c4d8b6e3d8652c3d/sphinx_needs/directives/needimport.py#L144

so yes sphinx will check if it's mtime is changed and re-build all dependant documents

sphinx_needs/directives/needimport.py
```diff
@@ -141,17 +157,17 @@ def run(self) -> Sequence[nodes.Node]:
                 f"Version {version} not found in needs import file {correct_need_import_path}"
             )
 
-        needs_config = NeedsSphinxConfig(self.config)
         data = needs_import_list["versions"][version]
 
+        # TODO this is not exactly NeedsInfoType, because the export removes/adds some keys
+        needs_list: dict[str, NeedsInfoType] = data["needs"]
```
ubmarco, line 163: don't know how to feel about calling a dictionary a list :)

sphinx_needs/directives/needimport.py
```diff
@@ -185,7 +203,9 @@ def run(self) -> Sequence[nodes.Node]:
                         location=(self.env.docname, self.lineno),
                     )
 
-        needs_list = needs_list_filtered
+        # note we need to deepcopy here, as we are going to modify the data,
+        # but we want to ensure data referenced from the cache is not modified
+        needs_list = deepcopy(needs_list_filtered)
```
ubmarco, line 208: maybe there's a code logic options that combines the 2 deepcopy

ubmarco (changes requested):
This should really help with many needimports of the same file.

docs/configuration.rst
```diff
@@ -1645,7 +1645,18 @@ keys:
...
             (*optional*) (*default*: ``external_link``)
 
+.. _needs_import_cache_size:
 
+needs_import_cache_size
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 3.1.0
+
+Sets the maximum number of needs cached by the :ref:`needimport` directive,
```
ubmarco, line 1655: `_ImportCache` stores
`self._cache: OrderedDict[tuple[str, float], dict[str, Any]] = OrderedDict()`
where `tuple[str, float]` is path and mtime.
Items are popped until the dict size is less than `needs_import_cache_size`.
So not the number of needs is relevant, but the number of distinct needs.json paths (+mtime combination).
If I'm right, please rephrase to make that clearer.

ubmarco (changes requested):
One more request from my side.

PR #1298, opened 2024-09-12
🐛 Fix removal of `Needextend` nodes

chrisjsewell:
We need to exhaust the entire `findall` iterator, before performing mutations on the doctree, otherwise this can lead to unexpected results (as has been found in the PACE project)


ubmarco (approved):
Maybe link an issue in the PR description if it exists.

PR #1299, opened 2024-09-13
👌 synchronise external/import need code

chrisjsewell:
Synchronise the code for preparing needs parameters, in both `needimport` and `external_needs`.

Additionally, allow propagation of `signature`/`sections` fields for external needs.


sphinx_needs/api/need.py
```diff
@@ -570,7 +572,7 @@ def add_external_need(
     external_css: str = "external_link",
     content: str = "",
     status: str | None = None,
-    tags: str | None = None,
+    tags: str | list[str] | None = None,
```
ubmarco, line 575: The docstring of `add_need` and `add_external_need` still says `:param tags: Tags as single string.`

sphinx_needs/external_needs.py
```diff
@@ -152,12 +158,11 @@ def load_external_needs(app: Sphinx, env: BuildEnvironment, docname: str) -> Non
...
-            need_params["links"] = need.get("links", [])
-            need_params["tags"] = ",".join(need.get("tags", []))
-            need_params["status"] = need.get("status")
-            need_params["constraints"] = need.get("constraints", [])
 
-            del need_params["description"]
+            # Remove unknown options, as they may be defined in source system, but not in this sphinx project
+            for option in list(need_params):
+                if option not in known_options:
+                    del need_params[option]
```
ubmarco, line 165: I feel this is worth a log message (type info) to the user.

ubmarco (changes requested):
Minor comments

sphinx_needs/external_needs.py
chrisjsewell: Eurgh, I'm worried this may lead to a tonne of "spurious" logging in existing projects if we do that, so want to consider if we would have a config or something to turn it off.
I would leave for a separate PR

sphinx_needs/api/need.py
chrisjsewell: fixed

ubmarco (approved):
LGTM

PR #1308, opened 2024-10-02
🔧 Remove deprecated `needfilter` directive

chrisjsewell:
This has been deprecated since v0.2.0, so high time to remove the code


danwos (approved):
wow, that's really old.

PR #1309, opened 2024-10-02
♻️ Add `needs_filter_max_time`/`needs_debug_filters`, deprecate `export_id`

chrisjsewell:
Add configuration options to debug and guard against long-running need filters:

- `needs_filter_max_time` warns if any call takes longer than the given time in seconds
- `needs_debug_filters` logs all calls and their run times to `debug_filters.jsonl`

This also deprecates the use of the `export_id` directive option and the writing of filters to the `needs.json`; this information should not be part of a need index and also was not compatible with running the `needs` builder (which skips the write phase)


danwos (approved):
Thanks for the new debug options :+1:

PR #1310, opened 2024-10-02
🔧 Make direct access to `env` attributes private

chrisjsewell:
These attributes should always be accessed via the API interface,
adding `_` to them makes this contract more explicit.


ubmarco (approved):
Internal APIs are internal, even when this change breaks somebody's workflow.
Approved.

Issue #1260, opened 2024-08-31
Sphinx-Needs version in needs.json


ubmarco:
As a consumer of needs.json, it would be helpful to have the Sphinx-Needs version that generated the needs.json as part of the file.

The file schema mutates over versions of SN, so this information can be used to define the expectation for structure and content.

PR #1311, opened 2024-10-02
👌 Add `creator` key to `needs.json`

chrisjsewell:
To record that the `needs.json` was created by `sphinx_needs` and also the version that created it.

### DIFF
--- a/sphinx_needs/builder.py
+++ b/sphinx_needs/builder.py
@@ -117,12 +117,7 @@ def build_needs_json(app: Sphinx, _exception: Exception) -> None:
     if isinstance(app.builder, NeedsBuilder):
         return
 
-    try:
-        needs_builder = NeedsBuilder(app, env)
-    except TypeError:
-        needs_builder = NeedsBuilder(app)
-        needs_builder.set_environment(env)
-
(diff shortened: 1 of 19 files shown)


ubmarco:
Fixes https://github.com/useblocks/sphinx-needs/issues/1260

ubmarco (approved):

arwedus:
@chrisjsewell , @danwos : What's the reason for this breaking change in the needs.json format?

https://github.com/useblocks/sphinx-needs/pull/1311/files#diff-66d8de18641b148e93348c5021b13c750846ff3edab47e669742022dd34082d9R17

-> By removing the filters and filters_amount fields, you have broken tools that use needs.json downstream. However, I don't know what these two field were used for, or if those were even used at all until now.

chrisjsewell:
> What's the reason for this breaking change in the needs.json format?

Because they have nothing to do with needs, as simple as that, and it does not work with the `needs` builder, that skips the entire write phase.
Thy will not be added back

If you want to know about filter usage use: https://sphinx-needs.readthedocs.io/en/latest/configuration.html#needs-debug-filters

PR #1312, opened 2024-10-02
♻️ Output `content` in `needs.json` not `description`

chrisjsewell:
I'm not sure if there was a legacy reason for this, but it is an unnecessary complication/confusion,
to convert between `content` and `description` and breaks the `needs_schema` and defaults.

Note, `needs.json` importing logic has been adapted to handle reading of both new and legacy formats


chrisjsewell:
> Should we decide for one field name and add a deprecation for reading needs.json when it's using the other name? 

will look at this in a separate PR, about optionally warning on unknown keys

PR #1314, opened 2024-10-03
👌 Add `needs_uml_process_max_time` configuration

chrisjsewell:
To guard against long-running `needuml` / `needarch` processing times. In particular, these can occur if the jinja template contains many calls to the filter function.


ubmarco (approved):
LGTM

PR #1315, opened 2024-10-03
♻️ Remove `target_id` core need field

chrisjsewell:
This is a duplicate of `id_parent` (since it is not modified when creating a part from a need)


ubmarco (approved):
Good change, the description of the field `ID of the data.` was not clear anyway.

PR #1316, opened 2024-10-04
👌 Warn on unknown need keys in external/import sources

chrisjsewell:
This PR centralises the declaration of what keys to omit from external/import sources, onto the `NeedsCoreFields` spec.
It then also uses `NeedsCoreFields` to identify unknown keys, and emits a warning if any are found.

Note, this could be breaking for existing builds, in that a warning may now be emitted,
but I decided not to put this under a "configuration flag", since users can simply choose to ignore it with `suppress_warnings`

---

One issue that was identified during this is the distinction of `title` vs `full_title`, whereby `title` is meant to be the potentially trimmed version of `full_title`. 
I think storing both fields is unnecessary and confusing; trimming of the title is a presentational issue that should be handled when necessary, not stored in the data.
Also, it is of note that the need directive pre-trims the title, leading to both `title` and `full_title` being trimmed

### DIFF
--- a/sphinx_needs/api/need.py
+++ b/sphinx_needs/api/need.py
@@ -49,6 +49,7 @@ def add_need(
     id: str | None = None,
     content: str | StringList = "",
     lineno_content: None | int = None,
+    doctype: None | str = None,
     status: str | None = None,
     tags: None | str | list[str] = None,
     constraints: None | str | list[str] = None,
@@ -290,7 +291,8 @@ def add_need(
         trimmed_title = title[: max_length - 3] + "..."
 
     # Calculate doc type, e.g. .rst or .md
-    doctype = os.path.splitext(env.doc2path(docname))[1] if docname else ""
+    if doctype is None:
+        doctype = os.path.splitext(env.doc2path(docname))[1] if docname else ""
 
     # Add the need and all needed information
     needs_info: NeedsInfoType = {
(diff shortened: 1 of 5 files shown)


ubmarco (approved):
LGTM

kreuzberger:
Hi! After upgrading from Sphinx to 4.1.0 from 3.0.0 i cannot build any more (Have Warnings treated as error on).
I ran into your case that i cannot import the needs.json files from my project (newly generated with 4.1.0) due to the warning with "Unknown keys".

The question is here for me why the keys should be unknown?

```
Unknown keys in import need source: ['testet', 'testet_back'] [needs.unknown_import_keys]
```
but these are valid "links" from
```
needs_extra_links = [
   {
      "option": "implements",
      "outgoing": "implementiert",
      "incoming": "ist umgesetzt durch",
   },
   {
      "option": "tests",
      "incoming": "testet",
      "outgoing": "ist getestet durch",
   }
]
```

chrisjsewell:
Hey @kreuzberger, the "name" of the link you mention above comes from the `option` key, i.e. `tests` not `testet`; the `incoming` key for a link is just for adding titles (https://sphinx-needs.readthedocs.io/en/latest/configuration.html#needs-extra-links)

kreuzberger:
ok, just saw the configuration flags for needs export and filtering.
i will adjust them to get the warnings removed. Thanks for patience :+1:

PR #1317, opened 2024-10-04
🔧 Simplify need directive

chrisjsewell:
`NeedDirective` has a lot of unused/unnecessary code, that appears to be legacy from before the `add_need` API was added. This is now removed.

Note, the trimming of the title, is already done in `add_need`, so not needed here.

Also, the replacement of `__` in `status` has been removed as has the documentation on "Multiline options" (added in https://github.com/useblocks/sphinx-needs/pull/283). I see no justification why you would want to use `status` this way, plus the documentation does not even make it clear that this is only for `status`


danwos (approved):
Looks okay for me.

PR #1328, opened 2024-10-10
🐛 Fix plantuml top-level import

chrisjsewell:
`sphinxcontrib.plantuml` is optional, should not be a top-level import


ubmarco (approved):
Thanks for the quick fix

PR #1330, opened 2024-10-13
🔧 Remove some internal fields from needs layout

### LINKED ISSUE #1329: 'delete: False' and 'jinja_content: False' shown by default in v4.0.0
Hello

I noticed that every need shows the options (I think since v4.0.0):
delete: False
jinja_content: False

Feels a bit cluttered to me: [Example in the Tutorial](https://sphinx-needs.readthedocs.io/en/latest/tutorial.html#T_SAFE)

So far, I couldn't figure out an easy way to hide those options from the output.
Is there an easy way to hide these options, or could this even be a bug?

chrisjsewell:
It should not be necessary to ever show these fields to users on the rendered need

fixes #1329

### DIFF
--- a/sphinx_needs/data.py
+++ b/sphinx_needs/data.py
@@ -126,7 +126,6 @@ NeedsCoreFields: Final[Mapping[str, CoreFieldParameters]] = {
     "delete": {
         "description": "If true, the need is deleted entirely.",
         "schema": {"type": "boolean", "default": False},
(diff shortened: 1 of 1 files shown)


ubmarco (approved):

ramevaf:
Will there be another version released? Using v4 without this fix will be quite annoying I guess 🫠

PR #1332, opened 2024-10-16
♻️ Improve access to dynamic configs: extra_options, functions, warnings

chrisjsewell:
These three configurations are complicated by the fact that they can be set both in the `conf.py` configuration, but also via functions from `sphinx_needs.api`.

This has lead to confusion, when to use `NEEDS_CONFIG` and when to use `NeedsSphinxConfig`, and indeed I have already had to fix numerous bugs when `NeedsSphinxConfig` was incorrectly used

In this PR we split access to these configs into:

- `NeedsSphinxConfig._extra_options`, `NeedsSphinxConfig._functions`, `NeedsSphinxConfig._warnings`, which access the "raw" sphinx configuration
- `NeedsSphinxConfig.extra_options`, `NeedsSphinxConfig.functions`, `NeedsSphinxConfig.warnings`, which access the "combined" sphinx config + API added values

We also make `NEEDS_CONFIG` private and merge in the use of `NEEDS_FUNCTIONS`.
This makes the code less prone to mistakes in the use of the wrong variable

### DIFF
--- a/sphinx_needs/api/configuration.py
+++ b/sphinx_needs/api/configuration.py
@@ -12,9 +12,9 @@ from sphinx.application import Sphinx
 from sphinx.util.logging import SphinxLoggerAdapter
 
 from sphinx_needs.api.exceptions import NeedsApiConfigException
-from sphinx_needs.config import NEEDS_CONFIG, NeedsSphinxConfig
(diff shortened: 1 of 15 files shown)


ubmarco (approved):
LGTM

PR #1337, opened 2024-10-25
✨ Add `needs_from_toml` configuration

chrisjsewell:
This configuration allows for a path to a [toml file](https://toml.io) to be specified.
In this case, needs configuration will be read from it and merged into the config from the `conf.py`.
An additional `needs_from_toml_table` config controls the toml table to read from.


sphinx_needs/config.py
```diff
@@ -301,6 +307,16 @@ def get_default(cls, name: str) -> Any:
             return _field.default_factory()
         return _field.default
 
+    from_toml: str | tuple[str, list[str]] | None = field(
+        default=None, metadata={"rebuild": "env", "types": (str, tuple, list)}
+    )
+    """Path to a TOML file to load configuration from.
+    
+    Optionally, a tuple of a path and a path to the section in the toml,
```
ubmarco, line 315: Was reading the sentence multiple times to understand what the tuple does. One more sentence would be good what tools and needs config leads to.

sphinx_needs/needs.py
```diff
@@ -356,6 +362,63 @@ def process_caller(app: Sphinx, doctree: nodes.document, fromdocname: str) -> No
...
+            "config",
+            None,
+        )
+        return
+    try:
+        with toml_file.open("rb") as f:
+            toml_data = tomllib.load(f)
+        for key in toml_path:
+            toml_data = toml_data[key]
+        assert isinstance(toml_data, dict), "Data must be a dict"
```
ubmarco, line 405: Can this also be a list like need types?

chrisjsewell: In what use case would you want it to be a list?

sphinx_needs/config.py
chrisjsewell: split off the section specification into a separate config option

docs/configuration.rst
```diff
@@ -79,6 +86,41 @@ Options
...
+   types = [
+      {directive="req", title="Requirement", prefix="R_", color="#BFD8D2", style="node"},
+      {directive="spec", title="Specification", prefix="S_", color="#FEDCD2", style="node"},
+   ]
+
+To specify a different section name to read from in the toml file, use the ``needs_from_toml_section`` option:
+
+.. code-block:: python
+
+   needs_from_toml_section = ["tool", "needs"]
```
ubmarco, line 120: Only understood this after reading the code.
Is section the right term? It's a table in toml. 

Can I use tool.needs in the list?

Would also change to plural 'sections' (or tables).

sphinx_needs/config.py
```diff
@@ -301,6 +307,16 @@ def get_default(cls, name: str) -> Any:
...
 
+    from_toml: str | None = field(
+        default=None, metadata={"rebuild": "env", "types": (str, type(None))}
+    )
+    """Path to a TOML file to load configuration from."""
+
+    from_toml_section: list[str] = field(
+        default_factory=lambda: ["needs"], metadata={"rebuild": "env", "types": (list,)}
+    )
+    """Path to the section in the toml file to load configuration from."""
```
ubmarco, line 318: Tables

sphinx_needs/needs.py
ubmarco: Nah I misunderstood the code at first.

ubmarco (changes requested):
Minor wording questions

sphinx_needs/config.py
chrisjsewell: changed

ubmarco (approved):

PR #1338, opened 2024-10-27
✨ Allow configuring description of extra option

chrisjsewell:
This PR allows for using a dict for a `needs_extra_options` item, that can have both a `name` and (optional) `description` key, e.g.:

```python
needs_extra_options = [
    "introduced",
    {"name": "impacts", "description": "What is the impact of this need?"},
]
```

---

Note, I envisage this extended dict format eventually also supporting other settings such as a data type, default, and possibly also limiting the need types it applies to, e.g.

```python
needs_extra_options = [
    {"name": "duration", 
     "description": "What is the impact of this need?",
     "type": "number",
     "default": 1,
     "need_types": ["req", "spec"]
    },
]
```

These though would require additional code changes

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -113,12 +113,12 @@ For example:
       {directive="spec", title="Specification", prefix="S_", color="#FEDCD2", style="node"},
    ]
 
-To specify a different `table path <https://toml.io/en/v1.0.0#table>`__ to read from in the toml file, use the ``needs_from_toml_table`` option.
+To specify a different `root table path <https://toml.io/en/v1.0.0#table>`__ to read from in the toml file, use the ``needs_from_toml_table`` option.
 For example to read from a ``[tool.needs]`` table:
 
 .. code-block:: python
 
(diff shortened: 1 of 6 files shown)


docs/configuration.rst
```diff
@@ -256,6 +256,20 @@ And use it like:
...
+    that will be output in the schema of the :ref:`needs.json <needs_builder_format>`.
+
+    For example:
+
+    .. code-block:: python
+
+       needs_extra_options = [
+           "my_extra_option",
+           {"name": "my_other_option", "description": "This is a description of the option"}
+       ]
```
danwos, line 271: A list of supported/allowed keys would be helpful, even if it is currently only  "name" and "description"

danwos (changes requested):
Looks great, just a hint for the docs

docs/configuration.rst
chrisjsewell: done

chrisjsewell: done 👍

danwos (approved):

PR #1347, opened 2024-11-01
♻️ Remove `delete` from internal needs and `needs.json`

chrisjsewell:
The `:delete:` option on a need directive deletes a need before creating/storing it, therefore it is literally impossible for it to be anything other than `False`.
Storing the field on a need is misleading, because it suggests that the need will be deleted (see e.g. #1383),
which is not possible with the current sphinx-needs logic.


danwos (approved):

christopheseyler:
That means a deleted requirement doesn't appear in the needs.json with the delete attribute set to true ?

chrisjsewell:
> That means a deleted requirement doesn't appear in the needs.json with the delete attribute set to true ?

deleted items have never appeared in the needs.json; that is the purpose of the need directive `delete` option

maybe they do currently, if someone makes a "custom" needs.json outside of sphinx-needs and imports it in, but this also should not be allowed

christopheseyler:
That is quite wierd. I though the purpose of this :delete: directive is to mark a requirement as "removed" and therefore, it would be a good practice to add a rational to this "delete" and then, to ensure the ID is not used by a new requirement

chrisjsewell:
> it would be a good practice to add a rational to this "delete" and then, to ensure the ID is not used by a new requirement

Quite possibly; I didn't add `delete` this is before my time, all I can tell you is the actual logic in the code 😅

If we want to ensure no other need is created with the same ID as one deleted, it would probably be better to keep track of these IDs separately.

danwos:
> That is quite wierd. I though the purpose of this :delete: directive is to mark a requirement as "removed" and therefore, it would be a good practice to add a rational to this "delete" and then, to ensure the ID is not used by a new requirement

No, the goal of `:delete:` is simply to remove it from the internal database, as it never got written down.
The main use case is variant handling, where two nearly identical need objects are defined with the same ID. But during the Sphinx build, only one is kept.
So having identical IDs in the sources is fine, as long as only one makes it to the final output.

PR #1353, opened 2024-11-22
Allow for multiple regex matches

### LINKED ISSUE #1351: Enable 'needs_string_links' to match more than one group.
Currently it's not possible to have more than one link inside an option via `needs_string_links`. It only matches the first option (as seen in the code linked below), even if the regex has multiple things that it caputures. 

What would be nice is maybe maybe through another option in the `needs_string_links` dictionary to enable it to generate links for all matches it finds. 

[Relevant Code](https://github.com/useblocks/sphinx-needs/blob/d5e84653d971fbccf8b9cb2a7dbcba23d1cda77a/sphinx_needs/utils.py#L461C1-L472C1)

MaximilianSoerenPollak:
This is a suggested change to close #1351. 
All it does is basically just iterating over each found match instead of just taking the first one.

I have tested it locally, and it works fine. 

Unsure however if there is other things that need to be changed that are effected by this.

### DIFF
--- a/sphinx_needs/utils.py
+++ b/sphinx_needs/utils.py
@@ -457,25 +457,24 @@ def match_string_link(
     try:
         link_name = None
         link_url = None
-        link_conf = matching_link_confs[
-            0
-        ]  # We only handle the first matching string_link
-        match = link_conf["regex_compiled"].search(data)
-        if match:
-            render_content = match.groupdict()
-            link_url = link_conf["url_template"].render(
-                **render_content, **render_context
-            )
-            link_name = link_conf["name_template"].render(
-                **render_content, **render_context
-            )
+        link_conf = matching_link_confs
+        for lc in link_conf:
+            match = lc["regex_compiled"].search(data)
+            if match:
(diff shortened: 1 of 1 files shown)


danwos:
Thanks for the PR, it looks promising and I like the feature enhancement.

However, can you please add/update a test case and maybe add some lines to the docs, to inform the user about this new possibility?

If any help is needed, just let me know.

MaximilianSoerenPollak:
One thing I wanted to investigate, It seems as if this already happens, at least in the Docs there is one place where multiple Values inside a option are matched and converted to links. The part of the docs is [here](https://sphinx-needs.readthedocs.io/en/latest/configuration.html#needs-string-links) under Example 6. 

Will check out if  there might not need to be any changes and maybe the regex in my testing was just not playing nice with something else.

If this isn't the case then I will adapt the docs and add some tests to cover this new feature.

danwos:
You are right, multiple findings for the same regex are supported.
But as far as I know, it does not check multiple regex, so for instance to have github and Jira links in one option.

Issue #1350, opened 2024-11-20
Regression/bug: esbonio support broken since 4.0.0


AlexanderLanin:
We're using https://github.com/swyddfa/esbonio for linting, preview etc, as are probably many others. This combination does no longer work since sphinx-needs 4.0.0. It works fine in sphinx-needs 2.1.0.

**Observation**
sphinx-needs throws https://github.com/useblocks/sphinx-needs/blob/6c97c47002a09c197da1a4969bded6028a5fb337/sphinx_needs/data.py#L781

**Reason GUESS**
Esbonio keeps sphinx "open" and runs build() on changes. I'm guessing this is in conflict with marking needs "post_processed" as they are revisted within the same sphinx process again.

AlexanderLanin:
It seems we found a quick fix:

```py
def setup(app: Sphinx):
    app.connect("env-before-read-docs", fix_sphinx_needs_restart_behavior)
    checks.add_warnings(app)

def fix_sphinx_needs_restart_behavior(_app: Sphinx, env, _documents):
    """ See https://github.com/useblocks/sphinx-needs/issues/1350 """
    # sphinx-needs 2.1.0
    if hasattr(env, "needs_warnings_executed"):
        env.needs_warnings_executed = False

    # sphinx-needs 4.1.0
    if hasattr(env, "_needs_warnings_executed"):
        env._needs_warnings_executed = False

    # sphinx-needs 4.1.0
    if hasattr(env, "_needs_is_post_processed"):
        env._needs_is_post_processed = False

```

PR #1358, opened 2024-11-29
Esbonio workaround docs

danwos:
Some docs update to document a solution to get esbonio working with the current release of Sphinx-Needs.

Based on #1350


chrisjsewell:
I would say this is deprecated by  #1359

Issue #1372, opened 2024-12-18
Problem importing filter-func in needpie object if the path is composed from multiple dots


Aymen-Soussi-01:
When creating a filter-func for needpie that contanis multiple . inside the import it, the build will fail and display this error: 
<Path>/index.rst:36: WARNING: does not contain a dot [needs.filter_func] 
 
This however is not the right error message as the error message correctly is too many values to unpack. 
 
To fix the error and make it more general, I propose that we split the import path from the right, taking anything to the right of the first dot (from the right) as the function and anything to the left as the module. 
This also works with the importer as written and will (if no dot is found) return the correct error message.

PR #1373, opened 2024-12-18
Allow filter-func in need-pie to have multiple dots in the import path. 

Aymen-Soussi-01:
I just changed the split of the path of filter-func to split on the first dot starting from the right.
This means anything to the right will be the function name and anything to the left will be the module import path 
This Fix will close the issue #1372

### DIFF
--- a/sphinx_needs/utils.py
+++ b/sphinx_needs/utils.py
@@ -326,7 +326,7 @@ def check_and_get_external_filter_func(
         return None
 
     try:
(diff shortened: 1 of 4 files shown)


danwos:
Nice catch and the fix seems to be quite small :)

However, can you please also add a test case for this use case, so that we know the fix is really working?

If any help is needed, just let me know.

danwos (approved):
Looks good for me. Thanks for the fix!! :+1:

PR #1385, opened 2025-01-26
🔧 Synchronize list splitting behaviour in need and needextend directives

chrisjsewell:
`NeedextendDirective` had a different function than `NeedDirective`, for splitting lists from directive options.

This PR combines the two, to make the behaviour consistent.


ubmarco (approved):
Thanks for cleaning this up.

sphinx_needs/api/need.py
```diff
@@ -767,15 +782,19 @@ def _split_list_with_dyn_funcs(
                 )
             text = text[2:]
         elif text[0] in ";|,":
```
ubmarco, line 784: This looks like there are multiple splitting characters. Maybe mention this in https://sphinx-needs.readthedocs.io/en/latest/filter.html#filter-options. The pages currently says `For :status:, :tags: and :types: values are separated by “;”.`. Have not found another mention of this.

chrisjsewell: Personally I would rather there were not these multiple characters, but that would be back breaking 🤷 
will look at docs in separate PR

PR #1386, opened 2025-01-26
♻️ Remove back link manipulation from `needextend`

chrisjsewell:
Back links are computed at the end of the need post-processing, after `needextend` have been applied.

Back links should always be in-sync with forward links, therefore it doesn't make sense to modify back links in this way.

Thankfully, the current availability of back link options in `needextend` directives is not documented, and so are unlikely to have been actually used.


ubmarco (approved):
Thanks for spotting this, LGTM.

PR #1387, opened 2025-01-26
♻️ Don't process dynamic functions on internal need fields

chrisjsewell:
For most "internal" need fields it does not make sense that these would be dynamic, and anyway this would fail since they are not string types.

In this PR, we skip dynamic function processing, for core fields that should not be altered by the user.

### DIFF
--- a/sphinx_needs/data.py
+++ b/sphinx_needs/data.py
@@ -68,6 +68,8 @@ class CoreFieldParameters(TypedDict):
     """Whether field should be excluded when importing needs (False if not present)."""
     exclude_json: NotRequired[bool]
     """Whether field should be part of the default exclusions from the JSON representation (False if not present)."""
+    allow_df: NotRequired[bool]
+    """Whether dynamic functions are allowed for this field (False if not present)."""
 
 
 NeedsCoreFields: Final[Mapping[str, CoreFieldParameters]] = {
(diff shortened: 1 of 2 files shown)


sphinx_needs/data.py
```diff
@@ -167,33 +174,38 @@ class CoreFieldParameters(TypedDict):
     "type": {
         "description": "Type of the need.",
         "schema": {"type": "string", "default": ""},
+        "allow_df": True,
```
ubmarco, line 177: How can the type and below type specific values be set by a df?

chrisjsewell: well thats a discussion to be had; I wouldn't expect them to, but they technically can already, so removing them would be a breaking change

ubmarco (approved):
Generally approve of this as this is something that users shouldn't be doing.
Not sure whether we break anyone's workflow.
I'd deprecate the dynamic functions on core fields, then finally remove that possibility. This PR makes our life simpler by making this configurable. Thanks.

sphinx_needs/functions/functions.py
```diff
@@ -228,14 +233,19 @@ def resolve_dynamic_values(needs: NeedsMutable, app: Sphinx) -> None:
...
     - ``**kwargs``: optional keyword arguments (specified in the function string)
     """
+
+    config = NeedsSphinxConfig(app.config)
+
+    allowed_fields: set[str] = {
+        *(k for k, v in NeedsCoreFields.items() if v.get("allow_df", False)),
+        *config.extra_options,
+        *(link["option"] for link in config.extra_links),
+        *config.global_options,
```
chrisjsewell, line 243:
```suggestion
```

sphinx_needs/functions/functions.py
```diff
@@ -228,14 +233,18 @@ def resolve_dynamic_values(needs: NeedsMutable, app: Sphinx) -> None:
     - ``*args``: optional arguments (specified in the function string)
     - ``**kwargs``: optional keyword arguments (specified in the function string)
     """
+
+    config = NeedsSphinxConfig(app.config)
+
+    allowed_fields: set[str] = {
+        *(k for k, v in NeedsCoreFields.items() if v.get("allow_df", False)),
+        *config.extra_options,
+        *(link["option"] for link in config.extra_links),
```
chrisjsewell, line 242:
```suggestion
        *(link["option"] for link in config.extra_links),
        *config.global_options,
```

PR #1388, opened 2025-01-26
🔧 Centralise extendable core need fields

chrisjsewell:


sphinx_needs/data.py
```diff
@@ -132,12 +138,14 @@ class CoreFieldParameters(TypedDict):
         "schema": {"type": ["string", "null"], "default": None},
         "show_in_layout": True,
         "exclude_external": True,
+        "allow_extend": True,
```
ubmarco, line 141: Isn't this evaluated during `add_need` so needextend will not see the need if it is deleted?

chrisjsewell: There is already a PR to completely remove this field, so it doesn't really matter: https://github.com/useblocks/sphinx-needs/pull/1347

chrisjsewell, line 141:
```suggestion
```

chrisjsewell: but removed anyway

ubmarco (approved):

PR #1389, opened 2025-01-26
👌 Add type warnings of extra options in external/import reads

chrisjsewell:
Currently, the value of all extra options is expected to be a string; other types are not supported in various aspects of sphinx-needs (such as `needextend`, dynamic functions and filtering), and in-fact are already silently converted to strings during processing,
see: https://github.com/useblocks/sphinx-needs/blob/92b47b129047df1b7aa9f40332d48deb369dc0c2/sphinx_needs/api/need.py#L791

This PR adds the `needs.mistyped_external_values ` and `needs.mistyped_import_values` warnings for non-string values, for `needs_external_needs` and `needimport` sources respectively. The warning is only emitted once per source, to avoid creating too many warnings.

Note, in the future it is hoped that more types will be supported via configuration in `needs_extra_options` (see https://github.com/useblocks/sphinx-needs/pull/1338)


ubmarco (approved):
LGTM

PR #1391, opened 2025-01-27
♻️ Make `needextend` argument declarative

chrisjsewell:
The argument for `needextend` can refer to either a single need ID or filter function.
Currently, the format cannot be known until all needs have been processed, and it is resolved during post-processing.
This is problematic for (a) user readability, (b) improving processing performance and issue feedback

This PR slightly modifies the argument processing to allow for two "explicit" formats:

- `<ID>`, if the argument is enclosed in `<>` it is always processed as a single ID
- `"filter string"`, if the argument is enclosed in `""` it is always processed as a filter string

If the string is not enclosed, then the logic is relatively similar to the existing:

- `ID`, if the argument is a single word (no spaces) then it is processed as a single ID
- otherwise, it is processed as a filter string

This change should not break most uses of `needextend`, but there are some corner-cases that would need to be adapted, e.g. a single `is_external` would now be processed as an ID, unless it was enclosed or expanded:

```restructuredtext
.. needextend:: "is_external"
.. needextend:: is_external is True
```

### DIFF
--- a/sphinx_needs/data.py
+++ b/sphinx_needs/data.py
@@ -567,7 +567,7 @@ class NeedsExtendType(NeedsBaseDataType):
     """Data to modify existing need(s)."""
 
     filter: str
-    """Filter string to select needs to extebd."""
+    """Filter string to select needs to extend."""
     filter_is_id: bool
     """Whether the filter is a single need ID."""
     modifications: dict[str, Any]
--- a/sphinx_needs/data.py
+++ b/sphinx_needs/data.py
@@ -566,9 +566,11 @@ class NeedsBarType(NeedsBaseDataType):
 class NeedsExtendType(NeedsBaseDataType):
     """Data to modify existing need(s)."""
 
-    filter: None | str
-    """Single need ID or filter string to select multiple needs."""
-    modifications: dict[str, str]
+    filter: str
+    """Filter string to select needs to extebd."""
+    filter_is_id: bool
+    """Whether the filter is a single need ID."""
+    modifications: dict[str, Any]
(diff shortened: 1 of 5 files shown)


sphinx_needs/data.py
```diff
@@ -566,9 +566,11 @@ class NeedsBarType(NeedsBaseDataType):
 class NeedsExtendType(NeedsBaseDataType):
     """Data to modify existing need(s)."""
 
-    filter: None | str
-    """Single need ID or filter string to select multiple needs."""
-    modifications: dict[str, str]
+    filter: str
+    """Filter string to select needs to extebd."""
```
ubmarco, line 570:
```suggestion
    """Filter string to select needs to extend."""
```

sphinx_needs/directives/needextend.py
```diff
@@ -54,20 +53,37 @@ def run(self) -> Sequence[nodes.Node]:
...
         elif strict_option == "FALSE":
             strict = False
 
+        modifications = self.options.copy()
+        modifications.pop("strict", None)
+
+        extend_filter = extend_filter.strip()
+        if extend_filter.startswith("<") and extend_filter.endswith(">"):
+            filter_is_id = True
+            extend_filter = extend_filter[1:-1]
```
ubmarco, line 70: Users provided `<>` or `""` will lead to an empty filter. Is that handled later on?

ubmarco (changes requested):
Just 2 minor remarks

ubmarco (approved):

PR #1393, opened 2025-01-29
✨ Add `c.this_doc()` check for `needextend` filters

chrisjsewell:
This allows for filtering of needs only in the same document as the `needextend` itself, e.g.

```restructuredtext
.. needextend:: c.this_doc() and status is None
   :status: open
```

Putting this under the `c` (for check) namespace allows for the possibility to add future special checks, without polluting the global namespace.

cc @arwedus

### DIFF
--- a/sphinx_needs/filter_common.py
+++ b/sphinx_needs/filter_common.py
@@ -582,7 +582,7 @@ def filter_single_need(
 
     filter_context["search"] = need_search
 
-    filter_context["n"] = NeedContext(need, origin_docname)
+    filter_context["c"] = NeedCheckContext(need, origin_docname)
 
     result = False
     try:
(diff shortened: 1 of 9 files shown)


ubmarco:
Test case would be nice to have.

chrisjsewell:
> Test case would be nice to have.

done

ubmarco (approved):

sphinx_needs/filter_common.py
```diff
@@ -549,16 +553,20 @@ def filter_single_need(
...
     Checks if a single need/need_part passes a filter_string
 
     :param need: the data for a single need
     :param config: NeedsSphinxConfig object
-    :param filter_compiled: An already compiled filter_string to safe time
-    :param need: need or need_part
     :param filter_string: string, which is used as input for eval()
     :param needs: list of all needs
+    :param current_need: set the current_need in the filter context as this, otherwise the need itself
+    :param filter_compiled: An already compiled filter_string to safe time
```
ubmarco, line 567:
```suggestion
    :param filter_compiled: An already compiled filter_string to save time
```

sphinx_needs/filter_common.py
```diff
@@ -549,16 +553,20 @@ def filter_single_need(
...
+    :param origin_docname: The origin docname that the filter was used called from, is any
```
ubmarco, line 568:
```suggestion
    :param origin_docname: The origin docname that the filter was called from, if any
```

ubmarco (changes requested):

ubmarco (approved):
Good change

Issue #890, opened 2023-03-09
Incorrect html rendering of math stuff with needextract


adegroote:
When using needextract, the math stuff is not properly rendered (cf both images)
![original_need_good_render](https://user-images.githubusercontent.com/377715/224081886-dc302330-1e98-40de-beea-7da1492d7ef8.png)
![extract_bad_render](https://user-images.githubusercontent.com/377715/224081912-e62435e8-d8a2-4666-a564-9c96dbc84a00.png)

Associated rst code

```
.. spec:: A spec with math stuff
   :id: MATH_STUFF

   A spec with nice math stuff :math:`x = y^2`

```

Note that if the needextract is on the same page, it is well rendered, but it is not the case as long as it is on  a different page

danwos:
Thanks for the bug report.

We had similar problems in the past with other sphinx functions and needextract.
Reason for this was, that we need to trigger the related sphinx feature on our own for the page, as sphinx has not noticed that this is needed for this additional page.

Fix should be a one liner, after I figured out the internal function which deals with math expressions.

PR #1397, opened 2025-02-04
Allow parsing needs options as multiline rst in custom layout

fournier-p:
Hello,

When using the `meta()` function in the custom layouts, it's not possible to parse the needs options as rst, only the links are handled in a custom way and the rest is parsed only with inline rst syntax.

I have the need to be able to put multiline rst for some custom options and thought a dedicated `meta_rst()` function would be the best way to handle those.

I made this in an hour to see the feasibility of such a function so don't hesitate to tell me if there is anything that needs to be changed or improved. I also didn't touch the tests, I would need a bit a guidance to add/update tests for this feature.

Status:
* [x] Implementation
* [x] Documentation
* [ ] Tests

### DIFF
--- a/docs/layout_styles.rst
+++ b/docs/layout_styles.rst
@@ -357,6 +357,7 @@ Available layout functions are:
 * :func:`meta_links <sphinx_needs.layout.LayoutHandler.meta_links>`
 * :func:`meta_links_all <sphinx_needs.layout.LayoutHandler.meta_links_all>`
 * :func:`meta_id <sphinx_needs.layout.LayoutHandler.meta_id>`
+* :func:`meta_rst <sphinx_needs.layout.LayoutHandler.meta_rst>`
 * :func:`image <sphinx_needs.layout.LayoutHandler.image>`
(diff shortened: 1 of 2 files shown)


chrisjsewell:
Heya @fournier-p thanks for the PR

I am a little hesitant about it, though, since it does not integrate well with the sphinx build system:
The layout creation here occurs in the "resolving" phase (see https://www.sphinx-doc.org/en/master/extdev/index.html#build-phases),
whereas, all parsing is already expected to happen in the "reading" phase.
Due to this, any parsing here will "miss out" on a lot of important transformations and post-processing.

For simple/local syntax, like making text bold/italic, ok this is not so much a problem.
But if there is any more complex syntax, like references, it will fail.

We already have issues with this for `needextract`, which attempts a similar thing, e.g. https://github.com/useblocks/sphinx-needs/issues/890

fournier-p:
Hey @chrisjsewell 

Indeed the goal was mainly for local syntax, like lists for example. I also tried using some directives inside the option and it failed.

I don't know if this scope of local rst syntax is ok ?

PR #1406, opened 2025-02-18
🔧 Deprecate use of dynamic functions in need type fields

chrisjsewell:
It is unintended and unlikely the dynamic functions would have been used in there fields, but just in case we issue a warning of the deprecation, so that we can remove them in the future

fields affected: `type`, `type_name`, `type_prefix`, `type_color`, `type_style`.

this follows on from #1387 and the fields now allowed are:

- ``title``
- ``status``
- ``tags``
- ``style``
- ``constraints``
- all `needs_extra_options`
- all `needs_extra_links`
- all `needs_global_options`

---

Note, there are still maybe some outstanding questions about where dynamic functions should be processed:
- should they be processed for the need title?
- should they be processed at all for needs coming from external and/or needimport `needs.json`?

### DIFF
--- a/docs/dynamic_functions.rst
+++ b/docs/dynamic_functions.rst
@@ -24,7 +24,7 @@ To refer to a dynamic function, you can use the following syntax:
 
       This need has id :ndf:`copy("id")` and status :ndf:`copy("status")`.
 
-Dynamic functions can be used for the directive title, and the following directive options:
+Dynamic functions can be used for the following directive options:
 
 - ``status``
 - ``tags``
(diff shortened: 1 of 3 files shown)


danwos (approved):

ubmarco:
I would keep dynamic function evaluation for the `title` field as I see a usecase for (partly) generated titles.

For imported and external needs, I think they should not be evaluated. The needs.json producer is in charge of providing non-dynamic field values. It also does not make sense to apply project-local dynamic functions to fields of foreign projects.

PR #1407, opened 2025-02-19
♻️ Remove `full_title` need field and only trim generated titles

chrisjsewell:
In the documentation for `needs_max_title_length`, it states:

> This option is used in conjunction with auto-generated titles as controlled by `needs_title_from_content`

However, this was not the case; it was being trimmed for all titles, irrespective of if they were auto-generated.

Trimming is now only applied to auto-generated titles (and the untrimmed title is still used for ID generation),

The existence of both `title` and `full_title` is confusing and unnecessary (in most cases these are equal), and so `full_title` is removed.


danwos (approved):

danwos:
Makes sense. Thanks for the implementation.

Issue #1409, opened 2025-02-24
Merge `global_options` with `extra_options`


chrisjsewell:
Currently, there are two ways to add custom fields to the needs schema:
 
- https://sphinx-needs.readthedocs.io/en/latest/configuration.html#needs-extra-options
- https://sphinx-needs.readthedocs.io/en/latest/configuration.html#needs-global-options

I feel having two ways

1. makes it more confusing for the user
2. makes the internal code more complex

I believe the features of the global options can essentially be merged into extra options and then deprecated:

- default; set a default value, if the field is not specified by the user
- filtered defaults; set a default value, if the field is not specified by the user, and the need passes a filter

This can be allowed by the new extra_option format #1338, e.g.

```python
needs_extra_options = [
    {
        "name": "example",
        "default": "my value",
        "filtered_defaults": [
            ('closed', 'status.lower() in ["done", "resolved", "closed"]'),
        ]
    }
]
```

danwos:
I like the syntax.

But what about internal options, like `style` or `layout`?
Are they defined in `needs_extra_options` as well?

If yes, maybe we shall call the config-var then just `needs_options` to indicate that it cares about all forms of options, not only new/extra ones.

I mean it's a huge breaking change. So in the end, we can rename stuff as well.
Or maybe have `needs_extra_options` and `needs_options` in parallel for a while, to write deprecation warnings.

chrisjsewell:
> But what about internal options, like `style` or `layout`?

yeh good point, I feel the problem is that `global_options` is conflating a few things that is can do:

1. setting defaults for `extra_options`, which I think is better co-located on `extra_option` items
2. setting defaults for `extra_links`, which I think is better co-located on `extra_links` items
3. setting defaults for internal fields; this should at least be constrained to options where it makes sense (similar to  #1387)
4. defining completely new fields; which I don't think it should be allowed to do; that should only be done in`extra_options`

Ideally I think `needs_extra_options` should be called `needs_custom_fields` and `needs_global_options`, if it is kept for internal fields, should be called `needs_field_defaults`.
this feels a lot more self-explanatory

PR #1410, opened 2025-02-24
♻️ Restrict `needs_global_options` keys

chrisjsewell:
As discussed in #1409,
`needs_global_options` has overlapping functionality with `needs_extra_options`.
It makes more sense that `needs_global_options` is only used to define defaults for fields,
whereas `needs_extra_options` is for specifying new fields.

This PR restricts the keys allowed to:

- any ``needs_extra_options`` item
- any ``needs_extra_links`` item
- ``collapse``
- ``constraints``
- ``hide``
- ``layout``
- ``status``
- ``style``
- ``tags``

Currently, just a deprecation warning is emitted, and no actual functionality is changed;
this will be left for change in a later version

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -279,13 +279,29 @@ And use it like:
 needs_global_options
 ~~~~~~~~~~~~~~~~~~~~
 .. versionadded:: 0.3.0
+.. versionchanged:: 5.1.0
 
-Global options are set on global level for all needs, so that all needs get the same value for the configured option.
+    Unknown keys are no longer accepted,
+    these should also be set in the :ref:`needs_extra_options` list.
+
+This configuration allows for global defaults to be set for all needs,
+for any of the following fields:
+
+- any ``needs_extra_options`` item
+- any ``needs_extra_links`` item
+- ``collapse``
+- ``constraints``
+- ``hide``
+- ``layout``
+- ``status``
+- ``style``
+- ``tags``
 
 .. code-block:: python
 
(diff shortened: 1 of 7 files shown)


danwos (approved):

danwos:
Thanks for the PR.

BTW: the way it works with the PR was how it was done releases ago.
You always needed both conf-options to add and modify the new extra_options.
But looks like this got somehow dropped/lost over time.

chrisjsewell:
> Thanks for the PR.
> 
> 
> 
> BTW: the way it works with the PR was how it was done releases ago.
> 
> You always needed both conf-options to add and modify the new extra_options.
> 
> But looks like this got somehow dropped/lost over time.

Ah good to know, so shouldn't break anything then 😅

santiagopagani:
@danwos Before this change, I was also using `needs_global_options` for defaulting the `content` field and enabling the `hide` tag for a special need type. This was working just fine, but now I'm getting warnings:
```
WARNING: needs_global_options 'content' must also exist in needs_extra_options, needs_extra_links, or ['constraints', 'layout', 'status', 'style', 'tags'] [needs.config]
WARNING: needs_global_options 'hide' must also exist in needs_extra_options, needs_extra_links, or ['constraints', 'layout', 'status', 'style', 'tags'] [needs.config]
```

However, `content` and `hide` are not extra options or extra links. They are intrinsic to need objects. Is there a simple way to get rid of this warning?

Thank you for your help.

PR #1416, opened 2025-03-03
List2need add list options

christopheseyler:
Superseed the PR #1345

This feature allows to set common options to all needs in a list defined by list2need directive.

Comparing to the PR #1345, I implemented a new **list-options** option with the ability to nest all options that will be common
(I removed the **hide** specific option as  it can easily be added in this new **list-options**

I left a regex matching on the list-options value allowing to aggregate options that could be defined in one list item through the meta-data feature.

### DIFF
--- a/docs/directives/list2need.rst
+++ b/docs/directives/list2need.rst
@@ -168,6 +168,27 @@ tags
 
 The tags ``A`` and ``B`` are attached to all ``NEED-A``, ``NEED-B``, ``NEED-C`` and ``NEED-D``.
 
+list-options
+~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


chrisjsewell:
Thanks @christopheseyler, this is definitely better, but bear with me a little on this as I want to have a bit more of a thought on it

chrisjsewell:
Now that #1790 is merged, `list2need` builds each need directly via `add_need` instead of generating reStructuredText through a template — so adding per-list or per-item options no longer means threading them through the Jinja round-trip. If this is still wanted, it should be a much more contained change on the new implementation.

Issue #1323, opened 2024-10-08
need2list results in incorrect source mapping


chrisjsewell:
From a parsing perspective, the code in https://github.com/useblocks/sphinx-needs/blob/master/sphinx_needs/directives/list2need.py is currently pretty horrible:
because it completely re-generates the text then uses `self.state_machine.insert_input` to insert the text into the text stream,
not only is this incompatible with non-rST parsers like MyST,
it also completely breaks any source mapping within that file; for both warnings and needs data
(see e.g. https://github.com/useblocks/sphinx-needs/blob/master/tests/__snapshots__/test_list2need.ambr)

There is a much better, "proper", way to do this I believe; using the new `generate_need` function and parsing the content with nested parsing,
I would like to do this, but don't have time straight away

chrisjsewell:
Fixed by the merged trilogy: #1788 pinned the directive's behaviour with a characterization suite, #1789 resolved the source line numbers through `get_source_info()`, and #1790 removed the text-regeneration + `state_machine.insert_input` splice entirely — each need is now built directly via `add_need`. `list2need` now works in MyST documents, generated needs carry their true per-item line numbers, and the file-wide lineno corruption after the directive is gone. The one residue is the per-need source *path* for include-nested content (the `docname` half discussed on #1692), which is a separate follow-up.

PR #1427, opened 2025-03-24
✨ Add `list-needs` directive

chrisjsewell:
The `list2need` directive has a number of notable issues:

- It uses a bespoke syntax that users have to learn, and contains numerous problematic edge cases (#1323)
- It uses jinja templating for parsing, which
  - breaks the source mapping for the entire file a `list2need` is part of
  - it is incompatible with other parsers such a MyST

This directive is intended to replace `list2need`, using parsing logic similar to the `list-table` directive, whereby the content is first parsed as a standard field list, then converted to needs.
This contains no bespoke syntax, preserves source mapping, and can be used within a MyST document.

See https://sphinx-needs--1427.org.readthedocs.build/en/1427/directives/list-needs.html

TODO: if agreed, look to "soft deprecate" `list2need`, so that at least new users don't use it, and old users can be guided to start porting

---

Comparision:

```restructuredtext
.. list2need::
   :types: req, spec, feat
   :tags: a,b,c

   * Need example on level 1
   * (NEED-002) Another Need example on level 1 with a given ID
     * Sub-Need on level 2 with status option set ((status='open'))
     * Another Sub-Need on level 2. Where this sentence will be used
       as content, the first one as title.
       * Sub-Need on level 3. With some rst-syntax support for
         the **content** by :ref:`list2need`
```

```restructuredtext
.. list-needs::
   :defaults:
      :tags: a,b,c

   :req: Need example on level 1
   :req id=NEED-002:
      Another Need example on level 1 with a given ID

      :spec status=open: Sub-Need on level 2 with status option set 
      :spec: Another Sub-Need on level 2.

         Where this sentence will be used as content, the first one as title.

         :feat collapse: Sub-Need on level 3.

            With all rst-syntax support for the **content** by :ref:`list-needs`
```

Most basic example:
```restructuredtext
.. list-needs::

    :req: Need example title
    :req: Another need example with nested needs.

       :spec: Sub-Need on level 2
       :spec: Another sub-need
```


ubmarco (changes requested):
Very useful improvement.
In addition to my comments I would directly deprecate list2needs.

ubmarco:
In its most basic form, the new directive looks like this:
```rst
.. list-needs::

    :req: Need example title
    :req: Another need example with nested needs.

       :spec: Sub-Need on level 2
       :spec: Another sub-need
```
I think this is also comprehensive and short. (I updated above example, it's messed up due to tabs/spaces mix.)
Only diff to `list2needs` is that the need type is explicitly specified. Cannot let the nesting level decide about that anymore.

We win
- standardized syntax that is part of the RST language
- syntactical separation of options (part of the field) and title markup (right of the field); this also means no custom syntax and no escaping mechanisms around brackets.
- no special symbol `.` that separate title and content; while I see the point of having the smallest possible syntax, it's hard to build a well-behaving parser around this (how to end a title with a dot?) without thinking about all the corner cases.
- avoid Jinja templating (which we could also fix in the list2needs directive)

We lose:
- list2needs users have to migrate their syntax - not sure if that is a heavily used feature; with a deprecation phase I think this is acceptable.

Any other opinions from the community?

chrisjsewell:
> Any other opinions from the community?

I would cc @christopheseyler, as this would essentially supersedes #1416 (and contains the `defaults` option)

christopheseyler:
I do not see the benefit comparing to the existing : the list2need provides a lightweight and more direct syntax to quickly write a list of requirements

`
.. list-need
    :req: Need example on level 1
    :req id=NEED-002: Another Need example on level 1 with a given ID
`

instead of 

`
.. need:: Need example on level 1
.. need:: Another Need example on level 1 with a given ID
    :id: NEED-002
`

The syntax of the actual list2need is more compact and more readable, IMO

chrisjsewell:
This is the answer to the question you asked here, @danwos — *"can we save the syntax
(mostly) but have a better implementation (without jinja)?"* — and to @ubmarco's *"avoid
Jinja templating (which we could also fix in the list2needs directive)"*: #1790 keeps
`list2need`'s syntax and its parsing exactly as they are, and replaces only the
generation step. The items are built with `add_need` instead of being rendered into a
need directive and handed back to the parser, which is what fixes the two things the
template cost:

| | before | after |
|---|---|---|
| a `{list2need}` fence in a MyST document | no needs at all, `MockingError: MockStateMachine has not yet implemented attribute 'insert_input'` | the needs are created, nesting and all |
| `lineno` of two items written on lines 10 and 11, and of an ordinary `.. req::` written on line 13 below them | 14, 20, 27 — and the error grows with each list in the file | 10, 11, 13 |

The rendered HTML of the existing `doc_list2need` test project is byte-identical before
and after, and the characterization tests added in #1788 pass unchanged apart from the
four assertions that pin what the template did.

That leaves #1416 free to proceed on top, and this PR takes no position on whether a
`list-needs` directive should exist as well — it only removes the implementation
objection to keeping this one.

chrisjsewell:
Closing: the reimplementation route this discussion converged on is now merged (#1788, #1789, #1790) — `list2need` keeps its existing syntax, loses the template/`insert_input` machinery, and now works in MyST documents with true per-item line numbers. Anything from the `list-needs` proposal still worth having (e.g. a `defaults` option) can be a follow-up on the new implementation.

PR #1429, opened 2025-03-24
✨ Add `is_import` need field

chrisjsewell:
This field records whether a need originated from a `needimport` source, similar to `is_external`


ubmarco (approved):
LGTM. Not sure how you intent to use that information. Depending on that, it might also be interesting where the imported need originated from (import key or path).

PR #1439, opened 2025-04-14
🔧 Use ubuntu-latest in CI

ubmarco:
Ubuntu 20.04 is phased out.


ubmarco:
According to https://github.com/actions/runner-images?tab=readme-ov-file#available-images the labels `ubuntu-latest` and `ubuntu-24.04` refer to the same image currently. They are updated every 2 years.
I feel `latest` is fine. We'll see on upcoming PRs in a year whether this was a bet well placed.

PR #1441, opened 2025-04-16
✨ Schema validation

ubmarco:
Experimental implementation of a schema validation, aka. ontology check.
This is based on jsonschema, but also supports network validations.

PR #1447, opened 2025-04-25
⬆️ Sphinx 8.1, drop Python 3.9

ubmarco:
Preliminary work for https://github.com/useblocks/sphinx-needs/pull/1441 in an isolated PR.


ubmarco:
CI fails for Windows only for Python 3.13. Need help or time investigating this.

ubmarco:
Windows problem fixed

sphinx_needs/directives/needimport.py
```diff
@@ -81,7 +82,7 @@ def run(self) -> Sequence[nodes.Node]:
         else:
             logger.info(f"Importing needs from {need_import_path}")
 
-            if not os.path.isabs(need_import_path):
+            if not posixpath.isabs(need_import_path):
```
chrisjsewell, line 85: If you use posixpath here, then you need to use posix paths for the entire logic in the function,
and confirm this is the actual behaviour you want on Windows.
It would also be better to use `pathlib.PurePosixPath` (see https://docs.python.org/3/library/pathlib.html)

sphinx_needs/directives/needuml.py
```diff
@@ -107,7 +108,7 @@ def run(self) -> Sequence[nodes.Node]:
         save_path = self.options.get("save")
         plantuml_code_out_path = None
         if save_path:
-            if os.path.isabs(save_path):
+            if posixpath.isabs(save_path):
```
chrisjsewell, line 111: same as above

chrisjsewell (changes requested):
see comments

sphinx_needs/directives/needimport.py
ubmarco: Look at the `:download:` role and going for `self.env.relfn2path()` instead of the custom logic.
This also removes the deprecated behavior.

chrisjsewell (approved):

PR #1448, opened 2025-04-26
👌 Reset directive option specs at start of build

ubmarco:
This fixes the problem, that the `option_spec` attributes are modified on a class level for:

- `NeedDirective`
- `NeedextendDirective`
- `NeedserviceDirective`

and thus remain changed for subsequent sphinx runs in the same interpreter session.

The problem was a finding of https://github.com/useblocks/sphinx-needs/pull/1441.

This is particularly of issue for SN test cases, which all run in the same Python interpreter,
and should ideally be reproducible, independent of any preceding tests run.

The PR defines a `reset_options_spec` class method for modified directives,
to revert the runtime changes during the `config-inited` event.


chrisjsewell (approved):
cheers!

Issue #1452, opened 2025-05-07
request to support `needs` as element in calc_sum


PhilipPartsch:
[calc_sum](https://sphinx-needs.readthedocs.io/en/latest/dynamic_functions.html#sphinx_needs.functions.common.calc_sum) does not provide the needs in scope to the filter function, so we cannot use them in the calc_sum filter string.

You can see in current [source](https://github.com/useblocks/sphinx-needs/blob/c25df13fd903bfbcf20bab3fccbdd09ec4e1ab2e/sphinx_needs/functions/common.py#L386) that we miss to pass `check_needs` as parameter to `filter_single_need`.

PR #1453, opened 2025-05-07
in calc_sum added needs.values() as parameter to filter_single_need()

PhilipPartsch:
try to implement #1452


PhilipPartsch:
@ubmarco, @danwos , @chrisjsewell : Could I get a review?

chrisjsewell (commented):
Heya, I would need to see a good use case of why this is required cheers

PhilipPartsch:
The Idea is to have the same filter algorithm as for needtable, needles, ...

PhilipPartsch:
Even the documentation specifies, we can use `needs` within filter string in calc_sum: https://sphinx-needs.readthedocs.io/en/latest/filter.html#filter-string

So the missing `needs` are a bug.

PR #1454, opened 2025-05-14
🔧 Allow template global_options

chrisjsewell:


ubmarco (approved):
Thanks.
The field `collapse` should also be allowed to be set via `needs_global_options`.

PR #1464, opened 2025-06-21
🛠️ Refactor Sphinx-Needs DataTables integration to use GridJS

iSOLveIT:
Changes for [issue #122](https://github.com/useblocks/ubtrace/issues/122):

- [x] Removed Data tables and all its dependencies
- [x] Use [GridJS](https://gridjs.io/docs/) to handle data table integration
- [x] Fix some testcases
- [x] Fix data table CSS to support both light and dark modes

## Motivation for this change:
Our used DataTables version is quite old (>~4 years) and is still based on jQuery, which was also removed from Sphinx some long time ago.
Also, the JS world has changed, docs get presented in React/Angular/Svelte apps, in which DataTables is causing some problems or at least is adding some unnecessary dependencies (jQuery).
Therefore, this PR cleans up our used JS environment and feels more modern now.


danwos (changes requested):
Looks already promising.
After building the official docs I found some problems:

* Pagination works fine, but we need an option how much elements shall be shown: 10, 25, 50, All.
Otherwise dealing with big tables will not be so comfortable.
* The sticky head is a nice idea.
  But it is confusing, as the header is transparent so that it shows row data as well:
  image

* If possible, let's remove the icons, as they may not match a specific theme and their look-alike depends on os/browser:
image

danwos:
@chrisjsewell what do you think?
This PR is also some kind of a clean-up, as JS libs are reduced.

danwos:
Another finding: Row style seems not to work

image
Source: https://sphinx-needs.readthedocs.io/en/latest/directives/needtable.html#style-row

image
Source: local build via `DOCS_THEME=furo sphinx-build -a -E docs docs/_build/html`

PR #1467, opened 2025-07-08
✨ Schema validation v2

ubmarco:
This is a follow-up on https://github.com/useblocks/sphinx-needs/pull/1441.
I plan not to merge https://github.com/useblocks/sphinx-needs/pull/1441 and merge this PR instead.

Differences:
- Aligning with standards in JSON schema for re-using subschemas via `$defs` and `$ref`
- Fully typed implementation including runtime checks of valid schema user input
- Auto inject the default string type if not given
- Replace `trigger_schema` with `select` which aligns with query language terminology
- Replace `trigger_schema_id` with the mentioned `$ref` mechanism
- New schemas root key `validate` with sub-keys `local` and `network` for the 2 validation types
- Network validation items does not allow the `select` key anymore as the selection happens by linking target needs.
  This cleans up an ambiguity in the other PR.
- Network validation errors now bubble up to the root json schema and are displayed to see exactly why the chain fails
- More network rule types for better control over debug schema output
- Rewrite test cases to use a declarative definition as yaml, so all pieces can be given in one place:
  - conf.py
  - ubproject.toml
  - index.rst
  - schemas.json
  - expected ontology warnings
- Simplified the code logic
- String patterns (regex) are constrained to a basic subset that works across engines
- Added docs
  - Examples and explanations
  - Comparison with `needs_warnings` and `needs_constraints` and migration path
- Many more test cases
- `items` with `minItems` and `maxItems` and `contains` with `minContains` and `maxContains` are now semantically equivalent to JSON schema spec


tests/schema/test_schema.py
```diff
@@ -0,0 +1,251 @@
...
+    "need_cnt",
+    [
+        10,
+        100,
+        1_000,
+        # 5_000,
+        # 10_000,
+    ],
+)
+@pytest.mark.benchmark
```
chrisjsewell: there is a `tests/benchmark` folder, which this should go in

sphinx_needs/builder.py
```diff
@@ -41,7 +41,8 @@ def write(
         updated_docnames: Sequence[str],
         method: str = "update",
     ) -> None:
-        return
+        # make sure schema validation is done
```
chrisjsewell:
```suggestion
        # we override this method, to stop any document output files from being written,
        # however, from this method triggers the `write-started` event,
        # which we still want for triggering schema validation.
        # TODO since sphinx 8.1 `Builder.write` is typed as `final` and a new `Builder.write_documents` method is added,
        # see https://github.com/sphinx-doc/sphinx/commit/d135d2eba39136941da101e7933a958362dfa999
        # once sphinx 7 is not supported, we should remove this `write` method and override `write_documents` to "do nothing"
```

sphinx_needs/builder.py
```diff
@@ -257,3 +258,39 @@ def build_needumls_pumls(app: Sphinx, _exception: Exception) -> None:
...
+
+    name = "schema"
+
+    def write(
+        self,
+        build_docnames: Iterable[str] | None,
+        updated_docnames: Sequence[str],
+        method: str = "update",
+    ) -> None:
+        # make sure schema validation is done
```
chrisjsewell, line 279:
```suggestion
        # we override this method, to stop any document output files from being written,
        # however, from this method triggers the `write-started` event,
        # which we still want for triggering schema validation.
        # TODO since sphinx 8.1 `Builder.write` is typed as `final` and a new `Builder.write_documents` method is added,
        # see https://github.com/sphinx-doc/sphinx/commit/d135d2eba39136941da101e7933a958362dfa999
        # once sphinx 7 is not supported, we should remove this `write` method and override `write_documents` to "do nothing"
```

docs/configuration.rst
```diff
@@ -39,10 +39,11 @@ Find below a list of all warnings, which can be suppressed:
 
 .. need-warnings::
 
-.. _inc_build:
+.. _`inc_build`:
```
chrisjsewell: can we do this formatting in a different PR please, so that it does not obscure what has been changed only for schema validation documenting

sphinx_needs/api/need.py
```diff
@@ -492,7 +492,6 @@ def _create_need_node(
         # nor will it find/add any child needs
         node_need["hidden"] = True
         return [node_need]
-
```
chrisjsewell: please remove this unconnected change

sphinx_needs/api/need.py
```diff
@@ -492,7 +492,6 @@ def _create_need_node(
...
     return_nodes: list[nodes.Node] = []
```
chrisjsewell:
```suggestion

    return_nodes: list[nodes.Node] = []
```

sphinx_needs/config.py
```diff
@@ -365,6 +408,54 @@ def get_default(cls, name: str) -> Any:
         default_factory=list, metadata={"rebuild": "env", "types": (list,)}
     )
     """Path to the root table in the toml file to load configuration from."""
+    schemas: SchemasFileRootType = field(
```
chrisjsewell: please rename this to something like `schemas_root` or whatever makes sense,
so that all schema options are `schema_<field>`
ideally I think we should have in the `ubproject.toml` for this to be under its own namespace e.g.

```toml
[needs.schema]
field = ...
```

We could do some magic in the `from_toml` function to support this

sphinx_needs/needs.py
```diff
@@ -159,6 +165,28 @@
...
+        )
+
+    try:
+        with Path(json_file).open("rb") as fp:
+            json_data = json.load(fp)
+        assert isinstance(json_data, dict), "Data must be a dict"
+    except Exception as exc:
+        raise NeedsConfigException(f"Could not load JSON file: {exc}") from exc
+
+    config["needs_schemas"] = json_data
```
chrisjsewell: Please use `NeedsSphinxConfig` for accessing this key, so that type safety is enforced.
I guess we should also check that not both `needs_schemas` and `needs_schemas_from_json` are set at the same time, and warn about this, so there is no silent override.

sphinx_needs/schema/config.py
```diff
@@ -0,0 +1,439 @@
...
+    default: NotRequired[str]
+    """Default value used in IDE scenarios for autocompletion."""
+
+
+class ExtraOptionNumberSchemaType(TypedDict):
+    """
+    Float extra option schema.
+
+    The option will still be stored as a string in Sphinx-Needs,
+    but during schema validation, the value will be coerced to the given type.
```
chrisjsewell, line 61: This is bad though, the values should be coerced to these from the start, if read from a (string) directive, or just checked if read from a `needs.json`.
This will definitely be the case in ubcode.
What is the path to change sphinx-needs, so that it can handle storing the data as the correct type?

chrisjsewell (changes requested):
Initial review comments

(without yet looking over the full implementation of the schema checking)

tests/schema/test_schema.py
ubmarco: Done

sphinx_needs/builder.py
ubmarco: Committed

docs/configuration.rst
ubmarco: Sorry, not intentional. Reverted.

sphinx_needs/api/need.py
ubmarco: Reverted

sphinx_needs/schema/config.py
ubmarco: While I agree with you that the internal types should change with the definition, my plan was to do that in a 2-step approach. Users should adopt the ontology in a non-breaking way (switch from warnings/constraints to this, transpile from SHACL, ...). This takes time and efforts. Then in the meantime we change the internal representation which affects at least filter functions, and data/schema in exported needs.json. My current feeling is changing the types is a bigger thing in the codebase that delays this PR. The adoption of the new interface and the schema feedback is already of good value for end users. We can switch the type system while adoption happens.

sphinx_needs/needs.py
ubmarco: Done, however type checking happens later in detail, left a comment for that.
I added a check if both schema_definitions and schema_definitions_from_json are given.

sphinx_needs/config.py
ubmarco: Done, the field is now called `needs_schema_definitions`

sphinx_needs/schema/config_utils.py
```diff
@@ -0,0 +1,578 @@
...
+    inject_type_extra_link_schemas(needs_config)
+
+    # check for disallowed regex patterns
+    validate_regex_patterns_extra_options(needs_config)
+    validate_regex_patterns_extra_links(needs_config)
+
+    if not needs_config.schema_definitions:
+        # nothing to validate
+        return
+
```
chrisjsewell, line 84: please emit a sphinx warning here, that schemas are in beta

chrisjsewell (changes requested):
see comment, emit a warning

then ok thats it for now; this PR can be merged, but there will be changes that need to be made when I get into it

sphinx_needs/schema/config_utils.py
ubmarco: Done

Issue #1437, opened 2025-04-14
`need_incoming` / `need_outgoing`/ `need_part` misuse sphinx reference role


chrisjsewell:
The anatomy of a reference role is that

- By default, the `title == target`, e.g. `` :name:`target` ``
- The title can be separated from the title using `<>`, e.g. `` :name:`title <target>` ``
- The reference can be disabled if the content starts with `!`, e.g. `` :name:`!target` `` or `` :name:`!title <target>` ``
- The `<` or `!` can be escaped with `\`, e.g.  `` :name:`title \<target>` `` or ``:name:`\!target` ``

Firstly, it is of note that for all `need` / `need_incoming` / `need_outgoing`/ `need_part`, they are set as `sphinx.XRefRole`.

For all of these, the disabling of the role with `!` works, i.e. a reference/part is not created for these.
It is not clear if this is even intended, nor is it documented.

For the title/target syntax, the `need` role works as expected.
For `need_incoming` / `need_outgoing`/ `need_part`, though this syntax is parsed, it is not used in an understandable way:
e.g. for `` :name:`title <target>` ``,
- for `need_incoming` / `need_outgoing` only the `target` is used and the `title` is disregarded
- for `need_part` only the `title` is used and the `target` is disregarded

---

Additionally, `need_part`, splits up id/content like `(id) content`, but does not account for space in between the  `)` and start of content

---

also there is no warning if `need_part` is not in a need directive, they are just silently ignored

PhilipPartsch:
Is it possible to support even `any` with `:need:`.

PR #1469, opened 2025-07-10
👌 Improve need part processing

chrisjsewell:
The `need_part`/`np` role is not a reference and should not use an `XRefRole` subclass (see #1437). 

In addition, the processing now correctly handles and warns on need parts that are not part of a need and `need` roles that reference an unknown part.


pyproject.toml
```diff
@@ -144,7 +144,7 @@ disable_error_code = ["no-redef"]
 
 legacy_tox_ini = """
 [tox]
-envlist = py10
+envlist = py310
```
ubmarco, line 147: my bad, thanks for the fix

sphinx_needs/roles/need_part.py
```diff
@@ -3,31 +3,72 @@
...
+        """Return the ID of the need this part belongs to."""
+        return self.attributes["need_id"]  # type: ignore[no-any-return]
+
+    @need_id.setter
+    def need_id(self, value: str) -> None:
+        """Set the ID of the need this part belongs to."""
+        self.attributes["need_id"] = value
+
+
+_PART_PATTERN = re.compile(r"\(([\w-]+)\)(.*)", re.DOTALL)
```
ubmarco, line 49: [re.DOTALL](https://docs.python.org/3/library/re.html#re.DOTALL) matches all symbols and newlines. AFAIK roles cannot span across newlines, correct?

ubmarco (approved):
LGTM in general, I just question the `re.DOTALL`

sphinx_needs/roles/need_part.py
chrisjsewell: no, role content can span across new lines

ubmarco: good to know, have never seen one in the wild

PR #1470, opened 2025-07-11
📚 Fix escape sequences

ubmarco:
This fixes the docs build warnings emitted by RST highlighting:
```
<unknown>:1: SyntaxWarning: invalid escape sequence '\.'
<string>:1: SyntaxWarning: invalid escape sequence '\.'
<unknown>:1: SyntaxWarning: invalid escape sequence '\w'
<string>:1: SyntaxWarning: invalid escape sequence '\w'
```

Not sure about the solution though, could also be the `need-example` directive has a bug.


chrisjsewell (approved):
cheers

PR #1478, opened 2025-08-01
‼️ Improve needs default field application (via needs_global_options)

chrisjsewell:
Previously defaults would be applied to any fields of a need with a "falsy" value, e.g. `None`, `False`, `0` , `""`, `[]`. ...

This is an issue, if the user wants to specifically set fields to these values, without them being overridden by defaults. Therefore, now defaults are only applied to fields with a missing or `None` value.

In the need directive context, this means that defaults will only be set for fields not specifically set in the directive options.

In the external/import need context, defaults will only be applied to missing keys in the `needs.json`, or fields with the value set to `null`.

For predicate defaults, the context passed to the expression has been restricted to a subset of fields, and these values are no longer mutated by previous defaults, i.e. each predicate gets the same immutable context data.
This makes these predicates more isolated / understandable, and allows for them potentially later being co-located with the type information.

----

‼️ Breaking

This PR may be breaking:

1. For users expecting defaults to be applied to external/import `needs.json`, although I would suggest this was never the intention for these defaults and that `needs.json` should, at least by default, be seen as static ingest data which has already had defaults etc applied.

2. For users with "exotic" predicate expressions, that potentially rely on the application of previous defaults.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -349,7 +349,22 @@ Defaults will be used if the field is not set specifically by the user and thus 
    }
 
 To set a default based on a one or more predicates, use the ``predicates`` key.
-These predicates are a list of (:ref:`filter string <filter_string>`, value), evaluated in order, with the first match set as the default value.
+These predicates are a list of ``(match expression, value)``, evaluated in order, with the first match set as the default value.
+
+A match expression is a string, using Python syntax, that will be evaluated against data from each need (before the resolution of defaults or dynamic functions etc):
+
+- `id` (`str`)
+- `type` (`str`)
+- `title` (`str`)
+- `tags` (`tuple[str, ...]`)
+- `status` (`str | None`)
+- `docname` (`str | None`)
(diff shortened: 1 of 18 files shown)


chrisjsewell:
> Problem with undefined unbound fields is they are problematic in filter strings.

Well, this is a problem with the current filter strings, if evaluated by Python. In jinja m, for example, there is a handling of undefined values. 
Undefined values are also a problem if, in the future, you wanted to allow different requirement types to have different fields, which I think is not unreasonable

ubmarco:
Nullable makes total sense for information not given. Agree.
All optional fields in SN should be nullable in DBs like sqlite. Things like `type`, `id` and `title` is always required and never `null`.
When doing schema checks, nullable comes down to not provided (will not be part of the model that gets validated).
So people can use `"unevaluatedProperties": false` to tell "I do not allow additional properties". The schema validator knows what to exclude by looking at the `null` state.

PR #1482, opened 2025-08-04
♻️ Lazily assess directive options

chrisjsewell:
docutils `Directive.option_spec` is designed to be a statically specified map of directive options names to validation/type coercion functions. 
In sphinx-needs, the `NeedDirective`, `NeedextendDirective`, and `NeedserviceDirective` directives have dynamic options, based on the schema of the need items.

Currently, the "hack" to get this to work is to dynamically set class variables on these classes. 
This is problematic because:

1. It modifies global state, which can be problematic for re-building in the same Python session.
2. It leads to confusing logic, spread across the code base.
3. It also means it is impossible in the future to have different need types with potentially different schemas.

This PR instead takes an approach from `sphinx/ext/autodoc/directive.py`, 
to allow all options to be collected, without specifying their names up-front,
then analysing them within the `Directive.run` method.

---

‼️ Breaking

This could be breaking, for any user that relied on particular existence/use of `Directive.option_spec` or `Directive.options`,
although this would be deemed quite an exotic use-case, and not one intended in the API.


sphinx_needs/directives/need.py
```diff
@@ -23,11 +22,13 @@
...
 )
 from sphinx_needs.layout import build_need_repr
-from sphinx_needs.logging import get_logger, log_warning
+from sphinx_needs.logging import WarningSubTypes, get_logger, log_warning
 from sphinx_needs.need_constraints import process_constraints
 from sphinx_needs.nodes import Need
 from sphinx_needs.utils import (
+    DummyOptionSpec,
     add_doc,
+    coersce_to_boolean,
```
ubmarco: There are more instances of this.
```suggestion
    coerce_to_boolean,
```

ubmarco (changes requested):
Implementation is really cool. The `DummyOptionSpec` took me some time to understand, it's part of the lazy evaluation to fulfill the docutils contract, right? Other than that, just one review finding.

chrisjsewell:
> Implementation is really cool. The `DummyOptionSpec` took me some time to understand, it's part of the lazy evaluation to fulfill the docutils contract, right? Other than that, just one review finding.

Yep, it acts like docutils "expects" it to, maybe a little "hacky" but it's been used in core sphinx for many years so won't break:

https://github.com/sphinx-doc/sphinx/blob/f92316a2f56ce66cc1b74a2939a5db723a1a2b71/sphinx/ext/autodoc/directive.py#L32

sphinx_needs/directives/need.py
chrisjsewell: fixed cheers

ubmarco (approved):
Good work!

PR #1485, opened 2025-08-06
♻️ Store needs as `NeedItem` / `NeedPartItem`, rather than standard `dict`

chrisjsewell:
Currently, need data items are stored as standard Python dicts, which is also exposed to the user in various APIs and as the `needs.json`.
This is problematic as it does not allow for

1. any internal checks to ensure data is stored/updated consistently (e.g. fields like `id` should never be changed), 
2. any data to be "hid" from the user, or presented in a different way to its internal structure.
    - because the dict is a flat key/val store, it does not allow for distinction between internal / user set extra/link fields
    - we would like to improve how values that are dynamic/variant functions are stored (prior to evaluation)
    - we would also like to store source mapping data from distinct sources (directives, external, imports) in a more coherent way.
4. deprecations of any fields in a non-breaking manner

In this PR, we instead add a specific `NeedItem` class, which closely mimics a`dict`, but also allows for additional constraints and functionality.

We also introduce a `NeedPartItem` which is distinctly different from a `NeedItem`, e.g. it should not be mutated.

To ensure that no existing user code is accidentally breaking this, e.g. by directly add dictionaries to the list of needs,
we perform a check, at the end of the post-processing stage, to make sure all need items are instances of `NeedItem`
(and point them towards the "correct" `add_need` API)

---

‼️ Breaking

This would be breaking for any users doing "non-API" modifications / additions to the needs data, i.e. directly adding dict items.
It should not change interactions with standard APIs like `add_need` or filter strings, etc

### DIFF
--- a/sphinx_needs/need_item.py
+++ b/sphinx_needs/need_item.py
@@ -1,0 +1,333 @@
+# To maintain back-compatibility,
+# we need to implement MutableMapping semantics, these are the methods that need to be implemented:
+# - implement immutable: __getitem__, __iter__, __len__
+# - implement mutable: __setitem__, __delitem__
+# - inherit immutable: __contains__, keys, items, values, get, __eq__, and __ne__
+# - inherit mutable: pop, popitem, clear, update, and setdefault
+#
+# For NeedPartItem, we should not allow any mutability, as it is a view into a need.
+# For NeedItem, we allow mutability, but only for values, i.e. it should not allow adding or removing keys.
+from __future__ import annotations
+
+from collections.abc import Iterable, Iterator, Mapping, Sequence
+from typing import Any, Literal, overload
+
+from sphinx_needs.data import NeedsInfoType, NeedsPartType
+
+
+class NeedItem:
+    """A class representing a single need item."""
+
+    __slots__ = ("_data",)
+
+    _not_required = {"pre_content", "post_content", "constraints_error"}
+    _immutable = {
+        "id",
+        "type",
+        "docname",
+        "lineno",
+        "lineno_content",
+        "id_complete",
+        "id_parent",
+        "is_need",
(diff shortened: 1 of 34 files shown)


sphinx_needs/need_item.py
```diff
@@ -0,0 +1,333 @@
...
+            "constraints_passed",
+        ],
+    ) -> bool: ...
+
+    @overload
+    def __getitem__(
+        self,
+        key: Literal[
+            "id",
+            "id_parent",
```
ubmarco, line 95: id_parent should be optional, it's about need part parent - would it break the data model to change it to `str | None` ?

sphinx_needs/need_item.py
```diff
@@ -0,0 +1,333 @@
...
+            "type_style",
+            "content",
+            "pre_content",
+            "post_content",
+            "doctype",
+            "external_url",
+            "external_css",
+            "constraints_error",
+            "section_name",
+            "parent_need",
```
ubmarco, line 110: same here, `parent_need` is optional, it's about need nesting. Wouldn't `str | None` be better?

ubmarco (approved):
This looks good to me and is a required step to get people to use the API or complain if something is missing. We should inform core users about this change.
I just have 2 comments, please decide whether those are issues right now or not.

sphinx_needs/need_item.py
chrisjsewell: This particular PR is not intended to change any schema specifics; e.g. it shows that this change does not change any output needs.json
So it may well be a legitimate change, but should not be made here

sphinx_needs/need_item.py
chrisjsewell: Same as above

PR #1489, opened 2025-08-08
🐛 Fix: disallow need variants for list type fields

chrisjsewell:
The variant processing was turning these into strings, rather than list of strings, which does not make sense and breaks the need item schema.

Since this is the case, and has not been noticed before, it is doubtful that variants are ever used in this context.

### DIFF
--- a/sphinx_needs/api/need.py
+++ b/sphinx_needs/api/need.py
@@ -58,7 +58,7 @@ def generate_need(
     constraints: None | str | list[str] = None,
     parts: dict[str, NeedsPartType] | None = None,
     arch: dict[str, str] | None = None,
-    signature: str = "",
+    signature: None | str = None,
     sections: list[str] | None = None,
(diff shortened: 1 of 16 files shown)


chrisjsewell:
> Would it cost much time to fix it instead of deactivating it?

The question is; what does this syntax even mean, in terms of a list of strings?

For example here; `:links: [tag]:some|thing, other|wise`, are `some|thing` / `other;wise` single link references,
or a list of references; `["some", "thing"]`, and how does these even integrate with the initial parsing of these strings.

The idea of variants is not necessarily bad, but the implementation is a complete ill-though out mess.

I will be disabling this for now,
but a next step is to move the identification/parsing of dynamic functions and variants to the initial need item creation phase (e.g. when a directive is parsed) and then we will better be able to implement/orchestrate the parsing of string lists / dynamic functions / variant functions and how they interact together

PR #1491, opened 2025-08-13
♻️ split off `source` fields in `NeedItem` internal data

chrisjsewell:
The source mapping data (docname, lineno, ...) is now stored in a separate field, which allows for more clarity over the source of the need, allows more source specific information to be stored, and has better type safety.

The users of these objects should see no change in its API

### DIFF
--- a/sphinx_needs/api/need.py
+++ b/sphinx_needs/api/need.py
@@ -408,11 +408,11 @@ def generate_need(
 
 def add_need(
     app: Sphinx,
-    state: None | RSTState,
-    docname: None | str,
-    lineno: None | int,
-    need_type: str,
-    title: str,
+    state: None | RSTState = None,
+    docname: None | str = None,
+    lineno: None | int = None,
+    need_type: str = "",
+    title: str = "",
     *,
     need_source: NeedItemSourceProtocol | None = None,
     id: str | None = None,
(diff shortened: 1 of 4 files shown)


sphinx_needs/api/need.py
```diff
@@ -390,12 +408,13 @@ def generate_need(
...
-    state: None | RSTState,
-    docname: None | str,
-    lineno: None | int,
-    need_type: str,
-    title: str,
+    state: None | RSTState = None,
+    docname: None | str = None,
+    lineno: None | int = None,
+    need_type: str = "",
+    title: str = "",
```
ubmarco, line 415: Why would the need_type have a default? Isn't that required primary information when using the API?
For the title I remember it can potentially be generated from other sources.

chrisjsewell: So this was only because the docname and lineno were already before these arguments, and I can't add defaults for these without adding defaults for the others. Otherwise I have to change the order and potentially break user code already using add_need as it currently is (if they are not using keyword arguments)

ubmarco: Ok got it. I see in [generate_need](https://github.com/useblocks/sphinx-needs/blob/7311f1634a3690c110300220f820d9c7b579c0fa/sphinx_needs/api/need.py#L148) (which is called by `add_need`), that the given need type is checked against the registered types and errors if it does not match, so I'm ok with that.

ubmarco (approved):
I like the new need source implementation. It's the better structure to know where information comes from.

PR #1504, opened 2025-08-29
🔧 Improve storage of constraint results on `NeedItem`

chrisjsewell:
The result should be `None` before processing, set once, then immutable after, and should be consistent with the list of constraints.

Note, this is part of trying to rationalise the internal fields of a need item: 

image


sphinx_needs/need_item.py
```diff
@@ -163,6 +208,7 @@ class NeedItem:
     _immutable_core = {
         "id",
         "type",
+        "jinja_content",
```
ubmarco: What's the relation to this PR?

chrisjsewell: nothing really, just part of this rationalisation of internal fields

chrisjsewell: can take out and put in next PR, that will also make immutable; content, pre_content, post_content, doctype

chrisjsewell:
```suggestion
```

ubmarco (approved):

PR #1506, opened 2025-08-30
🔧 Improve storage of content generation on `NeedItem`

chrisjsewell:
Add `NeedsContent` construct, to store all data to do with content generation: `doctype`, `content`, `pre_content`, `post_content`, `jinja_content`, `template`, `pre_template`, `post_template`

These are all immutable, after the content has been initially set

note, this is part of the rationalisation of internal need fields:

image

### DIFF
--- a/sphinx_needs/data.py
+++ b/sphinx_needs/data.py
@@ -280,40 +280,40 @@ NeedsCoreFields: Final[Mapping[str, CoreFieldParameters]] = {
         "exclude_import": True,
     },
     "jinja_content": {
-        "description": "Whether the content should be pre-processed by jinja.",
+        "description": "Whether the content was pre-processed by jinja.",
         "schema": {"type": "boolean", "default": False},
         "exclude_external": True,
     },
     "template": {
-        "description": "Template of the need.",
+        "description": "Whether the content was created from a jinja template.",
         "schema": {"type": ["string", "null"], "default": None},
         "exclude_external": True,
         "allow_default": "str",
     },
     "pre_template": {
-        "description": "Pre-template of the need.",
+        "description": "Whether the pre_content was created from a jinja template.",
         "schema": {"type": ["string", "null"], "default": None},
         "exclude_external": True,
         "allow_default": "str",
     },
     "post_template": {
-        "description": "Post-template of the need.",
+        "description": "Whether the post_content was created from a jinja template.",
         "schema": {"type": ["string", "null"], "default": None},
         "exclude_external": True,
         "allow_default": "str",
     },
     "content": {
-        "description": "Content of the need.",
+        "description": "The main content of the need.",
         "schema": {"type": "string", "default": ""},
     },
     "pre_content": {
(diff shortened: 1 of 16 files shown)


sphinx_needs/data.py
```diff
@@ -280,40 +280,40 @@ class CoreFieldParameters(TypedDict):
...
     },
     "jinja_content": {
-        "description": "Whether the content should be pre-processed by jinja.",
+        "description": "Whether the content was pre-processed by jinja.",
         "schema": {"type": "boolean", "default": False},
         "exclude_external": True,
     },
     "template": {
-        "description": "Template of the need.",
+        "description": "Whether the content was created from a jinja template.",
```
ubmarco: if that key actually holds the template I think the description should not start with `Whether` as it indicates a boolean like in `jinja_content`.

```suggestion
        "description": "Jinja template that was used for the need content.",
```

sphinx_needs/data.py
```diff
@@ -280,40 +280,40 @@ class CoreFieldParameters(TypedDict):
...
         "schema": {"type": ["string", "null"], "default": None},
         "exclude_external": True,
         "allow_default": "str",
     },
     "pre_template": {
-        "description": "Pre-template of the need.",
+        "description": "Whether the pre_content was created from a jinja template.",
```
ubmarco: same here

sphinx_needs/data.py
```diff
@@ -280,40 +280,40 @@ class CoreFieldParameters(TypedDict):
...
         "schema": {"type": ["string", "null"], "default": None},
         "exclude_external": True,
         "allow_default": "str",
     },
     "post_template": {
-        "description": "Post-template of the need.",
+        "description": "Whether the post_content was created from a jinja template.",
```
ubmarco: same here

sphinx_needs/data.py
```diff
@@ -338,7 +338,7 @@ class CoreFieldParameters(TypedDict):
         "allow_extend": True,
     },
     "constraints_results": {
-        "description": "Mapping of constraint name, to check name, to result.",
+        "description": "Mapping of constraint name, to check name, to result, None if not yet checked.",
```
ubmarco, line 341: a bit unclear how this maps. Is constraint name mapping to a tuple (check name, result) or is it a nested dict constraint name[check name] = result?

chrisjsewell: yes it is a nested dict, this is made clear by the actual typing of the field.
This is not part of this PR though, it is just a fix for a test, so I would not change that here

sphinx_needs/data.py
chrisjsewell: fixed

sphinx_needs/data.py
chrisjsewell: fixed

sphinx_needs/data.py
chrisjsewell: fixed

ubmarco (approved):

PR #1508, opened 2025-09-01
🔧 Fix extension available check for plantuml

AlexanderLanin:
It's important whether the extension is available or not, not whether it's in the config. This is relevant when extensions were added via `app.setup_extension`


ubmarco:
Closed in favor of https://github.com/useblocks/sphinx-needs/pull/1521.
Sorry @AlexanderLanin, credits go to you :)

PR #1509, opened 2025-09-01
♻️ Improve storage of part data on NeedItem

chrisjsewell:
Parts can now only have an `id` and `content` key, plus backlink fields.

When merging a part with a need, to create a `NeedPartItem` (used in analysis directives):

- The `id` and `content `are overridden
- All link types are set to empty (parts cannot link to anything)
- All backlinks are overridden with only direct links to that part (not to the parent need)
- All parts are removed (parts cannot contain parts)

Also, functions `check_links` and `create_back_links` have been merged into `update_back_links`


ubmarco (commented):
Thanks for the refactoring.
One question for the outgoing links to need parts. In the docs example

Image

the need `impl_my_car_1` only links to parts of `my_car_1`. However in the needtable below

Image

the need parent `my_car_1` is also linked by `impl_my_car_1`. Does this make sense? Are parents inheriting the links from the part children or is that just a representational thing in needtable? The needs.json looks like the child -> parent inheritance is part of the data structure, see line 3042:

Image

Another thing I spotted in this section of the docs

Image

The dotted line does not exist as stated in the text. This however does also not exist in the [current docs](https://sphinx-needs.readthedocs.io/en/latest/roles.html#need-part-np).

chrisjsewell:
> The dotted line does not exist as stated in the text. This however does also not exist in the current docs.

and yeh also here,if it is just mirroring what is already the behaviour, then I would leave it to another PR to fix

sphinx_needs/directives/need.py
```diff
@@ -460,82 +438,56 @@ def format_need_nodes(
...
-                        if (
-                            option_back
-                            not in needs[need_id_main]["parts"][need_id_part]
-                        ):
-                            needs[need_id_main]["parts"][need_id_part][option_back] = []  # type: ignore[literal-required]
-                        needs[need_id_main]["parts"][need_id_part][option_back].append(  # type: ignore[literal-required]
-                            key
-                        )
+                if linked_need := needs.get(need_id_main):
+                    linked_need.add_backlink(link_type, key)
```
chrisjsewell, line 453: Note, this is where it always creates a backlink to the parent, as well as the part (as per the current code)
This could easily be changed, if we feel it is correct to do so

Issue #1500, opened 2025-08-24
needimport: change ReferenceError to warning


PhilipPartsch:
Currently the implementation of need import raise an error in case the `needs.json` file is missing.
In bigger projects such needs.json files are generated by pipelines, and if a developer wants to execute a local build he would have to copy all such files to his local setup.
Often this is not really needed, as only the repo internal needs shall be changed.

A change to a warning, would give the developer a possibility to work with the `sphinx-build` without been blocked about the non-existence of the `needs.json`.

[Here](https://github.com/useblocks/sphinx-needs/blob/6f90e9908d3b6602f1b64034700cfbc3e24bb0e3/sphinx_needs/directives/needimport.py#L95) is the current exception, which we want to downgrade to a warning.

ubmarco:
Hi @PhilipPartsch,
this should be possible. Depending on the `-W` and `--keep-going` flags the same outcome as now can be done with warnings, too.

However, I'm not sure about the use case. Why can a need import be ignored and still get valid build output? Won't you have a lot of downstream issues like broken links? I mean there is a reason the import exists. Also one thing to consider: You will not be able to ignore single needimports, you can only disable the whole warning category. Isn't that dangerous to do a catch-all ignore?

PhilipPartsch:
The developers want to use ubcode for visiual studio code to make fastly changes on a multi repo setup.
Currently they would have to copy all generated needs.json files from the corresponding pipeline manually to thier local machines.
The request is to avoid this manual copy. My idea was to downgrade the exception to a warning, so developers still can make changes without being blocked. 
Do you have another idea?

danwos:
After a talk in the EaC round with Philip, this change is fine for me and makes sense.

ubmarco:
PR https://github.com/useblocks/sphinx-needs/pull/1510 merged.
@PhilipPartsch Thanks for the proposal and for the PR.

PR #1510, opened 2025-09-01
changed error to warning

PhilipPartsch:
try to implement #1500

### DIFF
--- a/sphinx_needs/directives/needimport.py
+++ b/sphinx_needs/directives/needimport.py
@@ -92,8 +92,11 @@ class NeedimportDirective(SphinxDirective):
             )[1]
 
             if not os.path.exists(correct_need_import_path):
-                raise ReferenceError(
-                    f"Could not load needs import file {correct_need_import_path}"
+                log_warning(
+                    logger,
+                    f"Could not load needs import file {correct_need_import_path}",
+                    "needimport",
+                    location=(self.env.docname, self.lineno),
                 )
 
             try:


sphinx_needs/directives/needimport.py
```diff
@@ -92,8 +92,11 @@ def run(self) -> Sequence[nodes.Node]:
             )[1]
 
             if not os.path.exists(correct_need_import_path):
-                raise ReferenceError(
-                    f"Could not load needs import file {correct_need_import_path}"
+                log_warning(
```
ubmarco, line 98: This will still raise a `FileNotFoundError` exception below (caught by `OSError`). There is also a TODO to add exception handling below. I think we should return a simple docutils paragraph node that warns the reader about the missing needimport. And then, as you do, emit a warning.

ubmarco (changes requested):

sphinx_needs/directives/needimport.py
```diff
@@ -92,9 +92,20 @@ def run(self) -> Sequence[nodes.Node]:
...
 
             if not os.path.exists(correct_need_import_path):
-                raise ReferenceError(
+                warning_text = (
                     f"Could not load needs import file {correct_need_import_path}"
                 )
+                log_warning(
+                    logger,
+                    warning_text,
+                    "needimport",
```
ubmarco, line 101: A different subtype would make sense for this case, so it can be ignored isolated.

```suggestion
                    "needimport_missing",
```

ubmarco (changes requested):
Thanks for the update.
I tested the change and the implementation looks good now. Only remark is the type of warning. It should be more distinct so that specific warning can be ignored isolated.

sphinx_needs/directives/needimport.py
ubmarco: I changed my mind, let's not overcomplicate this. If people want to ignore warnings around needimport, so be it. We can split if there is a usecase.

ubmarco (approved):

danwos (approved):

PR #1540, opened 2025-10-07
make jsonschema dependency optional

### LINKED ISSUE #1525: sphinx-needs 6.0 adds too many additional dependencies
```
dependencies = [
    "fqdn",
    "idna",
    "isoduration",
    "jsonpointer>1.13",
    "jsonschema==4.25.1",
    "rfc3339-validator",
    "rfc3987",
    "uri-template",
    "webcolors>=1.11",
]
```

These dependencies are being added via `jsonschema[format]` extra. It is unclear why these are required now and there is no way to opt-out.

sigma67:
closes #1525


sigma67:
idk about the docker tests, seems to be some kind of unrelated auth problem

> Error: Username and password required

pyproject.toml
```diff
@@ -40,9 +39,13 @@ dependencies = [
 
 [project.optional-dependencies]
 plotting = ["matplotlib>=3.3.0"] # for needpie / needbar
+schema = [
+  "jsonschema[format]>=3.2.0",
```
ubmarco, line 43: We should split this, e.g. into groups `schema` and `schema-formats`. If `jsonschema` is available, all validations except the [string patterns](https://sphinx-needs.readthedocs.io/en/latest/schema/index.html#string-type) are possible. `jsonschema` might already be installed in some environments by other libraries.
If `schema-formats` is not installed, it would remove these dependencies:
```
 + arrow==1.3.0
 + fqdn==1.5.1
 + isoduration==20.11.0
 + jsonpointer==3.0.0
 + rfc3339-validator==0.1.4
 + rfc3987==1.3.8
 + types-python-dateutil==2.9.0.20250822
 + uri-template==1.3.0
 + webcolors==24.11.1
```
`jsonschema` itself only requires:
```
 + attrs==25.4.0
 + jsonschema==4.25.1
 + jsonschema-specifications==2025.9.1
 + referencing==0.36.2
 + rpds-py==0.27.1
```

Edit: See below, I propose to only have `schema-formats` and leave `jsonschema` as a required dependency.

README.rst
```diff
@@ -54,6 +54,12 @@ Using pip
 
     pip install sphinx-needs
 
+To use schema validation features of sphinx-needs, you need to also install ``jsonschema``, which is available *via* the ``schema`` extra:
```
ubmarco, line 57:
```suggestion
To use schema validation features of sphinx-needs, you need to also install ``jsonschema``, which is available via the ``schema`` extra:
```

tests/no_jsonschema_tests.py
```diff
@@ -0,0 +1,13 @@
+"""These tests should only be run in an environment without sphinx_needs installed."""
```
ubmarco, line 1: Not sure how this test relates to what you are implementing. `sphinx_needs` is certainly installed :)

tests/no_jsonschema_tests.py
```diff
@@ -0,0 +1,13 @@
...
+import pytest
+
+
+@pytest.mark.parametrize(
+    "test_app",
+    [{"buildername": "html", "srcdir": "doc_test/doc_needsfile"}],
+    indirect=True,
+)
+def test_needsfile(test_app):
+    """Test the build fails correctly, if matplotlib is not installed."""
```
ubmarco, line 12: Typo?

sphinx_needs/utils.py
```diff
@@ -419,6 +420,19 @@ def import_matplotlib() -> matplotlib | None:
     return matplotlib
 
 
+@lru_cache
+def import_jsonschema() -> ModuleType | None:
+    """Import and return matplotlib, or return None if it cannot be imported.
```
ubmarco, line 425:
```suggestion
    """Import and return jsonschema, or return None if it cannot be imported.
```

sphinx_needs/utils.py
```diff
@@ -419,6 +420,19 @@ def import_matplotlib() -> matplotlib | None:
...
+
+    Also sets the interactive backend to ``Agg``, if ``DISPLAY`` is not set.
```
ubmarco, line 427: Copy paste error.

pyproject.toml
```diff
@@ -30,7 +30,6 @@ dependencies = [
   "sphinx>=7.4,<9",
   "requests-file~=2.1",             # external links
   "requests~=2.32",                 # external links
```
ubmarco, line 32: `jsonschema` was a dependency before for version `5.1.0` as we see in this PR as it was used to check the schema of imported needs.json files. I think that will still be required. This is a basic functionality. So I would only add a group `schema-formats` that only cares for the additional formats dependencies.

pyproject.toml
sigma67: ok yeah I agree, most of the trouble comes from the [format] extra

tests/no_jsonschema_tests.py
sigma67: see the ci.yaml changes, this follows the same structure as the no_matplotlib tests.

The relevant ci job uninstalls the dependency jsonschema and runs only this file

pyproject.toml
sigma67: valid, will fix this

pyproject.toml
sigma67: Could you clarify what changes are required as a result of this? How can we skip only the string patterns during validation?

tests/no_jsonschema_tests.py
ubmarco: I would expect a positive test that checks whether the build without formats still work.
Then a negative test that checks whether the warning is triggered that certain packages are missing.

sigma67:
> One more note, the check whether `schema-formats` is required or not should be done by looking at needs_extra_options.schema and the schema definition rules early during config loading. For needs_extra_options that is fairly easy, for the schemas it's some tree walking as in `def resolve_refs()` or `def populate_field_type()`.

That sounds a bit beyond my experience with this project. Feel free to take over that part if you have time

PR #1546, opened 2025-10-10
🔧 Derive core `FieldSchema` from `NeedsCoreFields`

chrisjsewell:
To ensure these are kept in sync, rather than duplicating data for both


sphinx_needs/api/need.py
ubmarco: I just looked this up because you changed to conversion to tuple.
I'm not sure what to pass here. I think `constraints` refer to the top level keys in https://sphinx-needs.readthedocs.io/en/latest/configuration.html#needs-constraints. According to above type it can either be `str` or `list[str]`, so we need to adapt the docs here too:
```suggestion
    :param constraints: Constraint names as comma separated string or list of strings
```

sphinx_needs/data.py
```diff
@@ -64,8 +64,9 @@ class CoreFieldParameters(TypedDict):
     """Description of the field."""
     schema: dict[str, Any]
     """JSON schema for the field."""
-    allow_default: NotRequired[Literal["str", "str_list", "bool"]]
-    """Whether the field allows custom default values to be set, and their type,
+    add_to_field_schema: NotRequired[bool]
```
ubmarco, line 67: Add a docstring

PR #1547, opened 2025-10-10
✨ Introduce `needs_core_options`, deprecate `needs_variant_options`

chrisjsewell:
The goal of this PR is to enable co-location of all field properties.

Co-location makes it easier for users to understand, and potentially allows for reuse of fields across projects.

Previously this was not possible, for field properties that could be applied to core fields, such as `needs_variant_options`.

This changes the config from e.g.

```python
needs_extra_options = [
     "extra"
]
needs_extra_links = [
     {"option": "links"}
]
needs_variant_options = ["status", "extra", "links"]
```

to

```python
needs_core_options = {
    "status": {"variant_functions": True, "dynamic_functions": True}
}
needs_extra_options = [
    {"name": "author", "variant_functions": True, "dynamic_functions": True}
]
needs_extra_links = [
     {"option": "links", "variant_functions": True, "dynamic_functions": True}
]
```

Note also, this allows for the previously disallowed variants on links, and allows for equivalently allowing/disallowing dynamic functions

In follow-up PRs, it would be the goal to also deprecate `needs_global_options`, `needs_statuses` and `needs_tags`
(i.e. co-locate defaults and enums)

---

TODO documentation, more tests

### DIFF
--- a/sphinx_needs/config.py
+++ b/sphinx_needs/config.py
@@ -39,6 +39,10 @@ class ExtraOptionParams:
 
     description: str
     """A description of the option."""
+    dynamic_functions: bool | None
+    """If True, allow dynamic functions when defining this field."""
+    variant_functions: bool | None
+    """If True, allow variant functions when defining this field."""
     schema: (
         ExtraOptionStringSchemaType
         | ExtraOptionBooleanSchemaType
@@ -106,6 +110,8 @@ class _Config:
         name: str,
         description: str,
         *,
+        dynamic_functions: bool | None = None,
+        variant_functions: bool | None = None,
         schema: ExtraOptionStringSchemaType
         | ExtraOptionBooleanSchemaType
         | ExtraOptionIntegerSchemaType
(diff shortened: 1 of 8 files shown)


danwos:
I don't like `needs_fields` in the Sphinx context, as Sphinx itself is talking about directive options and "field" is used in [field lists](https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html).

Our docs also talk about options everywhere. So `needs_fields` could become quite confusing.

I would stay with `needs_options`.

ubmarco:
While I don't have a strong opinion on `needs_fields` vs `needs_options` I always thought directives options *are* field lists actually. Or they are implemented as such.
@chrisjsewell sees the name `field` interchangeably to property, just like a class can have fields.
@danwos also has a point because the term `option` is well known in SN over the years.

danwos:
I guess we were talking a little bit about different things.
This PR started with introducing `needs_core_options`.
I voted for merging `needs_core_options` with `needs_extra_options` to `needs_options`.

Then Marco enlarged the scope to cover also `needs_extra_links` under the same config-var.
I was not aware that the scope has changed in the discussion, and therefore still voted for `need_options` as config var for just options.

`needs_fields` makes sense, if it covers also links.
However, I'm not sure because such a list can become quite huge and hard to read.
We are talking about a setup with >40 options/links, each with easily 4 to 10 lines of config.
So up to 160-400 lines for a single config var. And that in toml is even more worst, at least for me.

Also Options and Links have different config-vars.
And users normally want to have a separated list, to easily see and configure all links in one place.

So I vote, as earlier, for `needs_options` for options. And maybe `needs_links` for links, if we need to merge core and extra links as well, and want to get the name in sync with `needs_options`. No almighty, common list.

So in the end I'm happy with two solutions:
* Takes this PR as it is (40% happy, as it brings us forward)
* Merge core and extra to `needs_options` (100% happy)

PR #1548, opened 2025-10-10
👌 Make `incoming`/`outgoing` keys optional in `needs_extra_links`

chrisjsewell:
Previously specifying links as below would eventually result in a `KeyError` later during the build.

```python
needs_extra_links = [{"option": "links"}]
```

Now, default values are created for missing `incoming` and `outgoing` keys, making configuration easier.

### DIFF
--- a/sphinx_needs/needs.py
+++ b/sphinx_needs/needs.py
@@ -677,7 +677,7 @@ def merge_default_configs(_app: Sphinx, config: Config) -> None:
 
     for link in needs_config.extra_links:
         if "outgoing" not in link:
-            link["outgoing"] = link['option']
+            link["outgoing"] = link["option"]
         if "incoming" not in link:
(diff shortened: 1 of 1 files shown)


sphinx_needs/needs.py
```diff
@@ -675,6 +675,12 @@ def merge_default_configs(_app: Sphinx, config: Config) -> None:
 
     needs_config.extra_links = common_links + needs_config.extra_links
 
+    for link in needs_config.extra_links:
+        if "outgoing" not in link:
+            link["outgoing"] = f"{link['option']} outgoing"
```
ubmarco: The link options are often defined in a way that can be used right away for outgoing links, e.g. `depends` or `links` or `derives`.
```suggestion
            link["outgoing"] = link['option']
```

PR #1557, opened 2025-10-20
🔧 Fix `check_needs_data` not checking remote files in `needimport`

zlatan-g:
After discussing with @chrisjsewell this PR was created to use `check_needs_data` to also validate remote jsons
(was previously indented and was only happening for the else case)

### DIFF
--- a/sphinx_needs/directives/needimport.py
+++ b/sphinx_needs/directives/needimport.py
@@ -113,14 +113,14 @@ class NeedimportDirective(SphinxDirective):
             except (OSError, json.JSONDecodeError) as e:
                 # TODO: Add exception handling
                 raise SphinxNeedsFileException(correct_need_import_path) from e
-
-            errors = check_needs_data(needs_import_list)
-            if errors.schema:
-                logger.info(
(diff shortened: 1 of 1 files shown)


chrisjsewell (approved):
great work! Indeed it looks like a mistake

sphinx_needs/directives/needimport.py
```diff
@@ -113,14 +113,13 @@ def run(self) -> Sequence[nodes.Node]:
...
-            if errors.schema:
-                logger.info(
-                    f"Schema validation errors detected in file {correct_need_import_path}:"
-                )
-                for error in errors.schema:
-                    logger.info(f"  {error.message} -> {'.'.join(error.path)}")
+        errors = check_needs_data(needs_import_list)
+        if errors.schema:
+            logger.info(
+                f"Schema validation errors detected in file {correct_need_import_path}:"
```
chrisjsewell, line 119: ah actually `correct_need_import_path` is only available from the local file path if/esle branch, so this needs to now account for also pointing to the remote path

chrisjsewell (changes requested):

Issue #1553, opened 2025-10-15
Setting per schema 'severity' not working


MrChocolateMoose:
# Bug Report
I have a `schemas.json` that has various schemas in it. I am not quite ready to make some of them violations (which I would then escalate to errors). I would like to keep those particular ones as "informational" for now.

In the document, it makes it seem like that you can accomplish this:
https://sphinx-needs.readthedocs.io/en/latest/schema/index.html#id2

However, I am not able to achieve that result

# Expected Results
I have several schema items with severity "violation" and several with severity "info".

`Simple Stripped Down Example`
```json
{
    "schemas": [
        {
            "id": "field-1-enum",
            "severity": "violation",
            "message": "field must be one of A, B, C",
            "select": {
                "required": [
                    "my_field_1"
                ],
                "properties": {
                    "type": {
                        "const": "my_type"
                    }
                }
            },
            "validate": {
                "local": {
                    "required": [
                        "my_field_1"
                    ],
                    "properties": {
                        "my_field_1": {
                            "type": "string",
                            "enum": [
                                "A",
                                "B",
                                "C",
                            ]
                        }
                    }
                }
            }
        },
        {
            "id": "field-2-enum",
            "severity": "info",
            "message": "field must be one of D, E, F",
            "select": {
                "required": [
                    "my_field_2"
                ],
                "properties": {
                    "type": {
                        "const": "my_type"
                    }
                }
            },
            "validate": {
                "local": {
                    "required": [
                        "my_field_2"
                    ],
                    "properties": {
                        "my_field_2": {
                            "type": "string",
                            "enum": [
                                "D",
                                "E",
                                "F",
                            ]
                        }
                    }
                }
            }
        },
```

I expect if I violate one of type Info that I see it displayed as an informational item in the build process, otherwise as a violation (which would be categorized as a warning?).

# Actual Results
When I violate informational items, it is showing that the severity is "violation" (Default).

I have found only having "info" also fails. It's like it isn't applying at all.

# Library Details
I am on sphinx-needs 6.0.1

ubmarco:
Please try the latest master commit, I pushed a fix for another bug report that might help you.
Just have not released that yet.

ubmarco:
Reported issue https://github.com/useblocks/sphinx-needs/issues/1513 and PR https://github.com/useblocks/sphinx-needs/pull/1535 that I mentioned.
Please check whether your report is a duplicate / the PR fixes your issue.

[...]

PR #1561, opened 2025-10-22
👌 Improve schema severity handling

ubmarco:
Various schema reporting improvements:
- Remove setting `needs_schema_severity`
- Output *all* violations to console unless suppressed by `suppress_warnings`
- Introduce severity based log `type`, so they can specifically be suppressed
  - info: `sn_schema_info`
  - warning: `sn_schema_warning`
  - violation: `sn_schema_violation`
- Severity `info` and `warning` is logged as `logger.warning()`
- Severity `violation` is logged as `logger.error()` for a different color
- Changes to schema_violations.json
  - Write *all* violations, irrespective of the severity and `suppress_warnings`
  - The `log_lvl` information reflect what Sphinx does.
     Possible values are `warning` or `error`.
   - Do not output `log_lvl` and `type` for children.
     This information is only required on the root, the nested fields are duplicates.
   - Do not print need IDs that have no warnings
- Do not print `Field:` in schema console reports if not set
- Update docs and tests accordingly

Fixes: https://github.com/useblocks/sphinx-needs/issues/1553

### DIFF
--- a/docs/schema/index.rst
+++ b/docs/schema/index.rst
@@ -723,44 +723,57 @@ error, even though ``priority`` is in the ``required`` list.
 Rule severity
 ~~~~~~~~~~~~~
 
-Each schema rule can specify one of the following severity levels:
+Each schema rule can specify one of the severity levels ``info``, ``warning``, or
+``violation`` as described in :ref:`severity_levels`. If not given, ``violation`` is used as
+a default.
 
-- ``info``: A non-critical constraint violation indicating an informative message.
-- ``warning``: A non-critical constraint violation indicating a warning.
-- ``violation`` : A constraint violation.
+The severity influences how validation failures are displayed in the console.
 
-  If that is not given, a violation
-
-rule specific default kicks in. The severity influences how validation failures are displayed
-in the console:
-
-.. code-block:: python
+.. code-block:: json
 
    {
      "severity": "warning",
-     "message": "Approval required due to high efforts"
+     "select": {
+       "properties": { "type": { "const": "feat" } }
+     },
+     "validate": {
+       "local": {
+         "properties": { "id": { "pattern": "^FEAT_[a-zA-Z0-9_-]*$" } }
+       }
+     }
+   },
+   {
+     "select": {
+       "properties": { "type": { "const": "spec" } }
+     },
+     "validate": {
+       "local": {
+         "properties": { "id": { "pattern": "^SPEC_[a-zA-Z0-9_-]*$" } }
+       }
+     }
    }
 
-All severities are reported during schema validation.
-
-**Console output example:**
+**Console output for above example:**
 
 .. code-block:: text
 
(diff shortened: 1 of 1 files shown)


chrisjsewell (approved):
From the description, that all looks fine, but unfortunately I don't have time to go through the code in detail

docs/schema/index.rst
```diff
@@ -699,24 +718,49 @@ in via ``allOf`` to be considered evaluated:
...
-- ``warning``: Warning message
-- ``info``: Informational message
+Each schema rule can specify one of the following severity levels:
 
-.. code-block:: json
+- ``info``: A non-critical constraint violation indicating an informative message.
+- ``warning``: A non-critical constraint violation indicating a warning.
+- ``violation`` : A constraint violation.
+
+  If that is not given, a violation
```
MrChocolateMoose: Image
bad indention/spacing

docs/schema/index.rst
```diff
@@ -699,24 +718,49 @@ in via ``allOf`` to be considered evaluated:
...
+
+.. code-block:: text
+
+   WARNING: Need 'FEAt' has validation warnings:
+     Severity:       warning
+     Field:          id
+     Need path:      FEAt
+     Schema path:    [0] > local > properties > id > pattern
+     User message:   id must be uppercase with numbers and underscores
+     Schema message: 'FEAt' does not match '^[A-Z0-9_]+$' [sn_schema_warning.local_fail]
```
MrChocolateMoose: is 't' supposed to be lowercase in this example?

docs/schema/index.rst
```diff
@@ -986,12 +1030,76 @@ Array fields store lists of homogeneous typed values:
...
+~~~~~~~~~~~~~~~~~
+
+The :ref:`severity_levels` affect how Sphinx emits validation messages.
+Here is how they map to Sphinx logging levels and console output.
+
+Schema severity **info**:
+
+- Mapped to Sphinx logging level ``warning``
+- Displayed as ``WARNING:`` in red color in console
+- Logged with type ``sn_schema_info``
```
MrChocolateMoose, line 1120: ~In order to treat these messages as purely information, then I would supress `sn_schema_info` ?~

MrChocolateMoose (approved):
Approved, but there is some minor typos ~and a clarification I had~ that would be helpful to handle prior to merging.

docs/schema/index.rst
MrChocolateMoose: I reread the proposal, and that was the case. resolving this comment.

docs/schema/index.rst
ubmarco: I changed it to `FEAT` to avoid confusing, as this also fails the regex which was my initial goal.

docs/schema/index.rst
ubmarco: I rewrote that section, thanks for the finding.

PR #1564, opened 2025-10-29
🐛 functions: Fix broken parent-child node relationships

tim-nordell-nimbelink:
The find_and_replace_node_content function breaks parent-child node relationships since it does not go through the docutils tools for assigning node children. Utilize one of the built-in docutils methods [1] for this since it correctly assigns the parent-child relationship of each node.

Notably, the sphinxcontrib-spelling [2] extension is broken by this broken parent-child relationship since if something is misspelled it tries to identify which source line was impacted. When a node is added as a child, setup_child(...) [3] code fixes up the source/line location with the parent's information if needed.

A test has been added at the top-level which covers every single test run by adding a build hook into the "doctree-resolved" hook. This test is fairly simple - check each node (outside of the top node) and validate that it has a parent assigned.

[1] https://github.com/docutils/docutils/blob/4e912fe000b1b6dc1466c0ad1c3d1787d40fd96d/docutils/docutils/nodes.py#L788-L794
[2] https://pypi.org/project/sphinxcontrib-spelling/
[3] https://github.com/docutils/docutils/blob/4e912fe000b1b6dc1466c0ad1c3d1787d40fd96d/docutils/docutils/nodes.py#L150-L157

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -56,6 +56,10 @@ Bug fixes
   of need load order (e.g. when using :ref:`needs_external_needs`)
   (:issue:`1371`)
 
+- 🐛 Fix parent-child relationship of newly created nodes for needs. This
+  fixes interoperability with Sphinx extensions that look up source lines,
+  like sphinxcontrib-spelling (:issue:`1564`)
+
 .. _`release:8.1.1`:
 
 8.1.1
--- a/sphinx_needs/functions/functions.py
+++ b/sphinx_needs/functions/functions.py
@@ -263,7 +263,10 @@ def find_and_replace_node_content(
             for child in node.children:
                 new_child = find_and_replace_node_content(child, env, need)
                 new_children.append(new_child)
-                node.children = new_children
+
+            node.children = new_children
+            for subchild in node.children:
+                node.setup_child(subchild)
         else:
             node = nodes.Text(new_text)
         return node
@@ -275,7 +278,10 @@ def find_and_replace_node_content(
                 continue
             new_child = find_and_replace_node_content(child, env, need)
             new_children.append(new_child)
+
         node.children = new_children
+        for subchild in node.children:
+            node.setup_child(subchild)
     return node
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -255,6 +255,13 @@ def sphinx_test_tempdir(request) -> Path:
     return sphinx_test_tempdir
 
 
+def _check_parent_child(app: Sphinx, doctree: document, docname: str):
+    for idx, node in enumerate(doctree.findall()):
+        if idx == 0:
+            continue
+        assert node.parent is not None
+
+
 @pytest.fixture(scope="function")
 def test_app(make_app, sphinx_test_tempdir, request):
     """
@@ -332,6 +339,12 @@ def test_app(make_app, sphinx_test_tempdir, request):
     # Since we've bound the ``test_js`` function to the Sphinx object using ``__get__``,
     # ``test_js`` behaves like a method.
     app.test_js = test_js.__get__(app, Sphinx)
+
+    # Check created all parent-child node relationships after any other
+    # code within the doctree-resolved event by setting the priority to 999.
+    # Placing this test here provides coverage for quite a few tests in
+    # the suite, and hopefully will test any future features.
+    app.connect("doctree-resolved", _check_parent_child, priority=999)
 
     yield app
 


tim-nordell-nimbelink:
@danwos

chrisjsewell (commented):
## Review

Thanks for this — the root-cause diagnosis is spot on. `find_and_replace_node_content` rebuilds `node.children` by direct list assignment, which never wires up `child.parent`, leaving orphaned nodes. Routing through `setup_child()` is the right fix, and the bug it unblocks (sphinxcontrib-spelling walking back to the source line) is a great concrete motivation. CI is green across the full matrix. A few points below — none are blockers.

### `functions.py` — indentation of the new loop (please fix)

In the `nodes.reference` branch, both `node.children = new_children` and the new `setup_child` loop sit **inside** the `for child in node.children` loop:

```python
for child in node.children:
    new_child = find_and_replace_node_content(child, env, need)
    new_children.append(new_child)
    node.children = new_children
    for subchild in node.children:
        node.setup_child(subchild)
```

So for N children this re-assigns `node.children` and re-runs `setup_child` over the growing partial list on every iteration — O(N²) and re-parenting already-parented nodes repeatedly. The end state is correct, but it should run once after the loop. The pre-existing `node.children = new_children` was already mis-indented here; worth dedenting both:

```python
for child in node.children:
    new_child = find_and_replace_node_content(child, env, need)
    new_children.append(new_child)
node.children = new_children
for subchild in node.children:
    node.setup_child(subchild)
```

The `else` branch (the second hunk) already has this structure and looks correct. The two branches now share an identical `node.children = ...` + `setup_child` tail — minor, but a small helper or restructure could de-duplicate it.

### `tests/conftest.py`

- **`assert node.parent` → `assert node.parent is not None`.** docutils `Element` defines `__len__`, so an empty element is falsy. It happens to work here (a node's parent always contains at least that node, so `len >= 1`), but the assertion is really testing "is a parent set", and truthiness is a fragile/misleading way to express that.

- **`findall(include_self=False)` removes the `idx`/`continue` dance.** The `if idx == 0: continue` exists only to skip the document root; `doctree.findall(include_self=False)` does that directly:
  ```python
  def _check_parent_child(app, doctree, docname):
      for node in doctree.findall(include_self=False):
          assert node.parent is not None
  ```

- **Function name `test_check_parent_child`.** pytest doesn't collect from `conftest.py`, so it won't be mis-collected, but the `test_` prefix reads like a test case rather than an event handler. Consider `_check_parent_child` to avoid confusion.

- This wires a global invariant into **every** test via `test_app` — nice coverage, and `priority=999` correctly runs it after sphinx-needs' own `doctree-resolved` handlers. Worth a one-line comment explaining the intent (and the priority), since any future transform that legitimately leaves a detached node will now trip every test at once.

### Changelog

The `Unreleased` section of `docs/changelog.rst` is updated per-PR — please add a Bugfix entry referencing `:pr:`1564`` (mentioning the parent/child relationship fix and the sphinxcontrib-spelling interaction).

Overall this is a solid, well-justified fix — happy to approve once the `functions.py` indentation is tightened up.

---

ubmarco:
I would appreciate a test.

tim-nordell-nimbelink:
> I would appreciate a test.

I did place a test in there. It's run against any test that uses the test fixture `test_app` - that test fixture is used to setup the test environment.

tim-nordell-nimbelink:
> ### `functions.py` — indentation of the new loop (please fix)

Done.

> The `else` branch (the second hunk) already has this structure and looks correct. The two branches now share an identical `node.children = ...` + `setup_child` tail — minor, but a small helper or restructure could de-duplicate it.

I'm leaving it as is.

> ### `tests/conftest.py`
> 
>     * **`assert node.parent` → `assert node.parent is not None`.** docutils `Element` defines `__len__`, so an empty element is falsy. It happens to work here (a node's parent always contains at least that node, so `len >= 1`), but the assertion is really testing "is a parent set", and truthiness is a fragile/misleading way to express that.

Done.

>     * **Function name `test_check_parent_child`.** pytest doesn't collect from `conftest.py`, so it won't be mis-collected, but the `test_` prefix reads like a test case rather than an event handler. Consider `_check_parent_child` to avoid confusion.

Done.

>     * This wires a global invariant into **every** test via `test_app` — nice coverage, and `priority=999` correctly runs it after sphinx-needs' own `doctree-resolved` handlers. Worth a one-line comment explaining the intent (and the priority), since any future transform that legitimately leaves a detached node will now trip every test at once.

Done.

> ### Changelog
> 
> The `Unreleased` section of `docs/changelog.rst` is updated per-PR — please add a Bugfix entry referencing `:pr:`1564`` (mentioning the parent/child relationship fix and the sphinxcontrib-spelling interaction).

Done.

> _Generated by [Claude Code](https://claude.ai/code)_

;)

tim-nordell-nimbelink:
I updated the commit message to have the emoji + PR # referenced in the subject line as well.

ubmarco:
The failing docs-linkcheck is fixed by https://github.com/useblocks/sphinx-needs/pull/1732. Let's merge it tomorrow, then rebase this PR.

chrisjsewell (approved):
Thanks for the quick turnaround — all my review points are addressed:

- ✅ `functions.py`: the `node.children = ...` assignment and `setup_child` loop are now dedented out of the `for child` loop in both branches (fixes the O(N²) re-parenting)
- ✅ `assert node.parent is not None`
- ✅ renamed to `_check_parent_child`
- ✅ comment added explaining the global hook and the `priority=999` ordering
- ✅ changelog entry added

Fine with leaving the two branches un-deduplicated.

LGTM 👍

One trivial, non-blocking nit: the changelog uses `:issue:`1564``, but 1564 is the PR number, so `:pr:`1564`` is the more accurate role (GitHub redirects either way, so no need to hold the PR on it).

The red `docs-linkcheck` is unrelated to this change — per @ubmarco it's fixed by #1732 and this PR will be rebased once that lands.

PR #1571, opened 2025-11-11
🐛 needs.json for nullable fields

ubmarco:
Nullable fields should reflect that property in the needs.json schemas section.


chrisjsewell (approved):
Yeh you'll need to update all the snapshots to get the tests passing, but the logic is correct cheers

PR #1577, opened 2025-11-25
🧪 Fix schema benchmark tests

ubmarco:
Benchmark tests were broken and this went undetected.
Benchmark should run in a small form on each PR to make sure they are not broken.


chrisjsewell:
Thanks @ubmarco, FYI this isn't actually even a benchmark at the moment lol, because it tries to use `@pytest.mark.benchmark`, which isn't a thing, rather than the `benchmark` fixture: https://pytest-benchmark.readthedocs.io/en/latest/usage.html
But that's ok I can fix that; splitting it into a test (for a small run) and benchmarks (for larger runs)

PR #1596, opened 2025-12-12
👌 Validate all schemas compile during config resolution

chrisjsewell:
During resolution of schemas, coming from config `needs_extra_links`, `needs_extra_options` and `needs_schema_definitions`,
check that these schemas actually compile.
This ensures that the later schema validations do not fail due to this, and also removes the need for "bespoke" regex checking; in-turn this actually allows more regex syntaxes than previously (but still not lookahead/lookbehind).

Additionally, do not use the `typeguard` dependency, to check schemas coming from `need_extra_options` and `need_extra_links` and instead.
The replacement checks are "simpler", but we need to remove `typeguard`, because it is not compatible with python 3.14, and I believe these simpler checks, together with the schema validation, will suffice.
Note, validation of `SchemasRootType` is still using `typeguard` for now, so we can not remove the dependency, but this can be done in a follow-up PR.

### DIFF
--- a/tests/schema/__snapshots__/test_schema.ambr
+++ b/tests/schema/__snapshots__/test_schema.ambr
@@ -588,8 +588,13 @@
 # ---
 # name: test_schema_config[schemas_validate_pattern_atomic_groups_error]
   '''
-  Schemas entry '[0]' is not valid:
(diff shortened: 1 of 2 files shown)


ubmarco (approved):
This simplifies the logic a lot. Good work!

Issue #1566, opened 2025-11-04
sphinx-needs 6.1.0 is incompatible with python 3.14


PhilipPartsch:
it looks like Sphinx-Needs 6.1.0 is incompatible with Python 3.14:
build with [python 3.13.7](https://github.com/PhilipPartsch/playground/actions/runs/19064137171/job/54450487761) is working

build with [python 3.14.0](https://github.com/PhilipPartsch/playground/actions/runs/19063645699) is not working

ubmarco:
Good hint. We would need a PR on Sphinx-Needs side to let our CI evaluate this.

PR #1597, opened 2025-12-12
♻️ Remove typeguard dependency

chrisjsewell:
The `typeguard` dependency is a blocker to supporting Python 3.14 [#1566](https://github.com/useblocks/sphinx-needs/issues/1566#issuecomment-3626096968)
Currently, its only use is to validate configuration for need schemas

In this PR, we replace this functionality with validating the configuration against JSON schemas, using the existing `jsonschema-rs` dependency.
We generate these schemas via converting the `TypedDict` configuration to `pydantic` models first.
Note that we only use `pydantic` as a test dependency, to auto-generate the schemas and ensure they are up-to-date, rather than as a runtime dependency, since we don't want to introduce an additional dependency (with version constraints) that would make it more difficult for sphinx-needs to integrate into people's existing Python environments.

### DIFF
--- a/sphinx_needs/schema/config_utils.py
+++ b/sphinx_needs/schema/config_utils.py
@@ -96,6 +96,10 @@ def validate_schemas_config(app: Sphinx, needs_config: NeedsSphinxConfig) -> Non
 
     # set idx for logging purposes, it's part of the schema name
     for idx, schema in enumerate(needs_config.schema_definitions["schemas"]):
+        if not isinstance(schema, dict):
+            raise NeedsConfigException(
+                f"Schema entry at index {idx} in needs_schema_definitions.schemas is not a dict."
+            )
         schema["idx"] = idx
 
     # check severity and inject default if not set


sphinx_needs/schema/config.py
```diff
@@ -438,7 +340,7 @@ class NeedFieldsSchemaType(AllOfSchemaType):
     Intented to validate multiple fields on a need type.
     """
 
-    type: Literal["object"]
+    type: NotRequired[Literal["object"]]
```
chrisjsewell, line 343: @ubmarco I believe this uncovered a bug here that `typeguard` did not pick up on; since `{"allOf": [...]}` is a valid `NeedFieldsSchemaType`, due to its inheritance of `AllOfSchemaType`, then `type` cannot be required

sphinx_needs/schema/config_utils.py
ubmarco, line 82: Please also check whether all items are dictionaries.
This `"schemas": [1, 2],` leads to 
```
      File "/home/marco/git/useblocks/sphinx-needs/sphinx_needs/schema/config_utils.py", line 99, in validate_schemas_config
        schema["idx"] = idx
        ~~~~~~^^^^^^^
    TypeError: 'int' object does not support item assignment
```

sphinx_needs/schema/config.py
ubmarco: Actually having `type` as a sibling to `allOf` is not a problem, because users can validate a given schema, but independently also the `allOf` key. It just does nothing if it has no other keys such as `properties` or `required`.

sphinx_needs/schema/config_utils.py
chrisjsewell: added 👍

sphinx_needs/schema/config.py
chrisjsewell: as discussed, currently there are existing tests that only contain `allOf` and thus fail validation failed.
We can change in back to required in a later PR, if we feel this is necessary, and also change the tests.
This would be an "edge case" breaking change for users though

PR #1605, opened 2025-12-17
♻️ Move `needs_statuses` and `need_tags` checking to schema validation

chrisjsewell:
This PR refactors `needs_statuses` and `needs_tags` checking, from explicit validation logic (at need creation time) to the new generic schema-based validation system. 
The changes rename `extra_option` schema validation to `option` validation (covering both core and extra options) and automatically generate schema constraints from the `needs_statuses` and `needs_tags` configuration, for example:

- `needs_statuses = ["a", "b"]` to `{"status": {"type": "string", "enum": ["a", "b"]}}` and 
- `needs_tags = ["a", "b"]` to `{"tags":  {"type": array", "items": {"type": "string", "enum": ["a", "b"]}}}` 

This fixes issues with these being evaluated too early and not accounting for dynamic functions, needextend, etc ...
and will also facilitate #1547

### DIFF
--- a/tests/test_broken_docs.py
+++ b/tests/test_broken_docs.py
@@ -77,7 +77,12 @@ def test_broken_links(test_app: SphinxTestApp):
 def test_broken_statuses(test_app: SphinxTestApp):
     test_app.build()
     assert get_warnings(test_app) == [
-        "<srcdir>/index.rst:11: WARNING: Need could not be created: Status 'NOT_ALLOWED' not in 'needs_statuses'. [needs.create_need]"
+        "ERROR: Need 'SP_TOO_002' has schema violations:",
+        "  Severity:       violation",
+        "  Field:          status",
+        "  Need path:      SP_TOO_002",
+        "  Schema path:    options > schema > properties > status > enum",
+        '  Schema message: "NOT_ALLOWED" is not one of "open" or "implemented" [sn_schema_violation.option_fail]',
     ]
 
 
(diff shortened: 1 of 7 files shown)


ubmarco (approved):
LGTM. One comment, and I also see a Copilot message.

tests/test_broken_docs.py
```diff
@@ -77,7 +77,12 @@ def test_broken_links(test_app: SphinxTestApp):
 def test_broken_statuses(test_app: SphinxTestApp):
     test_app.build()
     assert get_warnings(test_app) == [
-        "<srcdir>/index.rst:11: WARNING: Need could not be created: Status 'NOT_ALLOWED' not in 'needs_statuses'. [needs.create_need]"
+        "ERROR: Need 'SP_TOO_002' has schema violations:",
```
ubmarco, line 80: This makes SN more resilient, but the error also appears later in the build process, right?

chrisjsewell: It does appear later yes, but that it actually more "correct", since otherwise the validation is wrong, if e.g. the status is a dynamic function, etc.
It is of note that these warnings no longer include "source mapping" to the original need, this may be a consideration for all of these schema warnings?

ubmarco: Yes, that should totally be there, I thought about this before.
We know the affected need ID, so should not be too hard.

chrisjsewell: okie, well if you want to maybe open an issue for that, or do the PR yourself, but otherwise I just keep it in my internal todo list lol

ubmarco: Here we go https://github.com/useblocks/sphinx-needs/issues/1606

PR #1611, opened 2025-12-22
✨ Add `needs_fields`, deprecate `need_extra_options`, `need_statuses`, `need_tags`

chrisjsewell:
This pull request migrates the configuration for extra options from the old `needs_extra_options` list-based format to a new, more flexible `needs_fields` dictionary-based format throughout the codebase, documentation, and tests. It deprecates the old config, adds validation and warnings for deprecated usage, and updates the schema logic to allow for easier extension and overriding of core fields. Additionally, it updates test snapshots and documentation to reflect the new configuration system.

`needs_fields` works similar to `needs_extra_options`, for defining additional fields, but also allows for core fields to be overridden/specialized.
In doing so, it also provides a pathway to consolidate more configuration that encapsulate a single field schema and allow future composability of schemas

Core field specialization, works analogous to [SysML2](https://www.omg.org/sysml/sysmlv2/), in that schemas inherit from the base schema, and any modifications must only "narrow" the type constraints, obeying the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).
The following specialization rules then apply:

- General (all types):
    - The ``type`` of the child must be the same as the parent.
    - For ``array`` type fields, the ``item_type`` of the child must be the same as the parent.

- String type constraints:
    - ``const``: Child cannot change the constant value.
    - ``enum``: Child's enum values must be a subset of the parent's enum values.
    - ``pattern``: Child must have the same pattern (cannot be overridden).
    - ``format``: Child must have the same format (cannot be overridden).
    - ``minLength``: Child's minimum length must be ≥ the parent's.
    - ``maxLength``: Child's maximum length must be ≤ the parent's.

- Boolean type constraints:
    - ``const``: Child must have the same constant value as the parent.

- Number and Integer type constraints:
    - ``const``: Child must have the same constant value as the parent.
    - ``enum``: Child's enum values must be a subset of the parent's enum values.
    - ``minimum``: Child's minimum must be ≥ the parent's.
    - ``maximum``: Child's maximum must be ≤ the parent's.
    - ``exclusiveMinimum``: Child's exclusive minimum must be ≥ the parent's.
    - ``exclusiveMaximum``: Child's exclusive maximum must be ≤ the parent's.
    - ``multipleOf``: Child's multipleOf must be a multiple of the parent's.

- Array type constraints:
    - ``minItems``: Child's minimum must be ≥ the parent's.
    - ``maxItems``: Child's maximum must be ≤ the parent's.
    - ``minContains``: Child's minimum must be ≥ the parent's.
    - ``maxContains``: Child's maximum must be ≤ the parent's.
    - ``contains``: Child must match the parent's constraint (cannot be overridden).
    - ``items``: The above rules are applied also to the item schema.

These changes modernize the configuration system for need fields, improve extensibility, and provide clear migration paths for users.

- **Configuration system migration and deprecation:** Migrated all usage of `needs_extra_options` to the new `needs_fields` dictionary format in documentation, example configs, and tests, providing a more structured and extensible way to define options.
- **Schema and core field enhancements:** Updated schema creation logic to allow overriding core fields via `needs_fields`, and added warnings for deprecated `needs_statuses` and `needs_tags` configs, guiding users to use `needs_fields` for these customizations.
- **Type and internal API updates:** Refactored type definitions: replaced `NeedExtraOption` with `NeedOption` as the base type, and updated related code for clarity and extensibility.

- **Test and documentation updates:** Updated test snapshots and documentation to reflect the new configuration system, including changing descriptions from "Added by needs_extra_options config" to "Added by needs_fields config".

## Future improvements

The internal representation of a single field schema currently looks like:

```python
@dataclass(frozen=True, kw_only=True, slots=True)
class FieldSchema:
    name: str
    description: str = ""
    schema: ExtraOptionSchemaTypes
    nullable: bool = False
    directive_option: bool = False
    allow_dynamic_functions: bool = False
    allow_variant_functions: bool = False
    allow_defaults: bool = False
    allow_extend: bool = False
    predicate_defaults: tuple[
        tuple[str, FieldLiteralValue | FieldFunctionArray],
        ...,
    ] = ()
    default: None | FieldLiteralValue | FieldFunctionArray = None
```

In principle, all of these attributes could be controlled by `needs_options`, in particular:

- Allow items of `needs_fields` to define `nullable`
- Allow items of `needs_fields` to define defaults and predicate defaults, and deprecate `needs_global_options`
- Allow items of `needs_fields` to define `allow_dynamic_functions` and `allow_variant_functions`, and deprecate `needs_variants`

Also, currently only a single schema is allowed for all needs in a project.
Similar to SysML2, in may be more "expressive" in the future, to allow multiple schemas in a single project,
for instance, allowing individual need types to have their own specialized schemas.
This though would require some thought/work on how filters work, since not all needs would share the same attributes (although, needs with the same parent schema would)

### DIFF
--- a/tests/doc_test/extra_options/conf.py
+++ b/tests/doc_test/extra_options/conf.py
@@ -2,7 +2,7 @@ extensions = ["sphinx_needs"]
 
 needs_extra_options = ["updated"]
 
-needs_options = {
+needs_fields = {
     "introduced": {},
     "impacts": {"description": "What is the impact of this need?"},
     1: {},
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -190,9 +190,9 @@ By default it is set to:
 
    Please take a look into the  `PlantUML Manual <https://plantuml.com/>`_ for more details.
 
-.. _`needs_options`:
+.. _`needs_fields`:
 
-needs_options
+needs_fields
 ~~~~~~~~~~~~~
 
 .. versionadded:: 7.0.0
@@ -200,24 +200,24 @@ needs_options
 This setting allows you to customise the fields that can be used for each need;
 how they are defined, processed, and validated.
 
-``needs_options`` must be a dictionary, where each key is the name of the option, and the value is another dictionary defining the option.
+``needs_fields`` must be a dictionary, where each key is the name of the field, and the value is another dictionary defining the field.
 
-Options can either "specialize" existing core options or define completely new ones.
+Fields can either "specialize" existing core fields or define completely new ones.
 
-For new options the following can be defined:
+For new fields the following can be defined:
 
-- ``description``: A description of the option (optional).
+- ``description``: A description of the field (optional).
     This will be output in the schema of the :ref:`needs.json <needs_builder_format>`, and can be used by other tools.
-- ``schema``: A schema definition for the option (optional).
+- ``schema``: A schema definition for the field (optional).
     This uses the same format as :ref:`needs_schema_definitions` and :ref:`needs_schema_definitions_from_json`.
-    If specified, the option value will be validated against this schema when needs are parsed, see :ref:`schema_validation` for more details.
-    By default the option schema will be ``{"type": "string"}``.
+    If specified, the field value will be validated against this schema when needs are parsed, see :ref:`schema_validation` for more details.
+    By default the field schema will be ``{"type": "string"}``.
 
 For example:
 
 .. code-block:: python
 
-   needs_options = {
+   needs_fields = {
        "introduced": {
            "description": "Date when the need was introduced",
            "schema": {
@@ -234,7 +234,7 @@ For example:
        },
    }
 
-These options would then be availble to set in need directives, such as:
+These fields would then be availble to set in need directives, such as:
 
 .. code-block:: rst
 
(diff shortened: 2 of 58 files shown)


tests/doc_test/extra_options/conf.py
```diff
@@ -1,12 +1,13 @@
...
-needs_extra_options = [
-    "introduced",
-    "updated",
-    {"name": "impacts", "description": "What is the impact of this need?"},
-    {},
-    1,
-]
+needs_extra_options = ["updated"]
+
+needs_options = {
```
ubmarco: can you use both at the same time?

chrisjsewell: In principle, yes currently, but you will get a deprecation warning and would be expected to migrate fully to `needs_options`. then at some point in a future sphinx-needs release we would look to completely drop `needs_extra_options` (this would be a good while in the future though)

.github/workflows/ci.yaml
```diff
@@ -16,18 +16,18 @@ jobs:
...
-    steps:
-      - uses: actions/checkout@v4
-      - uses: useblocks/ubc-action@main
-        with:
-          version: 0.19.1
-      - name: Run 'ubc check docs'
-        run: ubc check docs
-      - name: Run 'ubc build index docs'
-        run: ubc build index docs --show-warnings
+  # ubc:
```
chrisjsewell, line 19: @ubmarco obviously this is not currently supported by ubcode, but it should not be an overly difficult thing to add in.
The question would be if we do the addition there before/after we merge/release this

ubmarco: We should maintain a compatibility matrix of SN and the ubCode backend. Deactivating this for now is ok, just need to think of this once we support the new structure.

PR #1613, opened 2025-12-22
✨ Add `nullable` key to `needs_fields` items

chrisjsewell:
This pull request introduces support for a `nullable` property in need field definitions, allowing fields to be explicitly marked as allowing or disallowing unset (None) values. The changes affect the configuration, schema validation, and documentation, and include updates to both the codebase and the test suite to ensure correct handling and enforcement of field nullability.

For example:

```python
needs_fields = {
   "introduced": {
      "description": "Date when the need was introduced",
      "schema": {
         "type": "string",
         "format": "date",
      },
      "nullable": False,
   },
}
```

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -212,6 +212,8 @@ For new fields the following can be defined:
     This uses the same format as :ref:`needs_schema_definitions` and :ref:`needs_schema_definitions_from_json`.
     If specified, the field value will be validated against this schema when needs are parsed, see :ref:`schema_validation` for more details.
     By default the field schema will be ``{"type": "string"}``.
+- ``nullable``: If set to ``True``, the field can be set to ``None`` (optional), e.g. if no value is specifically given and no default applies,
+    If ``False``, the field must have a value (either explicitly set or via default/predicates), otherwise the need is invalid and will not be created.
 - ``predicates``: A list of ``(match expression, value)`` tuples (optional).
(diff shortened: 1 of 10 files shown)


docs/configuration.rst
```diff
@@ -212,6 +212,8 @@ For new fields the following can be defined:
     This uses the same format as :ref:`needs_schema_definitions` and :ref:`needs_schema_definitions_from_json`.
     If specified, the field value will be validated against this schema when needs are parsed, see :ref:`schema_validation` for more details.
     By default the field schema will be ``{"type": "string"}``.
+- ``nullable``: If set to ``True``, the field can be set to ``None`` (optional), e.g. if no value is specifically given and no default applies,
+    If ``False``, the field must have a value (either explicitly set or via default/predicates), otherwise the need is invalid and will not be created.
```
ubmarco, line 216: How would users know the default for `nullable`? This depends on the field, we should state this somewhere.

chrisjsewell: Obviously, this is only for inherited core fields, but yes we should show their "full" configuration in the `needs_fields` documentation.
Maybe leave this for a follow-up PR?

chrisjsewell: Done https://sphinx-needs--1613.org.readthedocs.build/en/1613/configuration.html#core-field-specialization

Issue #1534, opened 2025-10-05
With Sphinx-Needs 6.0.0 it is not possible to use variant information on links


PhilipPartsch:
With Sphinx-Needs < 6.0.0 it was possible to set variants on "Sphinx-Needs element" links.
With #1524 it is no longer possible.

@ubmarco:
What is the work around to get it working again?
Or can we enable the feature again?

ubmarco:
The actual PR that removed the feature is https://github.com/useblocks/sphinx-needs/pull/1489.
Please go through the rational for the decision.

During the implementation of strong typing, and after above PR was merged, we decided to introduce the new `<< >>` syntax for variants.
With this I see daylight for bringing this back.

Can you please give me an example of how you used this with links / how you plan to use it with SN 6?

Issue #915, opened 2023-04-21
Variant to allow multiple values


David-Le-Nir:
It seems not possible to define a variant value with a list of values, i.e I can't do

`:my_field: var_a: val1, val2; var_b: val3`

only the val1 appears for my_field and variant var_a.
if using space, it is seen as a single value.

I understand the complexity of creating a regexp to accept all possibilities

Actually it would be possible if only the `;` char was used to separate variant keys and not both `,` and `;`. IMHO this is the simplest fix we could provide for this use case 

Another possibility would be to allow the use of `[ ]` to define a list, something like
`:my_field: var_a: [val1, val2]; var_b: val3`
but I guess the parsing will be more complex

PR #1614, opened 2025-12-23
✨🎄 Add `parse_variants` to `needs_fields` / `needs_extra_link`, deprecate `needs_variant_options`

chrisjsewell:
This pull request introduces a new, more flexible way to enable variant support for need fields and links by adding a `parse_variants` option to the configuration. It deprecates the old `needs_variant_options` setting in favor of specifying `parse_variants` directly in `needs_fields` and `needs_extra_links`. The documentation and codebase are updated to reflect this change, improving clarity and flexibility for users configuring variant handling.

**Configuration and API changes:**

- Added a `parse_variants` boolean option to `needs_fields` and `needs_extra_links`, allowing users to enable variant parsing on a per-field or per-link basis. This replaces the previous global `needs_variant_options` list.

- Updated the API (`add_extra_option`) and configuration loading logic to accept and propagate the new `parse_variants` parameter.

**Deprecation and backward compatibility:**

- Deprecated the `needs_variant_options` configuration option. If used, a warning is issued, and backward compatibility is maintained by enabling `parse_variants` for the specified fields.
**Documentation updates:**

- Updated documentation in `docs/configuration.rst` and `docs/dynamic_functions.rst` to describe the new `parse_variants` option and mark `needs_variant_options` as deprecated. Examples and usage instructions are revised accordingly.

**Schema and internal logic changes:**

- Refactored schema and internal logic to use `parse_variants` instead of `allow_variant_functions`, and updated related variable names and validation.

**Example and test configuration updates:**

- Updated example project configuration (`docs/ubproject.toml`) to use the new `parse_variants` option instead of the deprecated `variant_options` list.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -178,4 +178,4 @@ VS Code Extension
 
 The VS Code extension `ubCode <https://marketplace.visualstudio.com/items?itemName=useblocks.ubcode>`_ provides 
 support for Sphinx-Needs.
-See more details in the `Documentation <https://docs.useblocks.com/ubcode/>`_.
+See more details in the `Documentation <https://ubcode.useblocks.com>`_.


sphinx_needs/needs.py
```diff
@@ -775,28 +781,34 @@ def check_configuration(app: Sphinx, config: Config) -> None:
...
+        }
+        for option in needs_config._variant_options:
+            # Check variant option is added to an allowed field
+            if option in link_types:
+                raise NeedsConfigException(
+                    f"Variant option `{option}` is a link type. This is not allowed."
+                )
+            if option not in extra_options and option not in allowed_internal_variants:
+                raise NeedsConfigException(
+                    f"Variant option `{option}` is not added in extra options. "
```
ubmarco: Line-Suggestions stop working for me... we should rephrase this to point to `needs_fields`.

sphinx_needs/needs.py
```diff
@@ -775,28 +781,34 @@ def check_configuration(app: Sphinx, config: Config) -> None:
...
+            None,
+        )
+        allowed_internal_variants = {
+            k for k, v in NeedsCoreFields.items() if v.get("allow_variants")
+        }
+        for option in needs_config._variant_options:
+            # Check variant option is added to an allowed field
+            if option in link_types:
+                raise NeedsConfigException(
+                    f"Variant option `{option}` is a link type. This is not allowed."
```
ubmarco, line 805: This is requested already, to support variants for link fields. Maybe a good time to think about this again.
https://github.com/useblocks/sphinx-needs/issues/1534
https://github.com/useblocks/sphinx-needs/issues/915
Generally the question is how to express array return values in variant literals.

chrisjsewell: Yeh this already done in this PR lol; parse_variant is also added to needs_extra_links

sphinx_needs/needs.py
chrisjsewell: fixed

chrisjsewell:
> If the outer type is an array, doesn't it split first?

Yes, the parsing logic for arrays is different to non-arrays, that's the point. 

Anyway, the parsing logic had not changed in this PR

Issue #1616, opened 2025-12-27
needtable :sort: option ignored with datatables style


sebastiansetzer:
The `:sort:` option works for `:style: table`, but for the (default) `datatables`, it doesn't. Datatables are always sorted by first column, until the user changes the sorting by pressing the little arrows.

I'm using Alabaster theme. Does that matter (would the issue be solved by another theme)?

PhilipPartsch:
Hello @sebastiansetzer,
thanks for the report.
The documentation of [needtable](https://sphinx-needs.readthedocs.io/en/latest/directives/needtable.html#sort) describes the behavior as intended.
 You can find it at the end of the chapter:

> Note
> 
> Sorting only works if you use the standard sphinx-table for output: :style: table. By default, tables generated with DatabTables uses Javascript to sort results.

Does this help you?

sebastiansetzer:
> The documentation of [needtable](https://sphinx-needs.readthedocs.io/en/latest/directives/needtable.html#sort) describes the behavior as intended.

Didn't see that, thanks for pointing it out.

One reason why I didn't see it may be that I searched for "datatable". And unfortunately, that typo "DatabTables" is in the docs, it's not a mistake in your answer.

> Does this help you?

It's a bit sad - Would be nice if `:sort:` would be used to define the default sorting in datatables.
But I've got a workaround, putting the column I want it to be sorted on to the front.

PR #1619, opened 2026-01-01
📚 Fix typo in sort of needtable documentation

PhilipPartsch:
In #1616, with a correct spelling the user would have found the restriction by himself. Here we fix it, so a search would find this restriction more easliy.


ubmarco (approved):

chrisjsewell (approved):
cheers!

PR #1639, opened 2026-02-02
🔧 Refactor `populate_field_type` to use type-directed schema walking

chrisjsewell:
This PR simplifies the `populate_field_type` function by replacing runtime keyword detection with type-directed traversal of the `SchemasRootType` structure.

### Refactored `populate_field_type` (~190 → ~130 lines)

**Before:** Generic recursive function operating on `Any`, detecting JSON schema keywords (`properties`, `items`, `contains`, etc.) at runtime to infer structure, with special-case handling for network context.

**After:** Type-directed implementation that surgically walks the known typed structure:
- `SchemasRootType` → `select` / `validate`
- `ValidateSchemaType` → `local` / `network`
- `NeedFieldsSchemaType` → `properties` / `allOf`
- `ResolvedLinkSchemaType` → recursive `items` / `contains`

### Key improvements

1. **Clearer structure** — 5 focused helper functions with single responsibilities:
   - `_process_validate_schema`
   - `_process_resolved_link`
   - `_process_need_fields_schema`
   - `_inject_field_type`
   - `_inject_type_and_item_type`

2. **Eliminated ambiguity** — No more `is_inside_network` path checks; the type structure dictates context

3. **Defensive handling** — All field accesses guarded with `isinstance()` checks; blanket exception handler wraps unexpected errors with context

4. **Unified field lookup** — Single pattern for extra/link/core fields with fallback to `NeedsCoreFields`

### Added tests

New test file `tests/schema/test_populate_field_type.py` with 32 tests covering:
- Type injection for extra fields, link fields, and core fields
- Array type injection with `items`/`contains`
- Recursive processing of `network` and `allOf`
- Error handling for unknown fields, type mismatches, and malformed schemas
- Defensive behavior for missing/invalid structure

## Breaking changes

None. All existing tests pass. Error messages for malformed schemas may differ slightly in wording but still provide equivalent information.


tests/schema/test_populate_field_type.py
```diff
@@ -0,0 +1,728 @@
...
+                }
+            },
+        }
+        fields_schema = create_mock_fields_schema(
+            link_fields={"links": ("array", "string")}
+        )
+
+        populate_field_type(schema, "[0]", fields_schema)
+
+        assert schema["validate"]["network"]["links"]["type"] == "array"
```
ubmarco, line 266: I do not expect this injection, as we never run the `items` or `contains` key through any validator. It's a custom modeling because the targets are resolved first, and we look into the `local` key first. However it's also not a mistake, just unnecessary.

Interestingly, it is also in the [modeling](https://github.com/useblocks/sphinx-needs/blob/2b9d23e4663e0d988e2a099a396a9ed466dca899/sphinx_needs/schema/config.py#L510).

docs/schema/local_validation_cache.md
```diff
@@ -0,0 +1,294 @@
+# Local Validation Caching System
+
+This document explains the local validation caching system implemented in `sphinx_needs/schema/core.py` to optimize schema validation performance.
+
+## Overview
+
+The caching system avoids redundant JSON schema validation when the same need is validated against the same schema multiple times. This is common in network validation where linked needs are traversed repeatedly from different starting points.
+
+## Problem Statement
```
ubmarco: In practice, this might not happen often. The network target need validation is rule specific. So that case only appears, if the same need is targeted by multiple needs selected by the same rule.

ubmarco (approved):
2 minor comments, good for merge

docs/schema/local_validation_cache.md
chrisjsewell: Oh ha, this document wasn't intended to be in this PR, it was for ubcode

PR #1642, opened 2026-02-08
🔧 Rename "option" to "field"

chrisjsewell:
Inline with moving from `needs_extra_options` to `needs_fields`.

This only changes internal naming, and wording in exception/warning text, so should not be backward-incompatible in any way.


ubmax (approved):

ubmarco (commented):
LGTM

PR #1645, opened 2026-02-09
‼️ Make `needs_fields` and `add_field` default to nullable and no default

chrisjsewell:
Currently, for all of:

```python
needs_extra_options = [{"name": "field1"}]
add_extra_option("field1")
needs_fields = {"field1": {}}
add_field("field1")
```

These create a field with:

- a string type
- a non-nullable value
- a default value of `""`

whereas for all of:

```python
needs_extra_options = [{"name": "field1", "schema": {"type": "string"}}]
add_extra_option("field1", schema={"type": "string"})
needs_fields = {"field1": {"schema": {"type": "string"}}}
add_field("field1", schema={"type": "string"})
```

These create a field with:

- a string type
- a nullable value
- no default value (i.e. the default is null)

The former is not intuitive, and derives from wanting to be back-compatible, when we moved from untyped extra options.

---

This PR does not change the behaviour of `needs_extra_options` / `add_extra_option`, which are anyhow deprecated,
but for:

```python
needs_fields = {"field1": {}}
add_field("field1")
```

These now create a field with:

- a string type
- a nullable value
- no default value (i.e. the default is null)

---

In order to aide users in transitioning from `needs_extra_options` to `needs_fields`,
we want to clearly indicate that the behaviour is different.

i.e. if they want exactly the same behaviour they should use:

```python
needs_extra_options = [{"name": "field1"}]
needs_fields = {"field1": {"nullable": False, "default": ""}}
```

Therefore, if none of `schema`, `nullable` or `default` is added to an item in `needs_fields` (or `add_field`),
then we emit a warning:

```
Field 'my_field' (from add_field) has no 'schema', 'nullable' or 'default' defined,
which defaults to a string schema with nullable=True and no default.
To aide with backward compatibility please define at least one. [needs.config]
```

i.e. this would emit a warning:

```python
needs_fields = {"field1": {}}
```

but these would not:

```python
needs_fields = {"field1": {"nullable": True}}
needs_fields = {"field1": {"default": "abc"}}
```


ubmarco (approved):
LGTM. I think after merge and before release we need documentation pieces that explain the new behavior of `needs_fields` in contrast to the old world.

PR #1648, opened 2026-02-10
🧪 Add additional fixture tests for schemas

chrisjsewell:


tests/schema/fixtures/config.yml
```diff
@@ -502,3 +502,104 @@ schemas_validate_pattern_relative_subroutine_error:
...
+            const: impl
+    schemas:
+      - id: "impl-links-schema"
+        message: impl must derive from spec and be tested
+        select:
+          $ref: '#/$defs/type-impl'
+        validate:
+          network:
+            links:
+              tests:
```
ubmarco, line 585: The 2 newly added tests in this file essentially test the same, just for different config data.
```
Additional properties are not allowed ('derived_from' was unexpected)
Additional properties are not allowed ('tests' was unexpected)
```
This does not look like the initial goal when reading the message fields:
```
Schema that mixes network.links and direct network.<type> - MUST be rejected
impl must derive from spec and be tested
```
Any children under network is always a link type, so there is no such thing as `network.links`. `network.<link-type>` makes more sense.
Maybe I miss something.

tests/schema/fixtures/unevaluated.yml
```diff
@@ -46,3 +46,177 @@ unevaluated_allof_error:
...
+              status:
+                type: string
+            unevaluatedProperties: false
+  rst: |
+    .. impl:: title
+        :id: IMPL_1
+        :status: approved
+
+# Test: Core fields with defaults should be reduced away
+# This validates that title/status/tags are reduced when they have default values
```
ubmarco, line 74: Technically, `title` is not on default, it is set to `My Title`.
It is reduced away because it is not part of the schema `properties`.

tests/schema/fixtures/unevaluated.yml
```diff
@@ -46,3 +46,177 @@ unevaluated_allof_error:
...
+            unevaluatedProperties: false
+  rst: |
+    .. spec:: target
+        :id: SPEC_1
+
+    .. impl:: title
+        :id: IMPL_1
+        :links: SPEC_1
+
+# Test: Schema properties whitelist - keep field only if in schema
```
ubmarco, line 197: This actually tests something different. `priority` is an extra field.
It is still there because it is not `None`.

ubmarco (changes requested):

PR #1655, opened 2026-02-15
🔧 Refactor schema validation: separate select filtering from local validation

chrisjsewell:
### Motivation

`get_ontology_warnings` served two fundamentally different purposes — **select filtering** and **local validation** — distinguished only by a `reduce: bool` flag, with different parameter conventions (select never used `field_properties`, `user_message`, or `user_severity`). This made the function's contract unclear and coupled the two code paths unnecessarily.

Separately, `NeedItem.iter_schema_items()` hardcoded a fragile list of 5 core field names (`id`, `type`, `title`, `status`, `tags`) plus 3 source fields, with a TODO noting it should be reworked. It missed 8 core fields that have `add_to_field_schema=True` (e.g. `collapse`, `hide`, `layout`, `style`, `constraints`, `template`, `pre_template`, `post_template`).

#### Split `get_ontology_warnings` → two private functions (`sphinx_needs/schema/core.py`)

- **`_check_select_match`** — returns `bool`. Builds a projection of the need via `need.items()` filtered by a caller-provided `select_field_names` set, then calls `validator.compiled.is_valid()`. When `schema_debug_active` is enabled, writes debug files using the existing `select_success`/`select_fail` rules (including the projected need and compiled schema), guarded so no `OntologyWarning` construction occurs unless debug mode is on. Both rules are in the default `schema_debug_ignore` list, so no new files appear unless the user opts in.

- **`_validate_need_local`** — local/network-local validation path. Accepts `field_properties`, `fail_rule`, `success_rule`, and optional `user_message`/`user_severity`. Uses `reduce_need()` for data shaping.

- **Removed** the original `get_ontology_warnings` and its `reduce: bool` parameter.

#### Remove `iter_schema_items` (`sphinx_needs/need_item.py`)

- **Removed** `iter_schema_items()` with its hardcoded field list.
- Select projection now uses `NeedItem.items()` filtered by `select_field_names` in the caller — `NeedItem` no longer encodes any schema policy.

#### Wire up the field names (`sphinx_needs/schema/core.py`, `sphinx_needs/schema/process.py`)

- `validate_type_schema` now accepts `fields_schema: FieldsSchema` instead of a pre-computed `select_field_names` frozenset. It derives the select field name set internally (only when a `select` schema is present), keeping the select projection logic self-contained in `core.py`.
- `process_schemas` simply passes `needs_schema` through — no manual field-name construction.

### Design notes

- **Select vs validate are now structurally separate**, making it easier to apply different `required`/`unevaluatedProperties` semantics to each.
- **No hardcoded field names** — the set is derived from `FieldsSchema` at runtime, automatically picking up new core, extra, and link fields.
- The split positions future work on reinterpreting `required` as "value is not `None`" to be scoped to `_validate_need_local` without affecting select semantics.

### Verification

All 162 schema tests pass, 220 snapshots unchanged, mypy clean, ruff clean.

### DIFF
--- a/sphinx_needs/need_item.py
+++ b/sphinx_needs/need_item.py
@@ -669,25 +669,6 @@ class NeedItem:
             self._computed.items(),
         )
 
-    def iter_schema_items(self) -> Iterable[tuple[str, Any]]:
-        """Return the items of the need item that are relevant for schema validation."""
-        # TODO - this should be reworked to be more robust;
-        # e.g. probably all core fields should be included
-        return chain(
-            (
-                (k, v)
-                for k, v in self._core.items()
-                if k in ("id", "type", "title", "status", "tags")
-            ),
-            (
-                (k, v)
-                for k, v in self._source.dict_repr.items()
-                if k in ("docname", "is_import", "is_external")
-            ),
-            self._extras.items(),
-            self._links.items(),
-        )
-
     def iter_core_items(self) -> Iterable[tuple[str, Any]]:
         """Return the core items of the need item."""
         return chain(
--- a/sphinx_needs/schema/core.py
+++ b/sphinx_needs/schema/core.py
@@ -24,6 +24,7 @@ from sphinx_needs.schema.config import (
 )
 from sphinx_needs.schema.reporting import (
     OntologyWarning,
+    save_debug_file,
     save_debug_files,
 )
 from sphinx_needs.schema.utils import get_properties_from_schema
@@ -199,7 +200,12 @@ def validate_type_schema(
     for need in needs.values():
         # maintain state for nested network validation
         if select_validator is not None and not _check_select_match(
-            need, select_validator, select_field_names=select_field_names
+            need,
+            select_validator,
+            select_field_names=select_field_names,
+            config=config,
+            schema_path=[schema_name, "select"],
+            need_path=[need["id"]],
         ):
             continue
 
@@ -635,6 +641,9 @@ def _check_select_match(
     validator: SchemaValidator,
     *,
     select_field_names: frozenset[str],
+    config: NeedsSphinxConfig,
+    schema_path: list[str],
+    need_path: list[str],
 ) -> bool:
     """Check whether a need matches a ``select`` schema.
 
@@ -642,17 +651,41 @@ def _check_select_match(
     keeping only fields whose name is in ``select_field_names`` and
     stripping ``None`` values.
 
+    When :attr:`NeedsSphinxConfig.schema_debug_active` is enabled, a
+    debug file is written for each evaluated need using the
+    ``select_success`` / ``select_fail`` rule.
+
     :param need: The need to test.
     :param validator: Compiled select schema validator.
     :param select_field_names: Field names to include in the projection
         (typically built from :class:`FieldsSchema` iterators plus
         identity and source fields).
+    :param config: Sphinx-Needs configuration (used for debug file writing).
+    :param schema_path: Schema path for debug reporting.
+    :param need_path: Need path for debug reporting.
     :return: ``True`` if the need matches the select schema.
     """
     needs_json: dict[str, Any] = {
         k: v for k, v in need.items() if k in select_field_names and v is not None
     }
-    return validator.compiled.is_valid(instance=needs_json)
+    is_match = validator.compiled.is_valid(instance=needs_json)
+
+    if config.schema_debug_active:
+        rule = (
+            MessageRuleEnum.select_success if is_match else MessageRuleEnum.select_fail
+        )
+        warning: OntologyWarning = {
+            "rule": rule,
+            "severity": get_severity(rule),
+            "need": need,
+            "reduced_need": needs_json,
+            "final_schema": validator.raw,
+            "schema_path": schema_path,
+            "need_path": need_path,
+        }
+        save_debug_file(config, warning)
+
+    return is_match
 
 
 def _validate_need_local(
--- a/sphinx_needs/schema/core.py
+++ b/sphinx_needs/schema/core.py
@@ -155,16 +155,38 @@ def validate_type_schema(
     schema: SchemasRootType,
     needs: NeedsView,
     field_properties: Mapping[str, NeedFieldProperties],
+    *,
+    fields_schema: FieldsSchema,
 ) -> dict[str, list[OntologyWarning]]:
-    """Validate needs against a type schema."""
+    """Validate needs against a type schema.
+
+    :param fields_schema: The fields schema, used to derive the set of
+        field names visible to ``select`` schemas.
+    """
     need_2_warnings: dict[str, list[OntologyWarning]] = {}
 
     schema_name = get_schema_name(schema)
-    validator = (
-        compile_validator(cast(NeedFieldsSchemaType, schema["select"]))
-        if schema.get("select")
-        else None
-    )
+    select_validator: SchemaValidator | None = None
+    select_field_names: frozenset[str] = frozenset()
+    if schema.get("select"):
+        select_validator = compile_validator(
+            cast(NeedFieldsSchemaType, schema["select"])
+        )
+        select_field_names = frozenset(
+            {
+                *fields_schema.iter_core_field_names(),
+                *fields_schema.iter_extra_field_names(),
+                *fields_schema.iter_link_field_names(),
+                # identity fields not in the user-configurable schema
+                # but essential for select matching
+                "id",
+                "type",
+                # source fields
+                "docname",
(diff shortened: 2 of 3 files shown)


ubmarco (changes requested):
Hey, I'm ok with the todo fix and the whole endeavor of splitting the code paths, but I disagree with some things:

The debug functionality for select is of the same value as for validate. Last week I was looking for exactly this: for docname in select, are those needs selected?

The PR description says there is a different semantics around required / unevaluatedProperties in select and validate.local. Why is that? I want to require fields also for select, to say pick needs that have this field set. For unevaluatedProperties however I can't think of a use case for select. 

The function name _validate_need_local is not intuitive as it also does network validation, as per the PR description.

chrisjsewell:
> The debug functionality for select is of the same value as for validate. Last week I was looking for exactly this: for docname in select, are those needs selected?

re-added in [1cd3425](https://github.com/useblocks/sphinx-needs/pull/1655/commits/1cd34255b305737824914d31bbe6b8a448ce256d)

> The function name _validate_need_local is not intuitive as it also does network validation, as per the PR description.

This function ALWAYS acts on data coming from a `local` key in the JSON file, whether that be `validate -> local` or `network -> ... -> local`, so I think the name is consistent.

> The PR description says there is a different semantics around required / unevaluatedProperties in select and validate.local. Why is that?

## Broader design discussion: `reduce_need` and JSON Schema semantics

### The fundamental mismatch

JSON Schema was designed for validating **incoming/sparse data** (like an API request body) where fields may genuinely be absent. But needs are **fully-materialized objects** — every field is always present with a default value. This means two standard JSON Schema keywords lose their natural meaning:

- **`required`** → always passes, because every field is present (with a default)
- **`unevaluatedProperties: false`** → always fails, because needs have dozens of internal fields (`docname`, `is_external`, computed fields, etc.)

`reduce_need` is a **data-shaping workaround** that strips the need back down to simulate sparse input, making these keywords "work." But the resulting semantics diverge from standard JSON Schema in subtle ways.

However, for `required` there is a much simpler reinterpretation that avoids data reduction entirely: **`required` means "the value is not `None`/`null`"**. See the [proposed resolution](#proposed-resolution-for-required) below.

### What users actually want vs. what the keywords mean

| Keyword | Standard JSON Schema meaning | What sphinx-needs users mean |
|---|---|---|
| `required: ["asil"]` | "The key `asil` must exist in the object" | "The field `asil` must have a **non-null value** — i.e. it was set to something meaningful, not left at `None`" |
| `unevaluatedProperties: false` | "No properties beyond those declared in `properties`/`allOf`" | "Ensure my schema covers all fields/links on this need — reject any field I didn't explicitly declare in `properties`/`allOf`" |

For `required`, the gap is narrower than it first appears — see [below](#proposed-resolution-for-required). For `unevaluatedProperties`, the mismatch is more fundamental.

### `select` vs `validate` — different roles, different keyword needs

A type schema has two parts:

- **`select`** — a **query/filter**: "which needs does this schema apply to?" If a need fails the select, it's silently skipped.
- **`validate.local`** — a **constraint**: "what must these selected needs satisfy?" Failures emit warnings.

This distinction matters for `required` and `unevaluatedProperties`:

| Keyword | In `select` (query) | In `validate.local` (constraint) |
|---|---|---|
| `required` | ✅ Useful — "only apply to needs where `asil` is not null" | ✅ Useful — "every matched need **must** have a non-null `asil`" |
| `unevaluatedProperties` | ❌ Probably wrong — needs always have many fields; using this as a filter would deselect almost everything, or force the select to declare every field it wants to ignore | ⚠️ Debatable — see table above for the semantic gap |

In the current code, `select` uses `reduce=False` (the curated field set), which means `unevaluatedProperties: false` in a select would reject any need with fields beyond those declared in the select's `properties`. This is almost certainly not what a user intends — they just want to match on a few fields. If `unevaluatedProperties` is ever meaningful, it's in `validate.local` (with `reduce=True`), not in `select`.

**Possible guardrail**: Warn or error at config time if a `select` schema contains `unevaluatedProperties`.

### Why `reduce_need` is fragile

It conflates two separate concerns into one data transformation:

1. **Provenance** (was this field explicitly set?) — addressed by stripping defaults, but can't distinguish "explicitly set to `[]`" from "never set, defaulted to `[]`"
2. **Scoping** (which fields should the schema see?) — addressed by stripping internal fields, but uses heuristics (is it in `schema_properties`? is it a core field?) rather than a clear rule

### Proposed resolution for `required`

#### Reinterpret `required` as "value is not `None`/`null`"

Rather than stripping fields or tracking provenance, `required` can be given a simple, sphinx-needs-specific meaning:

> **`required: ["asil"]`** → the field `asil` has a value that is not `None`

This works because of how need fields are already structured:

- **Nullable fields** (e.g. `status`, `style`, all extra fields) default to `None` when not provided. These are the fields users naturally put in `required`. A `None` check correctly identifies "not provided."
- **Non-nullable fields** (e.g. `id`, `title`, `docname`) always have real values and would always pass `required` — which is correct, since they are always set.

This means `reduce_need` is **not needed** for `required` at all. The check is a pure value inspection:

> **⚠️ Important caveat for users**: `required` only checks for non-null values. Fields that default to a non-`None` value — such as **link fields** (default `[]`) and **`tags`** (default `[]`) — will **always** be considered "present" by `required`, even when the user never set them. For these fields, use value constraints instead:
>
> - `"tags": {"minItems": 1}` — "at least one tag must be provided"
> - `"links": {"minItems": 1}` — "at least one link must exist"
>
> `required` is meaningful for **nullable fields** (extra fields, `status`, `style`, etc.) where the default is `None`.

```python
# During validation:
for field in schema["required"]:
    if need[field] is None:
        # emit warning: "field is required but has no value"
```

#### Why this is better than provenance tracking

An alternative approach would be to persist provenance info (a `frozenset[str]` of explicitly-set field names on `NeedItem`). While more theoretically correct, the "not None" approach is preferable:

| Concern | `value is not None` | Provenance tracking (`_explicitly_set`) |
|---|---|---|
| Implementation complexity | Trivial — pure value check | Must thread through `generate_need`, `needextend`, imports, external needs |
| "Explicitly set to default" edge case | Misses it (field set to `None` looks unset) | ✅ Handles it |
| Practical impact of edge case | Negligible — setting a field to `None` is the same as not setting it | Adds complexity for a near-impossible scenario |
| `needextend` handling | Automatic — if `needextend` sets a value, it's no longer `None` | Must explicitly add to the provenance set |
| External/imported needs | Automatic — null fields in JSON map to `None` | Must reconstruct provenance from import data |

The edge case that provenance catches — "user explicitly set a nullable field to `None`" — is effectively meaningless. Setting `status` to `None` *is* the same as not setting it.

#### Semantics in `select` vs `validate`

- In **`select`**: `required: ["asil"]` means "only apply this schema to needs where `asil` is not null." No data reduction needed.
- In **`validate.local`**: `required: ["asil"]` means "every matched need must have a non-null `asil`." Failures emit warnings.

Both work naturally with a simple null check.

### Remaining issue: `unevaluatedProperties`

`required` is resolved by the above, but `unevaluatedProperties` still needs `reduce_need` (or an alternative) because it's fundamentally about **scoping** — which fields the schema should see.

#### Option A: augment the schema, not the data

Instead of stripping unknown fields from the data, inject all *known* need fields into the schema as permissive properties:

```python
# Before compilation, inject known fields so they're "evaluated"
for field in all_known_need_fields:
    if field not in schema.get("properties", {}):
        schema.setdefault("properties", {})[field] = True  # allows anything
```

Now `unevaluatedProperties: false` only rejects truly *unknown* fields — which shouldn't exist since the extension controls what fields are available.

**Caveat**: The current fixtures show users using `unevaluatedProperties` to say "needs should ONLY have `asil`, reject `comment`." If we inject all fields, this use case breaks. This reveals the user is really asking for something different: "restrict which extra fields a need is allowed to have" — not standard `unevaluatedProperties` semantics. A more honest approach might be a custom keyword like `x-needs-allowed-fields`.

#### Option B: keep a scoped data projection

Keep a data projection via `iter_schema_items()` (the curated field set) for `unevaluatedProperties`, but **decouple it from `required`**. The `reduce_need` function would only handle scoping, not provenance.

### Summary

| Keyword | Resolution | Needs `reduce_need`? |
|---|---|---|
| `required` | Reinterpret as "value is not `None`" | **No** |
| `unevaluatedProperties` | Augment schema (Option A) or scoped projection (Option B) | **Depends on option** |

Document clearly that `required` has sphinx-needs-specific semantics: it checks for non-null values rather than key presence (which is always true for fully-materialized need objects).

ubmarco (approved):

Issue #1321, opened 2024-10-08
`parts`/`arch` fields not populated if need is hidden


chrisjsewell:
As noted in https://github.com/useblocks/sphinx-needs/pull/1318#issue-2570995275:

If a need directive is set with the `hide` option then, currently, its content is not parsed, which leads to no parts or umls being extracted from it

PR #1685, opened 2026-03-18
♻️ Fix `links_from_content` to use parsed doctree nodes instead of regex

### LINKED ISSUE #1668: BUG: dynamic function links_from_content()
The sphinx needs internal dynamic function links_from_content() is introducing a bug in our project.

Example:

```restructuredtext
.. req::
   :id: Example_ID_123
   :references: [[links_from_content()]]

   The function :need:`XYZ function<NEED_XYZ>` shall be validated on a validation database which is ground truth labeled according to the  `EXAMPLE Guide <https://my_example_link.com/Example.html>`_ and consists of data that includes a variety of scenarios and conditions that the system may encounter in real-world application. This shall be ensured by compliance to the amount and distributions stated for EXAMPLE PROJECT in `Example Label Data Requirements <https://github.com/my_example_repo/example.csv>`_.
```

ERROR MESSAGE:

```
"Sphinx-Build failed because of:
  /bazel_output_base/execroot/_main/my_project_root/example.rst:7: WARNING: Need 'Example_ID_123' has unknown outgoing link 'https://github.com/my_example_repo/example.csv' in field 'references'"
```

Most likely root cause: this line in links_from_content():

```python
links = re.findall(r":need:`(\w+)`|:need:`.+\<(.+)\>`", source_need["content"]) 
```

Second pattern too undeterministic.

Sanity Check:

This issue does not occur when using this pattern for a need role reference in the content of a need + label:

```restructuredtext
.. req::
   :id: Example_ID_123
   :references: [[links_from_content()]]

   The function :need:`NEED_XYZ` shall be validated on a validation database which is ground truth labeled according to the  `EXAMPLE Guide <https://my_example_link.com/Example.html>`_ and consists of data that includes a variety of scenarios and conditions that the system may encounter in real-world application. This shall be ensured by compliance to the amount and distributions stated for EXAMPLE PROJECT in `Example Label Data Requirements <https://github.com/my_example_repo/example.csv>`_.
````

chrisjsewell:
Closes #1668

### Problem

`links_from_content` used a regex (`` r":need:\`(\w+)\`" ``) on the raw content string to extract need references. This was brittle:

- `\w+` doesn't match IDs containing hyphens, dots, or other valid characters
- Titled references like `` :need:`My title <SOME_ID>` `` weren't reliably matched
- The regex doesn't support MyST Markdown
- The regex duplicated parsing logic that Sphinx already performs

### Solution

Replace the regex with doctree node traversal:

1. Retrieve the stored `Need` node via `SphinxNeedsData.get_need_node()`
2. Walk the node tree with a custom traversal (`_find_need_refs`) that yields `NeedRef` nodes but skips nested `Need` nodes (mirroring the `find_parts` pattern)
3. Return `NeedLink` objects directly from `ref_node["need_link"]` — the already-parsed, authoritative reference created via `NeedLink.parse_address`

This uses the same parsed representation that Sphinx uses for cross-references, so it handles all valid ID formats and role syntax correctly. Returning `NeedLink` directly avoids a lossy `to_filter_string()` → `from_string()` round-trip (since `parse_address` and `from_string` handle edge cases differently, e.g. IDs containing `[`).

### Trade-offs

Needs without a stored doctree node will now emit a `[needs.dynamic_function]` warning and return an empty list. This affects:

- **External needs** (from `needs_external_needs`) — no parsed node available
- **Need parts** (`NeedPartItem`) — no per-part node stored
- **Hidden needs** — node not stored due to early return in `_create_need_node` (see issue #1321)

### Changes

- **`sphinx_needs/functions/common.py`** — Rewrote `links_from_content()` to use node traversal; returns `list[NeedLink]` instead of `list[str]`; added `_find_need_refs()` helper that skips nested `Need` nodes; removed `import re`
- **`sphinx_needs/functions/functions.py`** — Updated `DynamicFunction` protocol return type and runtime type checks to accept `NeedLink` in list returns
- **`sphinx_needs/needs_schema.py`** — Updated `LinkSchema.type_check` / `type_check_item` to accept `NeedLink` alongside `str`
- **`tests/doc_test/doc_df_links_from_content/`** — New test doc project exercising self-content extraction, cross-need extraction, hyphenated IDs, titled refs, filtered extraction, and nested needs (refs in nested needs are not collected)
- **`tests/test_dynamic_functions.py`** — Added `test_doc_df_links_from_content`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,7 +65,7 @@ jobs:
         run: sudo apt-get install graphviz
       - name: Install graphviz (windows)
         if: matrix.os == 'windows-latest'
-        run: choco install --no-progress graphviz
+        run: choco install -y --no-progress graphviz
       - name: Set Up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
         with:
--- a/sphinx_needs/functions/common.py
+++ b/sphinx_needs/functions/common.py
@@ -7,8 +7,10 @@ Collection of common sphinx-needs functions for dynamic values
 from __future__ import annotations
 
 import contextlib
+from collections.abc import Iterator
 from typing import Any
 
+from docutils import nodes
 from sphinx.application import Sphinx
 
 from sphinx_needs.config import NeedsSphinxConfig
@@ -19,7 +21,9 @@ from sphinx_needs.filter_common import (
     filter_single_need,
 )
 from sphinx_needs.logging import log_warning
-from sphinx_needs.need_item import NeedItem, NeedPartItem
+from sphinx_needs.need_item import NeedItem, NeedLink, NeedPartItem
+from sphinx_needs.nodes import Need
+from sphinx_needs.roles.need_ref import NeedRef
 from sphinx_needs.utils import logger
 from sphinx_needs.views import NeedsView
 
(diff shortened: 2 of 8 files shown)


ubmarco (approved):
Nice, really appreciate fixing this long awaited shortcoming.

PR #1686, opened 2026-03-18
🧪 Add some more tests for links_from_content

chrisjsewell:
Specifically, for warning about unknown needs and for using within the `ndf` role

### DIFF
--- a/tests/__snapshots__/test_dynamic_functions.ambr
+++ b/tests/__snapshots__/test_dynamic_functions.ambr
@@ -49,6 +49,7 @@
             'links_back': list([
               'CON_SPEC_1',
               'CON_SPEC_2',
+              'CON_SPEC_5',
             ]),
             'section_name': 'LINKS FROM CONTENT',
             'sections': list([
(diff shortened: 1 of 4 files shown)


tests/__snapshots__/test_dynamic_functions.ambr
```diff
@@ -171,8 +172,26 @@
...
             'type_name': 'Specification',
           }),
+          'CON_SPEC_5': dict({
+            'content': "This content contains :need:`CON_REQ_2` and :ndf:`links_from_content()`, :ndf:`links_from_content('CON_REQ_4')`, :ndf:`links_from_content('unknown2')`",
+            'docname': 'index',
+            'external_css': 'external_link',
+            'id': 'CON_SPEC_5',
+            'lineno': 51,
+            'links': list([
+              'CON_REQ_2',
```
ubmarco, line 182: Is that duplication intentional / valid / should be deduplicated?

chrisjsewell: there is nothing in sphinx-needs stopping duplication; link values should perhaps be sets rather than lists, but that is not something for this PR, i.e. it is working as currently expected

ubmarco (approved):

PR #1692, opened 2026-04-06
Better lineno in needs JSON

### LINKED ISSUE #1349: Lineno gets shifted if rst_prolog is present
I have encountered a bug where the `lineno` options inside the `NeedsInfoType` is offset if there is an [rst_prolog](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_prolog) specified that isn't empty.
The  offset seems to be exactly the lines of the rst_prolog + 1. 

A quick fix if you need it (for example in location of warnigns) is to count the rst_prolog newlines add 1 and subtract that from the lineno.
Kind of like this: `correct_line_nr = int(need['lineno']) - rst_prolog_len`

sebastiansetzer:
This should fix https://github.com/useblocks/sphinx-needs/issues/1349.

The fix for the `include` and `rst-prolog ` use cases is simple.

I just wonder if anyone relies on having unique line numbers for one "docname" and docname being a full rst document, not an include file. If that's the case, maybe instead of fixing the `lineno` attribute, we should add a new attribute (maybe call it "source position"?)
On the other hand, the only reason I can think of why you would not want include file names as "source" attribute is if you had one that you included in multiple places. And if you do that, you won't have needs in there, because those would then get duplicated.

The fix for the `list2need` use case is abusing the "source" string to pass line number  information from the list2needs directive to the generated needs directives.
I considered passing this information through a generated "option", but when I declare this in `needs_extra_options`, it shows up for all needs (as "None" when unused) in the JSON, so that's a bit ugly.

I guess the best solution would be to refactor the `list2need` directive to create the needs objects directly, instead of generating rst text which is then injected back into the parser. This would avoid the need to pass line numbers, and it would even avoid calling `state_machine.insert_input`, so it wouldn't corrupt line numbers for subsequent needs.
Of course, corrupted line numbers of subsequent needs are already fixed by using `state_machine.get_source_and_line`, and you'd want to keep this for the `include` and `rst-prolog ` use cases, so the important reason for the refactoring is just avoiding to encode line numbers in the source string.

Before attempting to do this refactoring, I would like to understand why the current solution with `state_machine.insert_input` was chosen. Directly creating the needs objects seems more obvious, so I wonder if I'm missing the reason for Chesterton's fence here.

My test setup is attached - not integrated into the sphinx-needs test suite so far.
[rst-lineno-bug-tests.zip](https://github.com/user-attachments/files/26505620/rst-lineno-bug-tests.zip)

### DIFF
--- a/sphinx_needs/directives/list2need.py
+++ b/sphinx_needs/directives/list2need.py
@@ -113,9 +113,11 @@ class List2NeedDirective(SphinxDirective):
         # Retrieve tags defined at list level
         tags = self.options.get("tags", "")
 
+        _, content_first_line = self.state_machine.get_source_and_line(self.content_offset+1)
+
         list_needs = []
         # Storing the data in a sorted list
-        for content_line in self.content:
+        for content_lineno, content_line in enumerate(self.content):
             match = LINE_REGEX.search(content_line)
             if not match:
                 continue
@@ -171,6 +173,7 @@ class List2NeedDirective(SphinxDirective):
                     "content": content.lstrip(),
                     "level": level,
                     "options": {},
+                    "lineno": content_first_line + content_lineno
                 }
                 list_needs.append(need)
             else:
@@ -182,7 +185,6 @@ class List2NeedDirective(SphinxDirective):
                 )
 
         # Finally creating the rst code
-        overall_text = []
         for index, list_need in enumerate(list_needs):
             # Search for meta data in the complete title/content
             search_string = list_need["title"] + list_need["content"]
@@ -224,11 +226,10 @@ class List2NeedDirective(SphinxDirective):
             if presentation == "nested":
                 indented_text_list = ["   " * list_need["level"] + x for x in text_list]
                 text_list = indented_text_list
-            overall_text += text_list
 
-        self.state_machine.insert_input(
-            overall_text, self.state_machine.document.attributes["source"]
-        )
+            self.state_machine.insert_input(
+                text_list, self.state_machine.document.attributes["source"] + "\0%d" % list_need["lineno"]
+            )
 
         return []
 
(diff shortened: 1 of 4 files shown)


chrisjsewell:
Thank you for this, @sebastiansetzer — and sorry it sat so long. Your diagnosis was
right: the line numbers are wrong because the directive generates text and hands it back
to the parser, and the fix is to stop doing that. #1790 does exactly that, so the
NUL-encoding half of your PR is no longer needed — with no `insert_input` there is
nothing to encode around.

You asked whether there was a reason `insert_input` was used, a Chesterton's fence. We
went looking and did not find one: nothing depends on it, and it appears to have been
the expedient way to reuse the need directive rather than a decision defended at the
time. The other half of what you found — that `lineno` is wrong for *every* directive
after any `insert_input`, not just after a `list2need` — is real and is being fixed
separately in #1789, which makes the stored `lineno` agree with the location warnings
already report.

We did not carry the `docname` part, and it is only fair to say why rather than leave it
unexplained. Measured on your branch: `:need:` hyperlinks across a `.. include::` come out
dead, and a need directive inside `list2need` content raises a `KeyError` during the
build. Those looked like consequences of changing which document a need belongs to
rather than slips, so we left that alone.

Issue #1349 is closed by the pair of PRs. Nothing needed from you — this is just so the
credit is where it belongs.

chrisjsewell:
Closing as superseded, with thanks: #1789 (merged) fixed the generated needs' line numbers, and #1790 (merged) removed the `insert_input` splice that was the root cause, so both halves this PR targeted are addressed. The per-need source *path* (the `docname` half of the problem) remains open as a possible follow-up, as discussed above.

PR #1695, opened 2026-04-29
🐛 Sort need links and backlinks naturally in needs.json and HTML

### LINKED ISSUE #1371: Need links are not sorted in needs.json
Hi @chrisjsewell ,

in our project, we need a reproducible build, i.e. the needs builder must generate the same output if fed with the same input. That's why we have set reproducible_needs_json = True.
We noticed, however, that lists of linked needs (e.g. our "implements" or "derives" options) are not alphabetically sorted. We hoped that the order in which the links are added is stable, but noted that the order of the linked need ID entries changes depending on whether we add an external needs source (with needs_external_needs). However, the change has no influence on the elements contained in the list of IDs, it just changes the order.

Is this a known "feature" (i.e. the needs json builder does not sort the link list to save run-time) or is this a regression? I actually thought that your argument for reproducible build output was that all lists are sorted anyways, which I've shown you (see internal chat) that it's not the case with sphinx-needs 4.1.

ubmarco:
## Summary

Closes #1371.

Need link and backlink lists are now naturally sorted (e.g. ``REQ_2`` < ``REQ_9`` < ``REQ_10``) so build outputs are reproducible regardless of need load order, including when ``needs_external_needs`` is configured. Sorting happens unconditionally — no ``needs_reproducible_json`` flag is required — and applies to both ``needs.json`` and HTML output.

This matches the behaviour [`@ubmarco` confirmed](https://github.com/useblocks/sphinx-needs/issues/1371#issuecomment-4343807737) was desired: sort by default, in both outputs, irrespective of the ``needs_reproducible_json`` setting (which still keeps its existing role of stripping timestamps).

## Implementation

- ``NeedItem.sort_links()`` sorts ``_links``, ``_backlinks`` and per-part ``backlinks`` in place using a natural sort key (digit runs are compared as ints, with the result alternating ``str``/``int`` so mixed-type comparisons are well-defined).
- ``resolve_links()`` calls ``sort_links()`` on every need at the end of the post-processing pass, so all downstream consumers — JSON serialization in ``needsfile.py`` and the HTML rendering of outgoing/incoming refs in ``roles/need_outgoing.py`` and ``roles/need_incoming.py`` — see sorted lists.

## Tests

New ``tests/test_needs_sort.py`` is YAML-driven and snapshot-based, modelled on ``tests/schema/test_schema.py``:

- ``tests/fixtures/sort_links.yml`` declares 5 cases:
  - ``alphabetical_outgoing`` — basic alphabetical sort of outgoing links
  - ``natural_outgoing`` — natural sort with digits (proves ``REQ_10`` lands after ``REQ_9``)
  - ``mixed_alphanumeric`` — mixed letter+number IDs
  - ``backlinks_sorted`` — backlinks naturally sorted (``SPEC_1, SPEC_2, SPEC_10``)
  - ``custom_link_types`` — sort works for custom link types (``implements``, ``derives``)
- For each case the test reads ``needs.json``, derives link-field names from the embedded ``needs_schema`` (``field_type ∈ {links, backlinks}``), and snapshots the link/backlink fields per need. It also reads ``index.html``, extracts ``(source need, target need)`` ref pairs by matching ``title="<source>"`` on link anchors, and snapshots that mapping. Both snapshots clearly show the natural ordering.

Existing snapshots in 8 unrelated test files were updated for the new sorted order; all changes are mechanical reorderings within link lists.

## Test plan

- [x] CI passes
- [x] ``pytest tests/test_needs_sort.py`` — 5 cases pass, 10 snapshots match
- [x] ``pytest tests/`` — no link-related regressions

## Changelog

The bug-fix entry is in a new ``Unreleased`` section above 8.0.0 (now marked released on 19.03.2026).

### DIFF
--- a/sphinx_needs/need_item.py
+++ b/sphinx_needs/need_item.py
@@ -379,13 +379,12 @@ def _natural_sort_key(value: str) -> list[str | int]:
     """Build a sort key for natural ordering: digit runs are compared as ints.
 
     For example, ``REQ_2`` < ``REQ_9`` < ``REQ_10`` (instead of the default
-    lexicographic ``REQ_10`` < ``REQ_2`` < ``REQ_9``). The result alternates
-    between str and int, always starting with a str, so mixed-type comparisons
-    are well-defined.
+    lexicographic ``REQ_10`` < ``REQ_2`` < ``REQ_9``). Non-digit runs are
+    case-folded so the order is case-insensitive.
     """
     return [
-        int(part) if i % 2 else part
-        for i, part in enumerate(_NATURAL_SORT_RE.split(value))
+        int(tok) if tok.isdigit() else tok.casefold()
+        for tok in _NATURAL_SORT_RE.split(value)
     ]
 
 
@@ -907,21 +906,23 @@ class NeedItem:
                 part.backlinks[k] = []
 
     def sort_links(self) -> None:
-        """Sort all link and backlink lists in place using natural ordering.
+        """Sort all link and backlink lists in place, removing duplicates.
 
         Sorts outgoing links, backlinks, and part backlinks by the link's
         string representation, treating embedded digit sequences as integers
         so that e.g. ``REQ_2`` < ``REQ_9`` < ``REQ_10``. This makes the order
         of linked need IDs deterministic, regardless of the order in which
         they were added or the iteration order of the needs collection.
+        Duplicate :class:`NeedLink` entries (same id, part, and condition)
+        are collapsed.
         """
-        for value in self._links.values():
-            value.sort(key=_link_natural_sort_key)
-        for value in self._backlinks.values():
-            value.sort(key=_link_natural_sort_key)
+        for key, value in self._links.items():
+            self._links[key] = sorted(set(value), key=_link_natural_sort_key)
+        for key, value in self._backlinks.items():
+            self._backlinks[key] = sorted(set(value), key=_link_natural_sort_key)
         for part in self._parts.values():
-            for value in part.backlinks.values():
-                value.sort(key=_link_natural_sort_key)
+            for key, value in part.backlinks.items():
+                part.backlinks[key] = sorted(set(value), key=_link_natural_sort_key)
 
     def add_backlink(self, link_type: str, backlink: str | NeedLink) -> None:
         """Add a backlink to the need."""
--- a/sphinx_needs/need_item.py
+++ b/sphinx_needs/need_item.py
@@ -9,6 +9,7 @@
 # For NeedItem, we allow mutability, but only for values, i.e. it should not allow adding or removing keys.
 from __future__ import annotations
 
+import re
 from collections.abc import Iterable, Iterator, Mapping, Sequence
 from dataclasses import dataclass, field
 from itertools import chain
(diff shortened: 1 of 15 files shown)


ubmarco:
This PR is missing a performance analysis. Sorting links in big projects can be expensive.

sphinx_needs/need_item.py
```diff
@@ -363,6 +364,28 @@ def to_link_string(self) -> str:
         return f"{base}{open_b}{self.condition}{close_b}"
 
 
+_NATURAL_SORT_RE = re.compile(r"(\d+)")
```
chrisjsewell, line 375: In ubcode we do it a little differently

```python

NATURAL_SORT_RE = re.compile(r"(\d+)")


def _natural_sort_key(s: str) -> list[str | int]:
    """Return a sort key that orders strings in natural sort order.

    Numeric segments are compared as integers, so ``"REQ_2"`` sorts
    before ``"REQ_10"``.
    """
    return [
        int(tok) if tok.isdigit() else tok.casefold()
        for tok in cast(list[str], _NATURAL_SORT_RE.split(s))
    ]


def _natural_sort_strings(iterable: Iterable[str]) -> list[str]:
    """Sort strings in natural sort order."""
    return sorted(iterable, key=_natural_sort_key)


```

sphinx_needs/need_item.py
```diff
@@ -875,6 +898,23 @@ def reset_backlinks(self) -> None:
             for k in part.backlinks:
                 part.backlinks[k] = []
 
+    def sort_links(self) -> None:
```
chrisjsewell, line 908: While doing, perhaps also remove duplicates

sphinx_needs/need_item.py
ubmarco: Good point, changed this so it's in sync (case insensitive sorting)

sphinx_needs/need_item.py
ubmarco: Done

ubmarco:
## Performance impact analysis

### Project under test

- **Files:** 200 RST source files
- **Needs:** 1022, spread over 13 need types
- **Fields per need:** 174 — 36 core, 130 extra options, 8 outgoing-link relations
- **Outgoing-link density:** mean 3.11 links/need (median 3, min 0, max 40)

Each outgoing-link relation also produces a computed backlink list per target need; this PR's natural-sort change applies to both outgoing-link lists and the computed backlinks, both of which are reflected in `needs.json`.

Both runs use the same input corpus and produce the same `needs.json` — only the order of links/backlinks differs between baseline and PR, as expected.

### Methodology

- **Baseline:** sphinx-needs tag `8.1.1`
- **Under test:** this PR (`mh-natural-sort`)
- **Sphinx:** 8.x
- **Builder:** `sphinx-build -W -b html` — wall clock of the build subprocess only
- **Iterations:** 5 per branch, sequential, single local Linux x86_64 host

### Results

| Revision | n | mean | median | stdev | min | max |
|---|---:|---:|---:|---:|---:|---:|
| `8.1.1` (baseline) | 5 | 52.71 s | 52.49 s | 1.337 s | 51.32 s | 54.42 s |
| `mh-natural-sort` (PR) | 5 | 52.69 s | 51.72 s | 1.506 s | 51.48 s | 54.84 s |

**Δ (PR − baseline): −0.03 s (−0.05 %)** — well inside the run-to-run noise (stdev ≈ 1.4 s, ~2.6 % of mean). **No measurable regression at this corpus size.**

chrisjsewell (approved):

Issue #1697, opened 2026-05-01
Support jinja templates in needs-role-need-template


PhilipPartsch:
[needs-role-need-template](https://sphinx-needs.readthedocs.io/en/latest/configuration.html#needs-role-need-template) shall support ninja templates instead of python format. So it is possible to have different templates depending on information.

Python format does not support string parsing on conditions. (e.g. show different text depending on type, needpart, ...)

PR #1698, opened 2026-05-01
support jinja templates in needs-role-need-template

PhilipPartsch:
Implementation of #1697 1697

### DIFF
--- a/sphinx_needs/roles/need_ref.py
+++ b/sphinx_needs/roles/need_ref.py
@@ -194,18 +194,18 @@ def process_need_ref(
                     try:
                         # Build after potential title override via ref_name so
                         # templates see the latest values.
-                        need_context: dict[str, Any] = {
+                        need_context_role: dict[str, Any] = {
                             **dict_need,
                             "is_need": is_need,
                             "is_part": is_part,
                         }
-                        template_context: dict[str, Any] = {
-                            "need": need_context,
+                        template_context_role: dict[str, Any] = {
+                            "need": need_context_role,
                             **dict_need,
                             "is_need": is_need,
                             "is_part": is_part,
                         }
-                        link_text = role_template.render(template_context)
+                        link_text = role_template.render(template_context_role)
                     except Exception as exc:
                         log_warning(
                             log,
--- a/sphinx_needs/roles/need_ref.py
+++ b/sphinx_needs/roles/need_ref.py
@@ -71,7 +71,9 @@ def process_need_ref(
 
     role_template = None
     try:
-        role_template = compile_template(needs_config.role_need_template, autoescape=False)
+        role_template = compile_template(
+            needs_config.role_need_template, autoescape=False
+        )
     except Exception as exc:
         log_warning(
             log,
@@ -180,7 +182,9 @@ def process_need_ref(
                     link_text = f"{dict_need['title']} ({dict_need['id']})"
                 else:
                     try:
-                        link_text = role_template.render({"need": dict_need, **dict_need})
+                        link_text = role_template.render(
+                            {"need": dict_need, **dict_need}
+                        )
                     except Exception as exc:
                         log_warning(
                             log,
--- a/sphinx_needs/roles/need_ref.py
+++ b/sphinx_needs/roles/need_ref.py
@@ -128,15 +128,12 @@ def process_need_ref(
                 target_need
             )  # Transform a dict in a dict of {str, str}
 
+            is_part = need_id_part is not None
+            is_need = not is_part
+
+            dict_need["id_part"] = need_id_part or ""
             if need_id_part:
                 dict_need["id_complete"] = need_id_complete
-                dict_need["id_part"] = need_id_part
-                dict_need["is_need"] = "False"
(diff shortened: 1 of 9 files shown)


sphinx_needs/roles/need_ref.py
```diff
@@ -133,30 +159,40 @@ def process_need_ref(
 
             link_text = ""
             if ref_name and prefix in ref_name and postfix in ref_name:
-                # if ref_name is set and has prefix to process, we will do so.
-                ref_name = ref_name.replace(prefix, "{").replace(postfix, "}")
+                # Keep the user-facing [[...]] syntax, but render using Jinja.
```
chrisjsewell, line 159: I believe you should be able to do this in a more "proper" way, using this API.

```python

import minijinja

env = minijinja.Environment()
env.variable_start_string = "[["
env.variable_end_string = "]]"
```

Maybe you could add a way to apply this via the compile_template function

chrisjsewell (changes requested):

sphinx_needs/roles/need_ref.py
chrisjsewell: Confirmed — this is the right call. Rewriting `[[`/`]]` to `{{`/`}}` with `str.replace` is fragile (it also rewrites any literal `[[`/`]]` and can't distinguish delimiters from content). MiniJinja lets us set the delimiters directly, so the raw `[[ ... ]]` string is rendered as-is.

I've implemented the suggestion via `compile_template`, adding `variable_start_string` / `variable_end_string` parameters to `sphinx_needs/_jinja.py`:

```python
inline_template = compile_template(
    ref_name,
    autoescape=False,
    variable_start_string="[[",
    variable_end_string="]]",
)
link_text = inline_template.render(context)
```

The `str.replace(...)` step is then gone entirely. `render_template_string`, `compile_template`, `_new_env` and the cached-env helper all accept the two delimiter kwargs (defaulting to `{{`/`}}`), and the cache key includes them.

The change is on branch [`claude/sphinx-needs-pr-review-0vqlo6`](https://github.com/useblocks/sphinx-needs/tree/claude/sphinx-needs-pr-review-0vqlo6), rebased on the current `master` (this PR's branch is behind, and I don't have push access to the fork to update it directly). It also adds unit tests for the delimiter support and expands the role-template tests.

---

chrisjsewell (commented):
## General review

Nice feature — moving `needs_role_need_template` to Jinja is a clear improvement and fits the rest of the codebase now that everything renders through `sphinx_needs/_jinja.py`. Overall the implementation is sound. A few points, roughly in priority order:

### 1. Apply the delimiter suggestion (addressed)
As discussed in the inline thread, the `ref_name.replace("[[", "{{").replace("]]", "}}")` rewrite should be replaced by configuring MiniJinja's variable delimiters. I've implemented this on [`claude/sphinx-needs-pr-review-0vqlo6`](https://github.com/useblocks/sphinx-needs/tree/claude/sphinx-needs-pr-review-0vqlo6) (rebased on current `master`, since this branch is behind and I can't push to the fork). It threads `variable_start_string` / `variable_end_string` through `_jinja.py` and drops the string rewriting.

### 2. Silent-empty on unknown variables (behaviour change worth a note)
MiniJinja's default undefined behaviour renders unknown variables as an empty string, whereas the old `str.format` raised `KeyError`. So a typo like `{{ titel }}` now silently produces empty text instead of warning. This is probably acceptable (and hard to avoid for control structures), but it's a real behaviour change on top of the `{field}` → `{{ field }}` migration — worth calling out in the docs/changelog as breaking. (I added a `‼️` changelog entry + a `versionchanged` note on my branch.)

### 3. Duplicated context construction
`need_context`/`template_context` are built twice (inline branch and role branch) with identical shape. Small, but easy to factor into a helper — e.g. `_build_template_context(dict_need, is_part=is_part)` returning `{"need": {...}, **fields, is_need, is_part}`. Done on my branch.

### 4. `autoescape=False` is correct
Good call — `link_text` is wrapped in `nodes.Text(...)`, and docutils escapes at write time, so enabling autoescape here would double-escape. Just flagging it as verified, not a change request.

### 5. Testing — could be broader
The updated `test_role_need_template` covers the happy path well (control structures, filters, need parts, inline `[[...]]`). Two gaps I'd add:
- **Invalid-template fallback**: assert that a broken `needs_role_need_template` warns once and falls back to `title (id)` rather than crashing the build. (Added a `doc_role_need_template_invalid` fixture on my branch.)
- **Unit tests for the `_jinja` delimiter support** directly (default vs `[[`/`]]`, cache keyed on delimiters, block syntax preserved). (Added `tests/test_jinja.py`.)

### 6. Minor
- The PR branch is behind `master` and needs updating before merge; my branch is already rebased on the latest `master` if it's useful to pull from.
- `docs/configuration.rst` `versionchanged`: I used `8.2.0` to match the other unreleased entries — adjust if the target release differs.

Happy to open a PR from the review branch, or the diff can be cherry-picked into this one — whichever you prefer.

chrisjsewell:
Moving development to #1738

PR #1702, opened 2026-05-17
🐛 Fix needflow nodes rendering as black when no type color is set

### LINKED ISSUE #1664: needflow nodes extremely dark under dark mode (Sphinx 9.1.0, Sphinx‑Needs 7.0.0)
Environment:

OS: Windows 11

Sphinx: 9.1.0

Sphinx‑Needs: 7.0.0

Engine: tried both needs_flow_engine = "graphviz" and "plantuml"

Browser: one with forced dark mode (black boxes), one without (still quite dark)

Problem:

A simple .. uml:: PlantUML diagram renders with light boxes and dark text (looks fine).

The .. needflow:: assurance graph for my IEC 61508 SIL 2 Emergency Stop chain renders with very dark node backgrounds and very low text contrast. In a browser with forced dark mode the boxes look almost completely black.

Built‑in configs like :config: default (Graphviz) and :config: monochrome (PlantUML) do not significantly brighten the nodes.

Custom needs_graphviz_styles / needs_flow_configs attempts help only slightly; under dark mode the nodes remain much darker than normal UML diagrams.

Minimal needflow directive:

.. needflow::
   :root_id: HAZ_ESTOP_01
   :root_direction: both
   :link_types: ADDRESSES,ALLOCATES,REFINES,REALIZES,VERIFIES,EVIDENCES,REVIEWS
   :types: hazard,risk,sil,fsr,tsr,arch,verif,evidence,confirm
   :show_link_names:
   :root_depth: 10
   :config: default   # or monochrome

Request:

Is there a recommended “light”/high‑contrast style for needflow (PlantUML or Graphviz) that matches the docs examples?

If not, could you provide a sample needs_graphviz_styles / needs_flow_configs snippet that yields light node backgrounds and good contrast, suitable for browsers with forced dark mode?

I’ve attached:

The PNG of the needflow graph.

Image

ubmarco:
## Summary

Fixes #1664.

When a `needs_types` entry omitted the `color` field, `generate_need` fell back to `#000000`. That value was then forwarded to PlantUML as the node background colour and to Graphviz as `fillcolor` with `style=filled`, producing very dark, unreadable nodes — especially under browser forced dark mode.

The default is now an empty string. Each diagram emitter skips the colour suffix / `style=filled` when no colour is configured, so the diagram engine's own default node colour is used.

## Why this only surfaced recently

The `#000000` fallback has been in the code since 1.0, but it was unreachable until **3.0.0** (Aug 2024). Before PR #1185 (commit `eee2eb40`), the code used bracket access — `ntype["color"] or "#000000"` — so a `needs_types` entry missing `color` raised `KeyError` at build time and the field was *de facto* required. PR #1185 made `color` an optional `NotRequired[str]` and switched to `ntype.get("color") or "#000000"`, which silently substitutes the black fallback for users who omit the field.

Sphinx / PlantUML / Graphviz versions are not factors — the bug is purely on the Sphinx-Needs side and reproduces with every renderer version. Forced dark mode doesn't cause it; it just removes the surrounding light page that previously made the broken render mildly visible, which is why bug reports started arriving.

## Changes

- `sphinx_needs/api/need.py`: default `type_color` to `""` instead of `"#000000"`.
- `sphinx_needs/directives/needflow/_plantuml.py`: only emit the `#color` suffix when a colour is set.
- `sphinx_needs/directives/needflow/_graphviz.py` (subgraph path): only set `style=filled` when there is a fill colour. The plain-node path already had this guard.
- `sphinx_needs/directives/needuml.py`: same conditional `#color` emission as needflow.
- `sphinx_needs/directives/needgantt.py`: skip the `is colored in …` line when no colour is set.
- `sphinx_needs/config.py`: update the `NeedType.color` docstring.
- `docs/changelog.rst`: bug-fix entry under a new "Unreleased" section.

## Test plan

- [x] New parameterised regression test `tests/test_github_issues.py::test_doc_github_1664` covering both `plantuml` and `graphviz` engines with a `needs_types` entry that has no `color` field.
- [x] Verified the new test fails on the pre-fix code and passes after the fix.
- [x] Full test suite: 912 passed, 5 skipped (pre-existing benchmark errors due to a missing `sphinx_ai_index` extension are unrelated).

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,19 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+Bug fixes
+.........
+
+- 🐛 Fix ``needflow`` rendering very dark / black nodes when a need type has no
+  ``color`` set in ``needs_types`` (:issue:`1664`).
+  Previously a hard-coded ``#000000`` fallback was used as the fill color, which
+  produced unreadable nodes — especially under browser dark mode.
+  When no color is configured, no color is emitted and the diagram engine's
+  default node color is used.
+
 .. _`release:8.0.0`:
 
 8.0.0
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,6 +17,15 @@ Bug fixes
   When no color is configured, no color is emitted and the diagram engine's
   default node color is used.
 
+  .. note::
+
+     This is a minor behavior change for users with ``needs_types`` entries
+     that omit the ``color`` key: diagrams (``needflow``, ``needuml``,
+     ``needgantt``) that previously rendered such nodes as solid black will
+     now render them with the diagram engine's default node color (typically
+     light). To preserve the old appearance, set ``"color": "#000000"``
+     explicitly on the affected ``needs_types`` entry.
+
 .. _`release:8.0.0`:
 
 8.0.0
--- a/sphinx_needs/api/need.py
+++ b/sphinx_needs/api/need.py
@@ -373,7 +373,7 @@ def generate_need(
         "type": need_type,
         "type_name": need_type_data["title"],
         "type_prefix": need_type_data["prefix"],
-        "type_color": need_type_data.get("color") or "#000000",
+        "type_color": need_type_data.get("color") or "",
         "type_style": need_type_data.get("style") or "node",
         "title": core_values["title"],
         "status": core_values["status"],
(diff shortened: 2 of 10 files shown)


chrisjsewell (approved):
I have scrutinised the code, but the change makes sense in principle.
Perhaps this is somewhat of a breaking change (changing the default) and this should be made clear

ubmarco:
I added a note to the changelog entry to point out the possibly backwards incompatible change.

PR #1715, opened 2026-05-25
✨ Add `needs_variant_data` — rich structured variant data (deprecates `needs_filter_data`)

chrisjsewell:
## Summary

Introduces `needs_variant_data` (inline nested dict) and `needs_variant_data_file` (path to JSON file) as new configuration options, providing a namespaced, structured replacement for the flat `needs_filter_data`. In filter expressions, the data is accessible under the `var` namespace:

```rst
.. needtable::
   :filter: var.cpu == "arm" and var.build.debug == True

.. needtable::
   :filter: "feature_a" in var.build.features
```

`needs_filter_data` is deprecated with a warning pointing users to the new option.

## Motivation

- `needs_filter_data` is flat (`dict[str, str]`) and injects keys at root level, risking collisions with need field names.
- Users working with kconfig-like systems need nested structured data (e.g. `{"build": {"arch": "arm", "debug": true}}`).
- A dedicated `var` namespace avoids collisions and has been developed for compatibility with the ubcode query engine which already parses `var.a.b` attribute chains natively.

## Design Decisions

| Decision | Choice |
|----------|--------|
| Namespace | `var` (e.g. `var.cpu == "arm"`) |
| Config style | Single slot + file override (no named profiles) | | Switching | `sphinx-build -D needs_variant_data_file=other.json` | | `filter_data` | Deprecated with warning, keeps working | | Leaf types | `str`, `bool`, `int`, `float` |
| Arrays | Uniform scalar type only |
| Precedence | File loaded first, inline deep-merged on top (inline wins) | | Attribute access | Dot-only (`var.a.b`), no item access | | Missing key | Strict `AttributeError` (catches typos in filters) |

### Named profiles (rejected)

We considered letting users define multiple named variant configurations upfront and selecting between them:

```toml
[needs]
selected_variant = "arm_debug"

[needs.variant_profiles.arm_debug]
file = "variants/arm_debug.json"
```

**Why rejected:**
- Adds config complexity without clear benefit over just switching the file path directly.
- Users with multiple profiles already manage them as separate files; they don't need an indirection layer in sphinx config.
- The simpler "just point at a file" approach composes better with external tooling (CI matrices, kconfig generators, etc.) that already produce a single output file.
- Can always be added later as a convenience layer on top if demand materializes.

### Namespace name

Also considered:
- `env` — too easily confused with environment variables or Sphinx `BuildEnvironment`
- `cfg` — Rust-flavored, less obvious to Python/RST users
- No namespace (keep flat like `filter_data`) — loses the collision-safety motivation entirely

`var` was chosen as short, unambiguous, and evocative of "variant" / "variable".

### `SimpleNamespace` via `json.loads` object_hook (rejected)

Quick and elegant for prototyping, but rejected for production because:
- No validation of leaf types or uniform arrays
- No controlled error messages for missing keys (gives generic `AttributeError`)
- Exposes internal `SimpleNamespace` attributes (`__dict__`, etc.) to eval context
- No way to add `__contains__` for `"x" in var.tags` support without subclassing

### `__getitem__` fallback (rejected)

Considered allowing both `var.a.b` and `var["a"]["b"]`. Rejected because:
- Dot-access is significantly more readable in filter expressions: `var.build.debug` vs `var["build"]["debug"]`
- Consistent with Jinja2 template syntax where `foo.bar` and `foo["bar"]` are equivalent — users of Sphinx/Jinja are already familiar with dot-access for nested data
- Keeps the API surface minimal and forces a single idiomatic style
- Can be added later if needed

## User-facing Configuration

```python
# conf.py — inline (simple cases)
needs_variant_data = {
    "cpu": "arm",
    "debug": True,
    "build": {"optimization": 2, "features": ["feature_a", "feature_b"]},
}

# Or point to a file (kconfig-generated JSON, etc.)
needs_variant_data_file = "variants/arm_debug.json"

# Both can be combined: file loaded first, inline overrides on top
```

Switch at build time: `sphinx-build -D needs_variant_data_file=variants/x86_release.json docs/ _build/`

### Allowed Data Shape

```json
{
  "cpu": "arm",
  "debug": true,
  "build": {
    "optimization": 2,
    "features": ["feature_a", "feature_b"]
  }
}
```

Constraints:

- Top-level must be a dict
- Intermediate nodes: dicts only
- Leaf values: `str | bool | int | float | list[str] | list[bool] | list[int] | list[float]`
- Arrays must be uniform (all elements same scalar type)
- No `null` values, no dicts inside arrays

## Implementation

- **`sphinx_needs/variant_data.py`** (new) — `VariantDataProxy`, validation, JSON file loading, deep-merge
- **`sphinx_needs/config.py`** — Two new fields: `variant_data` and `variant_data_file`
- **`sphinx_needs/needs.py`** — Resolution in `prepare_env` (load file, merge, validate, store); deprecation warning in `check_configuration`
- **`sphinx_needs/filter_common.py`** — Injects `var` proxy at all filter eval sites (including `filter_single_need`, `filter_import_item`, `apply_default_predicate`)
- **`sphinx_needs/functions/functions.py`** — Injects `var` in variant expression context
- **`sphinx_needs/ubquery.py`** — AST fast-path support for `var.*` attribute chains (comparisons, `in`/`not in`), avoiding `eval()` overhead

### Fast-path details (`ubquery.py`)

The AST-based fast-path (`try_build_simple_predicate`) compiles simple filter expressions to native Python lambdas without `eval()`. This PR extends it to handle `var.*` attribute chains:

1. `_unpack_attribute_chain()` — unpacks `var.build.debug` AST into `("var", "build", "debug")`
2. `_resolve_chain()` — traverses the fallback context to resolve the chain at eval time
3. Comparison handlers recognize `ast.Attribute` chains rooted in `_FALLBACK_ROOTS` (`{"var"}`)
4. `"value" in var.field` membership tests are also compiled natively
5. Bare `var` as an `ast.Name` bails to the slow path (added to `_CONTEXT_ONLY_NAMES`)

## Documentation

- **`docs/configuration.rst`** — New `needs_variant_data` and `needs_variant_data_file` sections with full usage examples; `needs_filter_data` moved to "Deprecated Options" with a `.. deprecated:: 8.2.0` directive pointing to the new options
- **`docs/directives/need.rst`** — Updated `jinja_content` context reference
- **`docs/dynamic_functions.rst`** — Updated variant evaluation context reference
- All predicates context lists (under `needs_fields`, `needs_links`, `needs_global_options`) now reference `needs_variant_data`

## Tests

- **`tests/test_variant_data.py`** — Unit tests for `VariantDataProxy`, validation, deep-merge, file loading, fast-path compilation (28 tests)
- **`tests/test_variant_data_integration.py`** — Integration tests with Sphinx builds: inline config with `needs_warnings`, file loading with inline override (2 tests)
- **`tests/test_ubquery.py`** — `TestVariantDataFastPath` class: comparisons, nested access, membership, compound expressions, error cases (23 new tests)
- Updated existing tests (`test_needs_filter_data`, `test_variants`, `test_parallel_execution`) to expect the new deprecation warning

## Notes & Edge Cases

- **`-D` override granularity**: Sphinx `-D` cannot set nested dict keys individually. Switching is done by pointing at a different JSON file.
- **Array uniformity**: Validated at config-load time (fail fast with clear error).
- **`eval()` safety**: `VariantDataProxy` has no `__getitem__`, `__setattr__`, or `__delattr__` — only `__getattr__` for reads. Cannot mutate data from a filter.
- **Relative paths**: `needs_variant_data_file` is resolved relative to the Sphinx `confdir`.

## Future Improvements

- **Include variant data in `needs.json`**: The resolved variant data could be emitted in the `needs.json` metadata section, enabling downstream tools (e.g. ubcode) to know which variant configuration was active during the build. This would support diffing outputs across variants and reproducing builds.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -342,7 +342,7 @@ For ``predicates``, the match expression is a string, using Python syntax, that 
 - ``is_import`` (``bool``)
 - :ref:`needs_fields`
 - :ref:`needs_links` (``tuple[str, ...]``)
-- :ref:`needs_filter_data`
+- :ref:`needs_variant_data` (via the ``var`` namespace)
 
 For example:
 
@@ -454,7 +454,7 @@ For ``predicates``, the match expression is a string, using Python syntax, that 
 - ``is_import`` (``bool``)
 - :ref:`needs_fields`
 - :ref:`needs_links` (``tuple[str, ...]``)
-- :ref:`needs_filter_data`
+- :ref:`needs_variant_data` (via the ``var`` namespace)
 
 For example:
 
@@ -570,55 +570,101 @@ Use ``style_start`` and ``style_end`` like this:
    and orientation (``left``, ``rigth``, ``up`` and ``down``). We suggest to set the orientation
    in ``style_end`` like in the example above, as this is more often supported.
 
-.. _`needs_filter_data`:
+.. _`needs_variant_data`:
 
-needs_filter_data
-~~~~~~~~~~~~~~~~~
+needs_variant_data
+~~~~~~~~~~~~~~~~~~
 
-This option allows to use custom data inside a :ref:`filter_string`.
+.. versionadded:: 8.2.0
+
+Provides structured variant data accessible under the ``var`` namespace in :ref:`filter expressions <filter_string>`.
+This replaces the flat :ref:`needs_filter_data` option with a namespaced, nested data structure
+that avoids collisions with need field names.
+
+The dot-access syntax (``var.build.debug``) is more readable than dictionary-style access (``var["build"]["debug"]``)
+and mirrors `Jinja2 template syntax <https://jinja.palletsprojects.com/en/latest/templates/#variables>`__,
+which Sphinx users are already familiar with.
 
 Configuration example:
 
 .. code-block:: python
 
-   def custom_defined_func():
-       return "my_tag"
-
-   needs_filter_data = {
-       "current_variant": "project_x",
-       "sphinx_tag": custom_defined_func(),
+   needs_variant_data = {
+       "cpu": "arm",
+       "debug": True,
+       "build": {
+           "optimization": 2,
+           "features": ["feature_a", "feature_b"],
+       },
    }
 
-The defined ``needs_filter_data`` must be a dictionary. Its values can be a string variable or a custom defined
-function. The function get executed during config loading and must return a string.
-
-The value of ``needs_filter_data`` will be available as data inside :ref:`filter_string` and can be very powerful
-together with internal needs information to filter needs.
-
-The defined extra filter data can be used like this:
+The data is then available in filter expressions via the ``var`` namespace:
 
 .. code-block:: rst
 
-   .. needextend:: type == "req" and sphinx_tag in tags
-      :+tags: my_external_tag
+   .. needtable::
+      :filter: var.cpu == "arm" and var.build.debug == True
 
-or if project has :ref:`needs_fields` defined like:
+   .. needlist::
+      :filter: "feature_a" in var.build.features
+
+**Allowed data shape:**
+
+- Top-level must be a dictionary with string keys.
+- Intermediate nodes: dictionaries only.
+- Leaf values: ``str``, ``bool``, ``int``, ``float``, or uniform lists of these types.
+- Arrays must be uniform (all elements the same scalar type).
+- No ``None`` values, no dictionaries inside arrays.
+
+Accessing a missing key raises an ``AttributeError``, which helps catch typos in filter expressions.
+
+Default: ``{}``
+
+.. seealso::
+
+   :ref:`needs_variant_data_file` for loading variant data from a JSON file.
+
+.. _`needs_variant_data_file`:
+
+needs_variant_data_file
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 8.2.0
+
+Path to a JSON file containing variant data. The file is loaded and its contents
+are made available under the ``var`` namespace, just like :ref:`needs_variant_data`.
+
+If both ``needs_variant_data_file`` and ``needs_variant_data`` are set, the file is loaded first
+and the inline dictionary is deep-merged on top (inline values win on conflict).
+
+The path is resolved relative to the Sphinx ``confdir`` (the directory containing ``conf.py``).
+
+Configuration example:
 
 .. code-block:: python
 
-   needs_fields = {'variant': {}}
+   needs_variant_data_file = "variants/arm_debug.json"
 
-The defined extra filter data can also be used like:
+Where ``variants/arm_debug.json`` contains:
 
-.. code-block:: rst
+.. code-block:: json
 
-   .. needlist::
-      :filter: variant != current_variant
+   {
+     "cpu": "arm",
+     "debug": true,
+     "build": {
+       "optimization": 2,
+       "features": ["feature_a", "feature_b"]
+     }
+   }
 
-   .. needextract::
-      :filter: type == "story" and variant == current_variant
-      :layout: clean
-      :style: green_border
+This is particularly useful for switching configurations at build time:
+
+.. code-block:: bash
+
+   sphinx-build -D needs_variant_data_file=variants/x86_release.json docs/ _build/
+
+Default: ``None``
 
 .. _`needs_allow_unsafe_filters`:
 
@@ -2472,6 +2518,37 @@ final need and schema looks like.
 Deprecated Options
 ------------------
 
+.. _`needs_filter_data`:
+
+needs_filter_data
+~~~~~~~~~~~~~~~~~
+
+.. deprecated:: 8.2.0
+
+   Use :ref:`needs_variant_data` and :ref:`needs_variant_data_file` instead.
+   These provide a dedicated ``var`` namespace that avoids collisions with need field names
+   and supports nested structured data.
+
+This option allows to use custom data inside a :ref:`filter_string`.
+
+Configuration example:
+
+.. code-block:: python
+
+   def custom_defined_func():
+       return "my_tag"
+
+   needs_filter_data = {
+       "current_variant": "project_x",
+       "sphinx_tag": custom_defined_func(),
+   }
+
+The defined ``needs_filter_data`` must be a dictionary. Its values can be a string variable or a custom defined
+function. The function gets executed during config loading and must return a string.
+
+The value of ``needs_filter_data`` will be available as data inside :ref:`filter_string` and can be very powerful
+together with internal needs information to filter needs.
+
 .. _`needs_extra_options`:
 
 needs_extra_options
@@ -2672,7 +2749,7 @@ A match expression is a string, using Python syntax, that will be evaluated agai
 - ``is_import`` (``bool``)
 - :ref:`needs_fields`
 - :ref:`needs_links` (``tuple[str, ...]``)
-- :ref:`needs_filter_data`
+- :ref:`needs_variant_data` (via the ``var`` namespace)
 
 If no predicates match, the ``default`` value is used (if present).
 
--- a/docs/directives/need.rst
+++ b/docs/directives/need.rst
@@ -319,7 +319,7 @@ The option activates jinja-parsing for the content of a need.
 If the value is set to ``true``, you can specify `Jinja <https://jinja.palletsprojects.com/>`_ syntax in the content.
 
 The **:jinja_content:** option give access to all need data, including the original content
-and the data in :ref:`needs_filter_data`.
+and the data in :ref:`needs_variant_data`.
 
 If you set the option to **False**, you deactivate jinja-parsing for the need's content.
 
--- a/docs/dynamic_functions.rst
+++ b/docs/dynamic_functions.rst
@@ -170,7 +170,7 @@ Rules for specifying variant definitions
 * Variants gets checked from left to right.
 * When evaluating a variant definition, we use data from the current need object,
   `Sphinx-Tags <https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-t>`_,
-  and :ref:`needs_filter_data` as the context for filtering.
+  and :ref:`needs_variant_data` as the context for filtering.
   Sphinx tags are injected under the name ``build_tags`` as a set of strings.
 * You can set a *need option* to multiple variant definitions by separating each definition with either
   the ``,`` symbol, like ``var_a:open, ['name' in tags]:assigned``.|br|
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -189,6 +189,45 @@ If it is invalid or returns False, the related need is not taken into account fo
       .. needlist::
          :filter: "filter_example" in tags and (("B" in tags or ("spec" == type and "closed" == status)) or "test" == type)
 
+.. _filter_variant_data:
+
+Using variant data (``var``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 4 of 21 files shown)


chrisjsewell:
## Appendix: Kconfig-to-JSON Translation Analysis

This section analyses how Linux kernel [Kconfig](https://docs.kernel.org/kbuild/kconfig-language.html)
configurations could be converted to the JSON variant data format defined above,
what information can be captured, what is lost, and what extensions could be considered.

### What Kconfig Produces

After a Kconfig resolution (e.g. `make menuconfig` → `.config`), the output is a flat
key-value file of *resolved* symbol values:

```
CONFIG_SMP=y
CONFIG_NR_CPUS=8
CONFIG_PREEMPT_NONE=y
CONFIG_CC_OPTIMIZE_FOR_SIZE=n
CONFIG_DEFAULT_HOSTNAME="(none)"
```

A straightforward converter would emit:

```json
{
  "SMP": true,
  "NR_CPUS": 8,
  "PREEMPT_NONE": true,
  "CC_OPTIMIZE_FOR_SIZE": false,
  "DEFAULT_HOSTNAME": "(none)"
}
```

### What Maps Cleanly

| Kconfig concept | JSON representation | Notes |
|----------------|---------------------|-------|
| `bool` symbol (`=y` / `=n`) | `true` / `false` | Direct mapping |
| `int` / `hex` symbol | `int` (or `float`) | Hex values convert to decimal integers |
| `string` symbol | `str` | Direct mapping |
| `tristate` (`=y` / `=m` / `=n`) | `str`: `"y"`, `"m"`, `"n"` | See below — no native tristate in JSON |
| Menu hierarchy | Nested dicts | `{"NETWORKING": {"NETDEVICES": true, ...}}` |
| `choice` groups | Scalar (selected option name) or nested dict with one `true` entry | e.g. `{"PREEMPT": "NONE"}` |

### What Can Be Captured with Conventions

1. **Tristate values**: JSON has no tristate type. Two approaches:
   - **String encoding**: `"y"` / `"m"` / `"n"` — simple, but filters must compare strings
     (`var.MODULE_FOO == "m"`) rather than booleans.
   - **Boolean pair**: `{"enabled": true, "module": true}` — verbose but enables
     `var.MODULE_FOO.enabled == true and var.MODULE_FOO.module == true`.

   Recommendation: use string encoding (`"y"` / `"m"` / `"n"`) for compactness.
   Filter expressions become `var.MODULES == "y"` or `var.MODULES != "n"`.

2. **Menu structure as nesting**: The Kconfig menu tree can optionally be reflected as nested
   dicts (`{"NET": {"NETDEVICES": {...}}}`). This is useful for grouping but is optional —
   the flat `.config` output doesn't preserve it. A converter with access to the Kconfig
   source files can reconstruct hierarchy via `menu`/`endmenu` blocks and `depends on` inference.

3. **Choice groups**: A `choice` block allows exactly one selection. This maps naturally to a
   single string value naming the selected config:
   ```json
   {"PREEMPT_MODEL": "PREEMPT_NONE"}
   ```
   Or a sub-dict with boolean flags if multiple query styles are needed.

### What Is Lost / Not Capturable

| Kconfig concept | Why it's lost |
|----------------|---------------|
| **Dependencies** (`depends on`) | These are constraint *rules*, not data values. The resolved `.config` already satisfies them — they have no runtime meaning. |
| **Reverse dependencies** (`select`, `imply`) | Same as above — compile-time constraint logic. |
| **Conditional defaults** (`default X if Y`) | Resolution already applied; the chosen value is in `.config`. |
| **Prompts / help text** | UI metadata for the configuration tool, not variant state. |
| **Visibility rules** (`visible if`) | Only relevant during interactive configuration. |
| **Range constraints** (`range MIN MAX`) | Validation metadata, not a data value. Could be preserved as schema metadata but not as variant data itself. |
| **Source file locations** (`source "path/Kconfig"`) | Build system plumbing. |
| **Symbol type metadata** | JSON doesn't carry type annotations; a converter must infer `int` vs `string` from the Kconfig `type` declaration. |
| **Unset / undefined symbols** | Kconfig omits unset symbols from `.config`. The JSON would simply lack the key. With the current strict `AttributeError` on missing keys, filters referencing an absent symbol would fail loudly (which is desirable for catching typos but requires the converter to include *all* known symbols, even those set to `n`/default). |

### Practical Converter Approach

A minimal converter (`scripts/kconfig_to_variant_json.py`) would:

1. Parse the `.config` file (or use `kconfiglib` for richer access).
2. For each `CONFIG_X=val` line:
   - `=y` → `true`, `=n` → `false` (for bool) or `"y"` / `"m"` / `"n"` (for tristate)
   - Quoted values → `str`
   - Numeric values → `int`
3. Optionally group by menu structure if Kconfig source is available.
4. Write JSON conforming to the allowed data shape.

Using the Python [`kconfiglib`](https://github.com/ulfalizer/Kconfiglib) library gives
full access to symbol types, menu trees, and choice groups — enabling richer JSON output
than a raw `.config` parser.

### Possible Future Expansions

1. **`null` / optional values**: Currently disallowed. Allowing `null` would let a converter
   represent "symbol exists but is intentionally unset" (distinct from `false`). This is
   useful for tristate semantics (`null` = not applicable, `false` = disabled, `true` = enabled).
   Would require filter expression changes (`var.X is None`).

2. **Enum / tagged union types**: A dedicated enum type would cleanly represent tristate
   (`y | m | n`) and choice groups without string comparison. Example:
   ```json
   {"PREEMPT": {"_type": "choice", "value": "NONE", "options": ["NONE", "VOLUNTARY", "FULL"]}}
   ```
   This is heavy for the general case but could be useful for tooling that needs to
   present valid options in a UI.

3. **Metadata sidecar**: A companion schema file describing Kconfig-origin metadata
   (type, range, dependencies, help text) separate from the data values. The variant data
   JSON stays lean (values only), while a `variant_schema.json` carries the type info.
   This would enable IDE autocompletion and validation of `var.*` references.

4. **Array/set values for multi-select**: Some Kconfig patterns result in multiple related
   symbols being enabled (e.g. `HAVE_ARCH_*`). A converter could group these as arrays:
   ```json
   {"ARCH_FEATURES": ["TRACEHOOK", "SECCOMP", "KPROBES"]}
   ```
   This maps well to `"TRACEHOOK" in var.ARCH_FEATURES` filters.

5. **Conditional/variant-of-variants**: For builds that switch between multiple `.config`
   files (e.g. debug vs release kernel configs), the named profiles approach (currently
   rejected for simplicity) could be reconsidered as a second phase. This would let a single
   `ubproject.toml` reference multiple Kconfig-generated JSON files and select between them.

6. **Dot-separated flattening with prefix stripping**: Instead of nesting, a converter could
   emit `CONFIG_NET_NETDEVICES` as `var.NET.NETDEVICES` by splitting on underscores
   heuristically or by menu path. This is fragile without Kconfig source but is an option
   for projects with naming conventions.

### Summary

The resolved output of a Kconfig system (`.config`) maps well to the variant data JSON
format — scalar values fit directly, and the flat namespace is compatible with the
allowed shape. The primary design gap is tristate values (solved with string encoding)
and the loss of constraint/dependency metadata (which is acceptable since those are
build-system concerns, not runtime filter data). For teams using Kconfig to manage
hardware/build variants, a lightweight converter script is all that's needed to bridge
the two systems.

ubmarco (changes requested):
## Overview

I think the new `needs_variant_data` is what we want for variants handling.

## Manual findings

The deep merge of conf.py into the JSON could have some intended use case description.
I mainly see an injection of Sphinx build tags into the variant data.
Or other form of computed data.

I'm not sure yet about the deprecation / replacement of `needs_filter_data` yet.
There might be use cases besides the variant handling for this.
But after reading the documentation again, it just creates variables with string values.
So the new `needs_variant_data` is more powerful and flexible, and the old one can be easily replaced by the new one.

## AI Findings

### Bugs

**B1 — Uniformity check asymmetric for `bool`/`int` arrays** (`sphinx_needs/variant_data.py:45-56`)

`type(value[0])` is fine for `bool`-first arrays, but `int`-first arrays admit `bool` elements silently because `bool` is a subclass of `int`. Verified:

```
[1, True, 2]      → passes (BUG)
[True, 1, False]  → fails (correct)
```

The same hazard applies to the top-level scalar check `isinstance(value, _SCALAR_TYPES)`: `True` masquerades as `int`. If you want strictly uniform arrays, compare with `type(item) is first_type` instead of `isinstance`, and special-case `bool` so it isn't conflated with `int`.

**B2 — `test_repr` passes string where tuple is expected** (`tests/test_variant_data.py:54`)

```python
proxy = VariantDataProxy({"x": 1}, path="var")
```

`__init__` annotates `path: tuple[str, ...] = ()`. Passing the string `"var"` makes `__repr__` produce `"var.v.a.r"` (because `".".join("var")` iterates chars). The test passes only because the assertion is `"var" in repr(proxy)`. Should be `path=("var",)`.

**B3 — Missing version in deprecation directive** (`sphinx_needs/config.py:737`)

```rst
.. deprecated::
    Use :confval:`needs_variant_data` ...
```

The user-facing `docs/configuration.rst` correctly uses `.. deprecated:: 8.2.0`. The docstring in `config.py` omits the version. Sphinx will render this without a version label.

### Concerns

**C1 — CHANGELOG not updated**

`docs/changelog.rst` ends at `8.1.1`. The doc uses `.. versionadded:: 8.2.0` but there's no 8.2.0 changelog entry yet. Likely intentional (changelog generated at release time), but worth confirming.

**C2 — `var` namespace shadowing**

If a user has a need field literally named `var`, the variant proxy will silently override it in `filter_context`. Worth a doc note under the "Allowed data shape" section, or a runtime warning if `"var"` is detected in `needs_extra_options`/`needs_fields`.

**C3 — `VariantDataProxy` rebuilt every filter call**

In `filter_needs_and_parts`, `filter_single_need`, `apply_default_predicate`, `filter_import_item`, and per-iteration in `resolve_functions`, a fresh `VariantDataProxy(config.variant_data)` is allocated. For a large project running thousands of filter evals, this is unnecessary churn. Since the proxy is immutable, cache it once after `_resolve_variant_data_config` (e.g., on `data: SphinxNeedsData`) and reuse.

**C4 — `__getattr__` recreates nested proxies on every access**

`var.build.debug` allocates two proxies (`var.build`, then attribute read). Same fix as C3: pre-build the proxy tree once.

**C5 — `__eq__` defined without `__hash__`**

Defining `__eq__` implicitly sets `__hash__ = None`, making proxies unhashable. Probably fine, but if any downstream code stuffs `var` into a set/dict, it'll fail. Document or define `__hash__ = None` explicitly.

**C6 — Duplicate test surface**

`TestVariantDataFastPath` in `tests/test_ubquery.py` and `TestUbqueryFastPath` in `tests/test_variant_data.py` cover the same fast-path scenarios. Consolidate into one location to reduce maintenance.

**C7 — Untracked design docs in working tree**

`backlinks-schema-design.md`, `variants-in-code.md`, and `.claude/` show up as untracked. Confirm these aren't meant to ship with the PR (add to `.gitignore` or remove).

### Nits

- **N1** — `validate_variant_data` doesn't explicitly reject `None`. It works (None isn't in `_SCALAR_TYPES`), but the error message says "expected str/bool/int/float/list/dict, got NoneType" which is correct — just verify a test exists for it (it doesn't appear to).
- **N2** — Empty `needs_variant_data = {}` silently disables the `var` namespace. `var.anything` then raises `NameError` rather than the more helpful `AttributeError`. Consider always injecting the proxy when the config option is *touched*, even if empty.
- **N3** — `_resolve_variant_data_config` uses lazy `from … import` inside the function. Move imports to module top unless there's a circular-import reason.
- **N4** — The "Future Improvements" section in the PR body proposes emitting variant data in `needs.json`. Worth filing as a follow-up issue so it doesn't get lost.

ubmarco (approved):
LGTM

PR #1716, opened 2026-05-26
✨ Add `if` directive for variant-conditional content

chrisjsewell:
Introduces `.. if::` directive that evaluates a Python expression against the `var` namespace (from `needs_variant_data`) at parse time. If truthy, the directive body is parsed into the document; otherwise it is skipped entirely.

Supports all expression types: comparisons, boolean truthiness, nested attribute access (`var.build.opt > 1`), membership tests (`"x" in var.build.features`), and logical operators. Content may include section headers, nested directives, and needs.


sphinx_needs/directives/needif.py
```diff
@@ -0,0 +1,71 @@
...
+        except Exception as e:
+            log_warning(
+                LOGGER,
+                f"'if' directive expression failed: {expression!r} — {e}",
+                "if",
+                location=self.get_location(),
+            )
+            return []
+
+        if not result:
```
ubmarco: I vote for a check that `result` is actually a boolean.

ubmarco (changes requested):
LGTM, some findings:
- user facing docs are missing
- return statement should be bool
- maybe some more tests, such as
  - Attribute miss (var.unknown_key) — VariantDataProxy.__getattr__ raises AttributeError; verify it's caught and
  warned, not crashed.
  - Builtin access blocked (__import__('os').system(...)) — should fail with a NameError because __builtins__ is
  empty. Good defensive test.
  - needextend interaction: if a need inside an .. if:: is conditionally suppressed, what happens to needextend
  references to its ID? (Probably a "need not found" warning, which is fine — but worth a test.)

ubmarco (approved):

chrisjsewell:
@copilot fix the docs build warnings:

/home/docs/checkouts/readthedocs.org/user_builds/sphinx-needs/checkouts/1716/docs/directives/if.rst:8: WARNING: undefined label: 'variant_data' [ref.ref]
117	/home/docs/checkouts/readthedocs.org/user_builds/sphinx-needs/checkouts/1716/docs/directives/if.rst:11: WARNING: 'confval' reference target not found: needs_variant_data [ref.confval]

PR #1717, opened 2026-05-27
👌 Honor -D overrides when loading needs_from_toml

chrisjsewell:
Keep Sphinx CLI config overrides (`-D` / `confoverrides`) authoritative in `load_config_from_toml` by skipping TOML assignments for keys that were already overridden.

- Read override keys from `config.overrides`
- Skip setting `needs_<key>` when `<key>` or `needs_<key>` is overridden
- Skip setting schema entries when `needs_schema_` or `needs_schema_<key>` is overridden
- Add regression test verifying TOML does not overwrite `needs_reproducible_json` when provided via `confoverrides`

This preserves expected precedence: CLI overrides > TOML values > defaults.

### DIFF
--- a/sphinx_needs/needs.py
+++ b/sphinx_needs/needs.py
@@ -482,16 +482,27 @@ def load_config_from_toml(app: Sphinx, config: Config) -> None:
         return
 
     allowed_keys = NeedsSphinxConfig.field_names()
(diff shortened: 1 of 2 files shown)


ubmarco (approved):
Good change

PR #1718, opened 2026-05-29
🐛 Fix trailing separator in string_links for single-value fields

floroks:
The separator condition used len(data) which refers to the original string length instead of len(data_list) which is the split list. This caused a trailing "; " to always appear after single-value fields when needs_string_links is configured.

Added a regression test that verifies single-value string_links fields render without a trailing separator.

--
Florian Roks \<florian.roks@mercedes-benz.com\>, Mercedes-Benz Tech Innovation GmbH
[Provider Information](https://github.com/mercedes-benz/.github/blob/main/PROVIDER_INFORMATION.md)


ubmarco:
May I ask for a test?

floroks:
> May I ask for a test?

added a test, not sure if the test documentation referring to the original issue is acceptable, or a new issue should be created. the test however fails if the code is reverted, so it seems to be working perfectly fine.

another future extension could be making the separator configurable, not a huge fan of the semicolon

floroks:
pipeline build failures don't seem to be related to this MR

floroks:
rebased, pipeline now successful

chrisjsewell (approved):

chrisjsewell:
Thanks!

PR #1722, opened 2026-06-12
🐛 Fix spurious 'Unknown option' warning for needs_string_links fields

cpolzer:
## Problem

Fields declared via `needs_string_links` `options` (e.g. `'options': ['github']`) that are **not** also registered in `needs_fields` / `needs_extra_options` trigger a spurious `needs.directive` warning:

```
WARNING: Unknown option 'github' [needs.directive]
```

and the need is silently dropped from the build output.

## Root cause

`create_schema()` in `sphinx_needs/needs.py` only populates the schema's extra field registry from `needs_fields` and `needs_extra_options`. The `options` names inside `needs_string_links` entries are never registered.

Both `NeedDirective.run()` and `create_need()` validate option keys against `schema.iter_extra_field_names()` — any key absent from the registry hits the rejection path.

## Fix

After building the extra field schema from `needs_fields`, iterate `needs_config.string_links` and register any `options` field names not yet in the schema as **nullable string extra fields**. This is the correct layer — the schema is the single source of truth consulted by both the directive parser and `create_need()`.

Fields already registered via `needs_fields` are skipped, so there is no conflict when a project declares the field in both places.

## Tests

New file `tests/test_string_links.py` with three regression tests:

1. **No warning** — a field only in `string_links options` must not emit `Unknown option`
2. **Value stored** — the field value appears in `needs.json` (verified via the `needs` builder)
3. **Guard** — a truly unknown field (neither in `needs_fields` nor `string_links`) still warns

All existing `test_needtable` tests (which use `needs_string_links` with fields also in `needs_fields`) continue to pass.

## Checklist

- [x] Tests added
- [x] `tox -e ruff-check` passes
- [x] `tox -e mypy` passes
- [x] Changelog updated (`docs/changelog.rst`, Unreleased > Bug fixes)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,6 +24,13 @@ Improvements
 Bug fixes
 .........
 
+- 🐛 Fix spurious ``needs.directive`` "Unknown option" warning for fields
+  declared via :confval:`needs_string_links` ``options`` that are not also
+  listed in :confval:`needs_fields`.  Such fields are now automatically
+  registered as nullable string extra fields during schema creation, so they
+  are accepted by both the directive parser and ``create_need`` without any
+  additional configuration.
+
 - 🐛 Sort need link and backlink lists in ``needs.json`` and HTML output using
   natural, case-insensitive ordering (e.g. ``REQ_2`` < ``REQ_9`` < ``REQ_10``)
   and collapse duplicate entries, so build outputs are reproducible regardless
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -25,8 +25,8 @@ Bug fixes
 .........
 
 - 🐛 Fix spurious ``needs.directive`` "Unknown option" warning for fields
-  declared via :confval:`needs_string_links` ``options`` that are not also
-  listed in :confval:`needs_fields`.  Such fields are now automatically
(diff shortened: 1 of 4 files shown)


ubmarco:
The right solution is to add the fields into the configuration by the user/project.
Guessing the [field config](https://github.com/useblocks/sphinx-needs/blob/3cb6d54494409cdb689ff8ec5dc085ee5cdf9015/sphinx_needs/api/configuration.py#L139) from the `needs_string_links` configuration can lead to confusion.
I know of cases in the codebase (e.g. options for needgantt), however they are internal to SN, and also lead to problems in the past "Why can't I add a field of that name?". Just put the field into the ubproject.toml config (preferred), or conf.py.

ubmarco:
@cpolzer OK for you? I know it's more config effort, but at this stage I say `explict is better than implicit`.

cpolzer:
Partially, but closing this Mr then.

Edit: How do we then enable strict JSON schema validation with unevaluatedProperties: false ? in their needs_schema_definitions, the process_schemas() function (schema/process.py:29) validates all needs against the schema -> string fields will cause an error.

Is the goal to duplicate the string field as a field definition? 

Shall we would rather have regular fields enabled to act like a stringfield?

PR #1730, opened 2026-06-25
Track JSON files imported via needimport as dependencies

yhontyk:
## Problem
When JSON files are imported via `needimport`, sphinx-build imports the file once and doesn't pick up changes to these external files unless the RST file where the directive is declared changes or you rebuild the entire Sphinx environment. It's a little annoying when working with documentation incrementally/using `sphinx-autobuild` (make a small change -> rebuild to see how it looks). 

## Fix
[BuildEnvironment.note_dependency](https://www.sphinx-doc.org/en/master/extdev/envapi.html#sphinx.environment.BuildEnvironment.note_dependency) seems to be the way to declare an external file as a dependency, it makes Sphinx track changes to the file and rebuild it if the file timestamp changes. 

## Notes
I'm not sure how to handle changes to the remote files fetched from URLs, sorry, I am not a programmer. :( But I added a note about it to docs. I can open an issue for it too.

Let me know if I missed anything, thank you!

### DIFF
--- a/docs/directives/needimport.rst
+++ b/docs/directives/needimport.rst
@@ -56,6 +56,10 @@ The directive argument can be one of the following formats:
 
      .. needimport:: c:\absolute\path\to\needs.json
 
+.. note::
+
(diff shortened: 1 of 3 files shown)


chrisjsewell:
Yep looks good thanks 🙏

chrisjsewell (approved):

PR #1731, opened 2026-06-25
✨ Add `network_back` schema key to validate incoming links

### LINKED ISSUE #1590: Question: Incoming (top-down) link checks with sphinx-needs
I'm trying to understand if it is possible to perform top-down (incoming) trace checks in sphinx-needs. E.g., I have a system requirements document and I need to ensure that all (maybe filtered) needs have an outgoing trace to other items. I tried to do this in the `schema.json`.

E.g., I have the following `extra_link`

```toml
[[needs.extra_links]]
# req -> req, spec -> spec
# notice that 'directive' names == 'option' can't have multiple words ...
option = "decomposed-from"
incoming = "decomposes to"
outgoing = "decomposed from"
```

Upstream (outgoing) can be nicely done in the `schema.json`, e.g., I can ensure that all software requirements (that fit my "select") have at least one outgoing (upstream) trace to system requirements:

```json
{
  "schemas": [
    {
      "id": "srs-decomposed-from-sys",
      "severity": "violation",
      "message": "Software requirements use the link type 'decomposed from' to system requirements",
      "select": { "$ref": "#/$defs/item-srs" },
      "validate": {
        "network": {
          "decomposed-from": {
            "contains": { "local": { "$ref": "#/$defs/type-sys" } },
            "minContains": 1
          }
        }
      }
    }
  ]
}
```

However, something like the following using the "incoming" notation of the `extra_links` is not possible:

```json
{
  "schemas": [
    {
      "id": "sys-decomposes-to-srd",
      "severity": "violation",
      "message": "System requirements have at least one downstream link 'decomposes to' to software requirements",
      "select": { "$ref": "#/$defs/item-sys" },
      "validate": {
        "network": {
          "decomposes to": {
            "contains": { "local": { "$ref": "#/$defs/type-srs" } },
            "minContains": 1
          }
        }
      }
    }
  ]
}
```

Using `decomposed-from_back` for the link type also doesn't work, so I guess I can only check upstream (outgoing) using `schema.json`.

- I then tried to look at `need_constraints` but those apply globally so I didn't check those further (I need to define my trace types by item type).
- Dynamic functions also don't work (and are not intended for this check) since they can't look at incoming links ([#1457](https://github.com/useblocks/sphinx-needs/issues/1457)).

I guess this could be a separate extension where (separate) schemas can be created for incoming links (downstream checks). E.g., I could check that certain directives follow a schema downstream (incoming), but since top-down reviews are very common in safety-related projects I wanted to just quickly ask if something like this is maybe (not officially) supported or on the roadmap?

For now, I'm using `needtable` with the `_back` notation such that completeness can be verified quite easily using tables, but It would be great to have a schema definition for top-down reviews, e.g., also using more complex need types (I could imagine, e.g., adding another directive "Terminator" for items that are not supposed to have any outgoing links).

chrisjsewell:
## Summary

Adds a `network_back` key to schema validation, as a sibling of the existing `network` key, that validates a need's **incoming** links rather than its **outgoing** links.

> **Note:** this feature is being developed **in parallel with ubCode**, so that the two implementations validate identically (same rule codes, same message wording, same semantics).

## Motivation

The existing `network` key validates outgoing links: *"this need's `links` of type X must point to needs matching schema S, with min/max count constraints"*.

There was no way to assert constraints on **incoming** links. The motivating use case:

> *Every `req` must be covered by at least one `test`.*

Previously this could only be expressed as a forward rule on **every** `test`, which is awkward and easy to leave incomplete. With `network_back` it is expressed once, **on the req**:

```yaml
validate:
  network_back:
    links:               # incoming 'links'
      contains:
        local: { $ref: '#/$defs/test' }
      minContains: 1      # >=1 test links to me
```

## Semantics

`network_back` has the **exact same nested shape** as `network` (`items` / `contains` / `minContains` / `maxContains`, recursively containing `local` + `network` + `network_back`). The only difference is direction:

- `network` keys a link type and walks the need's **outgoing** links of that type (`need[link_type]`).
- `network_back` keys a link type and walks the need's **incoming** links of that type — the needs that link *to* this need via that outgoing type (the already-computed `need[link_type + "_back"]` field).

Details:

- `minContains` defaults to 1; `maxContains` unbounded.
- A back hop and a forward hop both consume one level of the shared recursion budget, so `network` and `network_back` can be freely mixed/nested within the same 4-level chain.
- `network_missing_target` (broken link) is **forward-only** — incoming links are resolved from existing needs and always exist.
- The existing message rules are reused (`network_contains_too_few`, `network_contains_too_many`, `network_items_fail`, `network_local_*`); direction is disambiguated by the `schema_path` containing `network_back`, and message text says "incoming links" for the back direction.
- The `select` predicate still selects the *validated* need (the target of the incoming links), not the sources.

## Changes

- `sphinx_needs/schema/config.py`: `network_back` added to `ValidateSchemaType`.
- `sphinx_needs/schema/core.py`: `recurse_validate_schemas` flattens both directions into one loop; incoming links traverse the `<link_type>_back` field.
- `sphinx_needs/schema/config_utils.py`: `$ref` resolution, type-injection, and link-type checks recurse into `network_back`.
- `sphinx_needs/schema/jsons/SchemasRootType.schema.json`: regenerated to allow `network_back`.
- `docs/schema/index.rst`: new "Incoming link validation (`network_back`)" section + coverage example.
- `tests/schema/fixtures/network.yml` + snapshots: cases for back-contains (too few / too many), back-items, and the coverage example.

## Tests

`tests/schema` passes (173 passed, 234 snapshots). `ruff` and `mypy` clean on the changed files.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,6 +14,13 @@ Unreleased
 Improvements
 ............
 
+- ✨ Add a ``network_back`` schema-validation key, the sibling of ``network``,
+  that validates a need's **incoming** links instead of its outgoing ones. It
+  reuses the same ``items`` / ``contains`` / ``minContains`` / ``maxContains``
+  structure and can be freely mixed and nested with ``network``, allowing
+  constraints like "every requirement must be covered by at least one test" to be
+  expressed once on the requirement (:pr:`1731`)
+
 - 🔧 Add a root ``context7.json`` configuration file so AI assistants using
   Context7 can discover the live Sphinx-Needs documentation and use the
   project-specific reference instead of stale training data (:issue:`1719`)
--- a/docs/schema/index.rst
+++ b/docs/schema/index.rst
@@ -353,7 +353,8 @@ A ``schemas.json`` file (or ``needs_schema_definitions`` dict) has the following
          "select": { /* Selection criteria - which needs to validate */ },
          "validate": {
            "local": { /* Local validation rules */ },
-           "network": { /* Network validation rules */ }
+           "network": { /* Outgoing-link validation rules */ },
+           "network_back": { /* Incoming-link validation rules */ }
          }
        }
      ]
@@ -378,7 +379,9 @@ Each object within the ``schemas`` array can contain:
 - ``validate`` (required): Validation rules with two subsections:
 
   - ``local``: Validates properties of the need itself (see :ref:`local_validation`)
-  - ``network``: Validates relationships with linked needs (see :ref:`network_validation`)
+  - ``network``: Validates outgoing links to other needs (see :ref:`network_validation`)
+  - ``network_back``: Validates incoming links from other needs
+    (see :ref:`network_back_validation`)
 
 See :ref:`schema_components` for detailed documentation of each component.
 
@@ -556,6 +559,57 @@ This validates that:
 1. A safe implementation links to safe specifications
 #. Those specifications in turn link to safe features
 #. Both link levels have minimum count requirements
+
+.. _`network_back_validation`:
+
+Incoming link validation (``network_back``)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, ``network`` validates a need's *outgoing* links.
+Sometimes a constraint is more naturally expressed from the target's point of view —
+for example *"every requirement must be covered by at least one test"*.
+
+The ``network_back`` key mirrors ``network`` but traverses the *incoming* links of a need:
+the needs that link **to** this need via the given outgoing link type
+(i.e. the computed ``<link_type>_back`` field).
+It accepts exactly the same ``items`` / ``contains`` / ``minContains`` / ``maxContains``
+structure (and may itself nest ``local``, ``network`` and ``network_back`` rules).
+
+The coverage example, expressed once on the requirement rather than on every test:
+
+.. code-block:: json
+
+   {
+     "id": "req-covered-by-test",
+     "select": {
+       "properties": { "type": { "const": "req" } }
+     },
+     "validate": {
+       "network_back": {
+         "links": {
+           "contains": {
+             "local": {
+               "properties": { "type": { "const": "test" } }
+             }
+           },
+           "minContains": 1
+         }
+       }
+     }
+   }
+
+Here ``links`` refers to the *outgoing* ``links`` link type of the source needs;
+the rule selects every ``req`` and requires at least one ``test`` to link to it.
+
+.. note::
+
+   - A back hop and a forward hop both consume one level of the shared recursion budget
+     (see the recursion limit below), so ``network`` and ``network_back`` can be freely mixed
+     and nested within the same 4-level chain.
+   - The ``select`` predicate always selects the *validated* need (the target of the incoming
+     links), never the sources.
+   - Broken-target errors (``network_missing_target``) only apply to outgoing links;
+     incoming links are resolved from existing needs and therefore always exist.
 
 .. important::
 
--- a/sphinx_needs/schema/config.py
+++ b/sphinx_needs/schema/config.py
@@ -538,6 +538,15 @@ class ValidateSchemaType(TypedDict):
 
     local: NotRequired[RefItemType | AllOfSchemaType | NeedFieldsSchemaType]
     network: NotRequired[dict[str, ResolvedLinkSchemaType]]
+    """Validate the need's outgoing links, keyed by link type."""
+    network_back: NotRequired[dict[str, ResolvedLinkSchemaType]]
+    """
+    Validate the need's incoming links, keyed by link type.
+
+    Mirrors ``network`` but traverses the needs that link *to* this need
+    via the given outgoing link type (the computed ``<link_type>_back`` field),
+    rather than the need's own outgoing links.
+    """
 
 
 class SchemasRootType(TypedDict):
--- a/sphinx_needs/schema/config_utils.py
+++ b/sphinx_needs/schema/config_utils.py
@@ -165,11 +165,17 @@ def _recursive_validate(validate: ValidateSchemaType, path: tuple[str, ...]) -> 
             raise NeedsConfigException(
                 f"schema for '{path_str}' is not valid:\n{exc}"
             ) from exc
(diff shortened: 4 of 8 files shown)


chrisjsewell:
# Review

PR #1731 adds a network_back schema-validation key (sibling of network) to validate a need's incoming links rather than outgoing ones — enabling constraints like "every req must be covered by ≥1 test" expressed once on the req.

My review process: read the full diff, traced the runtime path in sphinx_needs/schema/core.py, checked the helper changes in config_utils.py/config.py, inspected fixtures/snapshots/docs/JSON schema, and verified locally on the PR head (69dd352):

- ✅ tests/schema: 175 passed, 236 snapshots
- ✅ ruff and mypy clean on the 3 changed source files
- ✅ Confirmed no KeyError risk (_back fields default to [])

Key findings:

1. need_path arrow direction is misleading for incoming links — it renders REQ_1 > links > IMPL_1 when reality is IMPL_1 --links--> REQ_1 (main concern).
2. Dead/untested broken-link branch for the back direction (corroborated by the codecov "1 line missing" and my --cov run pointing at lines 341–364).
3. Fragile link_desc[:-1] singularization.
4. Missing changelog entry (repo PR checklist requires one).
5. Test gap for the "freely mixed/nested" claim — no fixture nests network_back inside network or exercises the shared recursion budget across directions.

I suggest #1 and #4 as the ones to resolve before merge, with the rest as maintainability improvements, and note the overall design (reusing the forward-link machinery) is clean.

chrisjsewell:
Thanks — addressed in `8ded2ce0`.

**1. `need_path` arrow direction (main concern)** ✅ — incoming hops are now annotated as `links (incoming)`, so a back hop renders e.g. `REQ_1 > links (incoming) > IMPL_1` (and, nested, `IMPL_1 > links > SPEC_1 > links (incoming) > IMPL_1`), making it clear the incoming source links *to* the need rather than the reverse. (Mirrored in ubCode for parity.)

**4. Changelog** ✅ — added an entry under Unreleased › Improvements (`:pr:`1731``).

**3. `link_desc[:-1]` singularization** ✅ — replaced with an explicit `link_desc_singular`.

**5. Mixed/nested test gap** ✅ — added `network_back_nested_in_network_{ok,error}` fixtures: a forward `network` whose `contains.network_back` requires the linked spec to have ≥1 incoming test, exercising a forward-then-back traversal across the shared recursion budget (with passing and failing variants).

**2. Dead broken-link branch for the back direction** — left in place intentionally: it's the shared broken-link handling used by both directions. For incoming links the sources come from resolved `<link>_back` fields so a `KeyError` can't occur (hence the coverage gap), but the branch is still reachable and necessary for the forward direction; special-casing it out for `network_back` would add a branch for no behavioural change. Happy to add a `# pragma: no cover`-style note on the back path if you'd prefer the coverage signal cleaned up.

`tests/schema` (175 passed, 238 snapshots), ruff and mypy all clean.

chrisjsewell (commented):
## Follow-up review (head `8ded2ce`)

Re-reviewed after the changes addressing the [previous review](https://github.com/useblocks/sphinx-needs/pull/1731#issuecomment-4807218162). Re-read the diff, re-traced the runtime path in `core.py`, and re-verified locally on the PR head:

- ✅ `tests/schema`: 163 passed, 8 skipped, 220 snapshots (incl. the new `network_back_*` fixtures)
- ✅ `ruff` and `mypy` clean on the changed source files
- ✅ Confirmed no `KeyError` risk on the `<link_type>_back` access — `NeedItem` initialises backlinks to `[]` for every link type (`need_item.py:489`), and `check_network_links_against_links` now also rejects unknown `network_back` link types.

### Prior findings — all resolved

1. **`need_path` direction** ✅ — incoming hops now render `… > links (incoming) > IMPL_1`, which is unambiguous. Verified against the new snapshots, including the nested case `IMPL_1 > links > SPEC_1 > links (incoming) > IMPL_1`.
2. **Dead broken-link branch** — acknowledged and left as shared forward/back handling; reasonable. (See minor note below.)
3. **`link_desc[:-1]`** ✅ — replaced with an explicit `link_desc_singular`.
4. **Changelog** ✅ — entry added under Unreleased › Improvements.
5. **Mixed/nested test gap** ✅ — `network_back_nested_in_network_{ok,error}` exercise a forward→back traversal through the shared recursion budget (`recurse_level + 1` is applied identically in both directions). Good coverage.

The flattening of `network`/`network_back` into a single `network_entries` loop is clean and keeps both directions on exactly one code path.

### Minor (non-blocking)

- **`network_items_fail` wording for the back direction.** `link_desc` is now `"incoming links"`, but the items-fail message keeps the outgoing-sounding preposition: `Items validation failed for incoming links of type 'links' to TEST_1`. The `to` reads as outgoing even though these are sources linking *in*. The `contains_too_few/too_many` messages avoid this (they use `ok:`/`nok:` lists). Since the `need_path` now disambiguates anyway this is cosmetic — but if you want full parity, `… incoming links of type 'links' from TEST_1` would match the direction. (Worth mirroring in ubCode if changed.)
- The `network_missing_target` (broken-link) branch is unreachable for `network_back` — as you note, sources come from resolved `_back` fields. No action needed; a one-line `# back links always resolve` note at the `except KeyError` would document why the back path can't reach it, if you'd like the coverage gap explained inline.

Design (reusing the forward-link machinery) is clean and the semantics are well documented. LGTM modulo the cosmetic wording note.

---

chrisjsewell:
Thanks! Both minor notes addressed in `d779594c`:

- **Items-fail preposition** — the message now uses a direction-aware preposition: `"to"` for outgoing links, `"from"` for incoming, so `network_back` reads `… incoming links of type 'links' from TEST_1`. (Worth noting: the items-fail *wrapper* warning is only ever surfaced via its children — `warnings.extend(items_nok_warnings)` — so this message string isn't currently emitted in the warning output; the fix keeps it correct for the JSON report and any future surfacing. Pre-existing behaviour, left as-is.)
- **Broken-link branch** — added an inline note at the `except KeyError` explaining it's unreachable for incoming links (sources come from resolved `<link_type>_back` fields, which always resolve).

Both mirrored in ubCode for parity. `tests/schema`, ruff and mypy all clean.

ubmarco (approved):

PR #1734, opened 2026-06-30
🧪 Add network-local reduction parity fixtures

chrisjsewell:
Adds two schema-validation fixtures that pin sphinx-needs' behaviour for a network-context `local` schema using `unevaluatedProperties`, to serve as the parity oracle for the ubcode port.

- **`network_local_unevaluated_reduced_ok`** — a linked spec with no extra fields, reduced to just `{type: spec}`, validates cleanly. Confirms `_validate_need_local` applies `reduce_need` inside the network path exactly as for top-level local validation.
- **`network_local_unevaluated_reduced_error`** — the linked spec sets an extra `asil` field the schema does not reference; reduction keeps actively-set extra fields, so `unevaluatedProperties: false` still fails. Confirms reduction does not mask real violations.

No behaviour change — fixtures + snapshots only.

### DIFF
--- a/tests/schema/__snapshots__/test_schema.ambr
+++ b/tests/schema/__snapshots__/test_schema.ambr
@@ -5090,6 +5090,48 @@
     }),
   })
 # ---
+# name: test_schemas[schema/fixtures/network-network_local_unevaluated_reduced_error]
+  '''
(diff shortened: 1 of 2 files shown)


ubmarco (approved):
What would I have to say against more testing?!

PR #1738, opened 2026-07-01
👌 Render needs_role_need_template with Jinja

chrisjsewell:
Implements #1697 and supersedes #1698, rebased on the current `master` (the #1698 branch is on a fork and behind master, so it couldn't be updated directly).

`needs_role_need_template` and the inline `:need:`[[...]] <ID>`` role text are now rendered with **Jinja** (via MiniJinja) instead of Python's `str.format`.

### What changed
- **Jinja rendering** for `needs_role_need_template` and the inline `[[...]]` role text. Filters and control structures (`{% if %}`, `| upper`, …) are now supported.
- **Applies the review suggestion from #1698** ([`#discussion_r3174125896`](https://github.com/useblocks/sphinx-needs/pull/1698#discussion_r3174125896)): instead of rewriting `[[`/`]]` → `{{`/`}}` with `str.replace`, the inline template is compiled with `[[` / `]]` configured as MiniJinja's variable delimiters. `sphinx_needs/_jinja.py`'s `compile_template` / `render_template_string` (and the env helpers) now accept `variable_start_string` / `variable_end_string`, and the env cache is keyed on them.
- **New template variables**: `id_complete`, `id_parent`, `id_part`, `is_need`, `is_part`, plus a `need` object exposing the same fields (e.g. `{{ need.type }}`).
- **Robustness**: an invalid template warns once and falls back to `title (id)` instead of breaking the build.
- **Docs + changelog**: migrated `{field}` → `{{ field }}` examples, documented the new variables, and added a changelog + `versionchanged` note.

### How breaking is this?
Only users with a **custom** `needs_role_need_template` (or inline `[[...]]` using format-specs) are affected — anyone on the default is unaffected, since the default was migrated too. The risky part is that, without mitigation, an old `str.format` template would produce **silent wrong output**, not a crash:

| Existing template | Old output (`str.format`) | New output (Jinja, no shim) |
|---|---|---|
| `"[{id}] {title}"` | `[REQ_1] My title` | `[{id}] {title}` ← literal braces |
| `"{title:*^20s}"` | `*****My title******` | `{title:*^20s}` ← literal |
| `"{content:.30}"` | first 30 chars of content | `{content:.30}` ← literal |
| `"{{literal}}"` (escaped brace) | `{literal}` | `` ← parsed as Jinja var, empty |

### Mitigation — backward-compat shim
To avoid that silent breakage, templates that use the legacy `str.format` syntax (a `{field}` reference with no Jinja `{{ }}` / `{% %}` / `{# #}` markers) are **still rendered with `str.format`**, and a one-time `needs.deprecated` warning is emitted pointing at the migration. So existing configs keep working unchanged; users just get a clear, actionable warning. The old syntax is documented as deprecated and will be removed in a future release. (`{{literal}}` — the escaped-brace form — is inherently ambiguous with Jinja and is treated as Jinja; this is an accepted edge case.)

### Tests
- `tests/test_role_need_template.py`: control structures, filters, need parts, inline `[[...]]`; a legacy-template fixture asserting old syntax still renders **and** warns; an invalid-template fallback fixture; and unit tests for the legacy-detection heuristic.
- New `tests/test_jinja.py`: unit tests for the delimiter support (default vs `[[`/`]]`, cache keyed on delimiters, block syntax preserved, invalid syntax raises).


PhilipPartsch (approved):
Looks wonderful, Thanks @chrisjsewell

chrisjsewell:
> Looks wonderful, Thanks @chrisjsewell 

Sorry I didn't put it into the latest release @PhilipPartsch , because me and @ubmarco were a little cautious on it being breaking to users. So just want to think on it a little more first (but I hope it is close)

ubmarco:
## Backwards-compatibility analysis

Since the concern with this change has always been *silent* breakage of existing
`needs_role_need_template` / inline `:need:`[[…]] <ID>`` configs, here's a full
accounting of who is affected and how loudly. (Examples assume a need with
`title = "My title"`, `id = "REQ_1"`.)

### Config template — `needs_role_need_template`

The render path is chosen by `_is_legacy_format_template`: a `{field}` reference with
**no** Jinja markers (`{{`, `{%`, `{#`) is rendered the old way with `str.format` (plus a
one-time `needs.deprecated` warning); everything else goes through Jinja.

| Template | Old (`str.format`) | New | Verdict |
|---|---|---|---|
| *default* — `{title} ({id})`, migrated to `{{ title }} ({{ id }})` | `My title (REQ_1)` | `My title (REQ_1)` | ✅ identical, no warning |
| `"[{id}] {title}"` (typical custom) | `[REQ_1] My title` | `[REQ_1] My title` (shim) | ✅ identical + deprecation warning |
| `"{title:.4}"` (format spec) | `My t` | `My t` (shim) | ✅ identical + warning |
| `"{{literal}}"` (str.format escaped brace) | `{literal}` | `` (empty) | ❌ **silent** — `{{` reads as a Jinja var |

So the common case — custom `{field}` templates, with or without format specs — keeps
working unchanged; the only *silent* break is the escaped-brace `{{ … }}` form, which the
PR description already flags as an accepted edge case. (Users who explicitly set the *old
default* string also get a harmless deprecation warning.)

### Inline role text — `:need:`[[…]] <ID>``

This path is **Jinja-only** — there is no legacy `str.format` fallback here, unlike the
config template.

| Inline text | Old | New | Verdict |
|---|---|---|---|
| `[[title]]` | `My title` | `My title` | ✅ |
| `[[ title ]]` (spaces) | KeyError → warn, empty | `My title` | ✅ improved |
| `[[title:.4]]` (format spec) | `My t` | parse error → warn, **empty** | ❌ breaks (warned) |
| `[[missing_field]]` | KeyError → warn, empty | undefined → **silent** empty | ⚠️ minor |

### Net assessment

- The mitigation covers the realistic majority: custom `{field}` config templates render
  byte-identically, with a clear one-time deprecation warning pointing at the migration.
- Residual breakage is **narrow**: escaped-brace config templates (silent, acknowledged)
  and inline format specs (warned). Both are uncommon.
- Two asymmetries worth a line of docs (not blockers):
  1. The inline path has no legacy shim, so `str.format` semantics that survive on the
     config path break inline. Either mirror the shim or document inline as Jinja-only.
  2. The deprecation guidance "migrate `{field}` → `{{ field }}`" isn't a valid conversion
     for format specs (`{title:.4}` → `{{ title:.4 }}` is invalid Jinja). Worth pointing
     format-spec users at Jinja filters (`| truncate`, `| center`, `format(...)`).

Overall the silent-breakage risk that kept this out of 8.2.0 looks well-contained. 👍

<sub>Branch updated to current `master` (only `docs/changelog.rst` conflicted, resolved
into a new *Unreleased* section); the `versionchanged` / `deprecated` markers now read
`unreleased` instead of `8.2.0`.</sub>

ubmarco (approved):

PR #1740, opened 2026-07-06
Feature/external sources in needsjson

cpolzer:
## Problem

When needs_builder_filter is relaxed to include external needs in needs.json, downstream consumers lose the provenance metadata. (information on the configured external needs in a producer -> links can for consumers not be generated anymore).

This probably is not an easy thing to merge - i would guess other tooling handling external needs would need to adapt first or follow?

The extended example is:

1.) I have a sphinx-needs "Project A" consuming external needs and publishing them to needs.json
2.) "Project B" consumes  "Project A" needs.json -> the links from project A cannot be rebuild because the external needs configuration of "Project A" is not known. 

## Solution

- external_sources list in the version block
- (and/or) external_source field on each external need
- Transitive inheritance from consumed JSON files

## Testing

tox -e py312 -- tests/test_external.py tests/test_needs_external_needs_build.py

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,28 @@
 Changelog
 =========
 
+Unreleased
+----------
+
+External sources provenance in needs.json
+.........................................
+
+When external needs are included in the ``needs.json`` output (by relaxing
+:ref:`needs_builder_filter`), the JSON now contains an ``external_sources``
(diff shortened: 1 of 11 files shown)


cpolzer:
Further thinking about it. Maybe we just want to store a base url on needs_builder building the needs.json that can be used to automatically configure the external needs...

PR #1750, opened 2026-07-30
🐛 Use the physical source directory for generated PlantUML diagrams

### LINKED ISSUE #1749: Use physical source directory for needuml PlantUML working directory
## Summary

`needuml` diagrams fail to render when their source document is provided by a mounted documentation tree. `sphinx-needs` stores a logical Sphinx docname in `incdir`, but `sphinxcontrib.plantuml` uses that value as its working directory. For mounted documents, that logical directory does not exist on disk.

The resulting warning is misleading because the PlantUML executable itself is available:

```
WARNING: plantuml command '…/src/plantuml' cannot be run [plantuml]
```

## Environment

- Sphinx 9.1.0
- sphinx-needs 8.0.0
- sphinxcontrib-plantuml 0.31
- sphinx-mounts 0.1.0
- Bazel runfiles-based documentation build on Linux

## Steps to reproduce

1. Build a Sphinx project using `sphinx-mounts` (or an equivalent extension) so that a mounted document has a logical docname such as `logging/features/example/index`, while its physical source file is elsewhere.
2. Add a `.. needuml::` directive to that mounted document.
3. Configure `sphinxcontrib.plantuml` with a working PlantUML executable and build HTML.

## Actual behavior

During `process_needuml`, the generated PlantUML node gets:

```python
puml_node["incdir"] = os.path.dirname(current_needuml["docname"])
```

For a mounted document this becomes a non-existent logical path, e.g. `logging/features/example`. `sphinxcontrib.plantuml` passes it as `cwd` to `subprocess.Popen`; that raises `ENOENT`, which is reported as if the PlantUML command were missing.

## Why this belongs in sphinx-needs

`sphinx-mounts` registers the mounted document with Sphinx correctly: `env.doc2path(docname)` resolves to the actual source file in the mounted/runfiles tree. The path is lost later, when `sphinx-needs` creates its generated PlantUML node and derives `incdir` from the logical `docname` instead of that resolved path. Therefore the fix belongs where that node is constructed, in `sphinx-needs`.

## Expected behavior

Generated `needuml` nodes should use the physical directory of the source document as `incdir`, so that PlantUML can resolve relative `!include` paths and start successfully.

## Suggested change

Use Sphinx's resolved document path when setting `incdir`, for example:

```python
source_path = env.doc2path(current_needuml["docname"])
puml_node["incdir"] = os.path.dirname(source_path)
```

This preserves the existing behavior for ordinary documents and makes mounted documentation work correctly.

## Workaround and verification

In the S-CORE docs-as-code integration we added a post-`doctree-resolved` hook that replaces `incdir` with `Path(node.source).parent` for generated PlantUML nodes. The PlantUML warnings disappeared after applying this workaround.

---

_This issue was generated by GPT-5.6 Terra from analyses in the S-CORE project._

ubmarco:
Closes #1749

## Problem

`sphinxcontrib-plantuml` starts the PlantUML process with
`cwd = os.path.join(srcdir, node["incdir"])`, so `incdir` decides where relative
`!include` paths are resolved.

Sphinx-Needs derived `incdir` from the **logical docname**:

```python
puml_node["incdir"] = os.path.dirname(current_needuml["docname"])
```

For a document whose source file does not physically live under `srcdir` — e.g.
one contributed by [sphinx-mounts](https://github.com/useblocks/sphinx-mounts),
which registers an absolute external path for its docname — that yields a
directory that does not exist. `subprocess.Popen` fails with `ENOENT`, which
`sphinxcontrib-plantuml` reports as:

```
WARNING: plantuml command '…/plantuml' cannot be run [plantuml]
```

The message is misleading: the PlantUML executable is fine, the `cwd` is not.

## Fix

Derive `incdir` from `env.doc2path(docname, base=False)` instead — the same
thing `sphinxcontrib-plantuml`'s own `uml` directive does — through a new shared
`set_plantuml_paths()` helper in `diagrams_common.py`.

The bug was present at **four** sites, all now routed through the helper:

| directive | file |
| --- | --- |
| `needuml` / `needarch` | `directives/needuml.py` |
| `needflow` (plantuml engine) | `directives/needflow/_plantuml.py` |
| `needsequence` | `directives/needsequence.py` |
| `needgantt` | `directives/needgantt.py` |

### Why `base=False` and not an absolute path

`doc2path(..., base=False)` returns a **srcdir-relative** path for an ordinary
document and the **absolute external** path for a mounted one. Since
`os.path.join` discards its left operand when the right one is absolute, PlantUML
ends up with the correct `cwd` in both cases.

Picking the relative form matters beyond taste:

- Ordinary documents get a **byte-identical** `incdir` to before
  (`dirname("a/b/c")` == `dirname("a/b/c.rst")`), so nothing changes for existing
  projects.
- `incdir` feeds `hash_plantuml_node()`, i.e. PlantUML's content-addressed cache
  key. Always emitting an absolute path would invalidate every cached diagram and
  make the cache non-portable across machines and checkout locations. This way
  only genuinely-mounted documents get an absolute key.

## Test

`tests/test_plantuml_incdir.py` builds a host project plus a sibling bundle
mounted via a ~20-line stand-in for sphinx-mounts (it reproduces the one trick
that matters — an absolute path in `Project._docname_to_path` — so the test
carries no new dependency), and asserts:

- an ordinary nested host document keeps `incdir == "sub"`;
- the mounted document gets the bundle's physical directory, not `"mounted"`;
- PlantUML actually runs and resolves a bundle-relative `!include`, with zero
  build warnings.

Verified failing before the fix (`['mounted', 'mounted'] != [<bundle>, <bundle>]`)
and passing after.

Full suite: 1018 passed. The two `tests/test_sn_collapse_button.py` failures on
this branch reproduce identically on `master` and are unrelated.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,6 +7,18 @@ Changelog
 Unreleased
 ----------
 
+- 🐛 Generated PlantUML diagrams (:ref:`needuml`, :ref:`needarch`,
+  :ref:`needflow`, :ref:`needsequence`, :ref:`needgantt`) now derive PlantUML's
+  working directory from the *physical* path of the source document instead of
+  its logical docname (:issue:`1749`).
+
+  Relative ``!include`` paths therefore also resolve for documents whose source
+  file does not live under ``srcdir`` — e.g. documents contributed by
+  `sphinx-mounts <https://github.com/useblocks/sphinx-mounts>`__ — which
+  previously failed with a misleading
+  ``WARNING: plantuml command '...' cannot be run``.
+  Ordinary documents keep the exact working directory they had before.
+
 - 📚 The documentation now uses the shared
   `sphinx-syntax-example <https://github.com/sphinx-extensions2/sphinx-syntax-example>`__
   ``syntax-example`` directive, in place of the bespoke ``need-example`` directive
--- a/sphinx_needs/diagrams_common.py
+++ b/sphinx_needs/diagrams_common.py
@@ -14,6 +14,7 @@ from urllib.parse import urlparse
 from docutils import nodes
 from docutils.parsers.rst import directives
 from sphinx.application import Sphinx
+from sphinx.environment import BuildEnvironment
 from sphinx.util.docutils import SphinxDirective
 
 from sphinx_needs.config import NeedsSphinxConfig, NeedType
@@ -108,6 +109,30 @@ def no_plantuml(node: nodes.Element) -> None:
     para += text
     content.append(para)
     node.replace_self(content)
+
+
+def set_plantuml_paths(
+    puml_node: nodes.Element, env: BuildEnvironment, docname: str
+) -> None:
+    """Point a generated PlantUML node at the physical location of ``docname``.
+
+    ``sphinxcontrib-plantuml`` runs the PlantUML process with
+    ``cwd = os.path.join(srcdir, node["incdir"])``, so ``incdir`` decides where
+    relative ``!include`` paths are resolved. Deriving it from the *logical*
+    docname breaks as soon as the document does not physically live under
+    ``srcdir`` -- e.g. a document contributed by ``sphinx-mounts``, whose source
+    file stays in an external tree. PlantUML then gets a non-existent ``cwd``,
+    which surfaces as the misleading "plantuml command cannot be run".
+
+    ``env.doc2path(docname, base=False)`` returns a srcdir-relative path for an
+    ordinary document and the absolute external path for a mounted one. Since
+    ``os.path.join`` discards its left operand when the right one is absolute,
+    both cases resolve correctly -- and ordinary documents keep the exact
+    ``incdir`` value they had before, so PlantUML's content-addressed cache
+    stays valid (and machine-independent) for them.
+    """
+    puml_node["incdir"] = os.path.dirname(env.doc2path(docname, base=False))
+    puml_node["filename"] = os.path.split(docname)[1]  # Needed for plantuml >= 0.9
 
 
 def add_config(config: str) -> str:
--- a/sphinx_needs/directives/needflow/_plantuml.py
+++ b/sphinx_needs/directives/needflow/_plantuml.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import html
-import os
 
 from docutils import nodes
 from sphinx.application import Sphinx
@@ -10,7 +9,11 @@ from sphinx_needs._jinja import render_template_string
 from sphinx_needs.config import NeedsSphinxConfig
 from sphinx_needs.data import NeedsFlowType, SphinxNeedsData
 from sphinx_needs.debug import measure_time
-from sphinx_needs.diagrams_common import calculate_link, create_legend
+from sphinx_needs.diagrams_common import (
+    calculate_link,
+    create_legend,
+    set_plantuml_paths,
+)
 from sphinx_needs.directives.needflow._directive import NeedflowPlantuml
 from sphinx_needs.directives.utils import no_needs_found_paragraph
 from sphinx_needs.filter_common import filter_single_need, process_filters
@@ -319,10 +322,7 @@ def process_needflow_plantuml(
                 puml_node["uml"] += create_legend(needs_config.types)
 
             puml_node["uml"] += "\n@enduml"
-            puml_node["incdir"] = os.path.dirname(current_needflow["docname"])
-            puml_node["filename"] = os.path.split(current_needflow["docname"])[
-                1
-            ]  # Needed for plantuml >= 0.9
+            set_plantuml_paths(puml_node, env, current_needflow["docname"])
 
             scale = int(current_needflow["scale"])
             # if scale != 100:
--- a/sphinx_needs/directives/needgantt.py
+++ b/sphinx_needs/directives/needgantt.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import os
 import re
 from collections.abc import Sequence
 from datetime import datetime
@@ -18,6 +17,7 @@ from sphinx_needs.diagrams_common import (
     get_debug_container,
     get_filter_para,
     no_plantuml,
+    set_plantuml_paths,
 )
 from sphinx_needs.directives.utils import (
     SphinxNeedsLinkTypeException,
@@ -351,10 +351,7 @@ def process_needgantt(
             puml_node["uml"] += create_legend(needs_config.types)
 
         puml_node["uml"] += "\n@endgantt"
-        puml_node["incdir"] = os.path.dirname(current_needgantt["docname"])
-        puml_node["filename"] = os.path.split(current_needgantt["docname"])[
-            1
-        ]  # Needed for plantuml >= 0.9
+        set_plantuml_paths(puml_node, env, current_needgantt["docname"])
 
         scale = int(current_needgantt["scale"])
         # if scale != 100:
--- a/sphinx_needs/directives/needsequence.py
+++ b/sphinx_needs/directives/needsequence.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import os
 import re
 from collections.abc import Sequence
 from typing import Any
@@ -18,6 +17,7 @@ from sphinx_needs.diagrams_common import (
     get_debug_container,
     get_filter_para,
     no_plantuml,
+    set_plantuml_paths,
 )
 from sphinx_needs.directives.utils import no_needs_found_paragraph
 from sphinx_needs.filter_common import FilterBase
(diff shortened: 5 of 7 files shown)


chrisjsewell (approved):
haven't checked through the actual code, but approve in principle

ubmarco:
## Review

Right root cause, right API. Deriving `incdir` from `env.doc2path(docname, base=False)` is exactly what `sphinxcontrib-plantuml`'s own `uml` directive does (`plantuml.py:140-145`), and collapsing four copy-pasted sites into one helper is the correct shape. I verified the `base=False` reasoning against Sphinx's `Project.doc2path`: with a falsy `base` no `srcdir` join happens, and `Project.discover` stores paths through `path_stabilize`/`canon_path`, so stored paths are forward-slash even on Windows — the "byte-identical for ordinary documents" claim holds on all platforms, which the green Windows/Sphinx 9.1 jobs confirm. All four call sites pass the correct `docname` key. Nothing here blocks merge.

### 1. [Medium] Two of the four refactored call sites are never exercised (correctness / test coverage)

The PR's stated value is that the bug existed at four sites and all four now route through the helper — but the fixture document only contains `needuml` and `needflow`, so `needsequence.py:194` and `needgantt.py:354` never execute in the new test. `needarch` is covered by proxy (it shares the unconditional line in `process_needuml`).

https://github.com/useblocks/sphinx-needs/blob/b8a550221860e15efb6f1228bc3e318774eb7775/tests/test_plantuml_incdir.py#L138-L145

I read all four sites and they do pass the right `docname`, so there is no present bug — the gap is regression protection for the two untested sites. Adding a `needsequence` and a `needgantt` directive to `BUNDLE_INDEX` (and extending the expected `incdirs["mounted/index"]` list) closes it cheaply.

### 2. [Medium] Mounted documents get a location-dependent PlantUML cache key (performance)

`incdir` is hashed into PlantUML's content-addressed key (`hash_plantuml_node`, `plantuml.py:179-185`), and that key becomes the `_images/plantuml-<key>.svg` filename. For mounted documents `incdir` is now an absolute path, so any relocation of the source tree changes the key — which is precisely the Bazel runfiles/sandbox setup from #1749. Result: the `plantuml_cache` never hits and every mounted diagram re-renders (one JVM per diagram) after a relocation, while `_images` accumulates orphans.

https://github.com/useblocks/sphinx-needs/blob/b8a550221860e15efb6f1228bc3e318774eb7775/sphinx_needs/diagrams_common.py#L133-L135

The docstring acknowledges the non-portability but frames it as acceptable because it only affects mounted documents — which are exactly the documents this PR exists to fix. `os.path.relpath(dirname(physical), env.srcdir)` yields the identical `cwd` (PlantUML joins it onto `srcdir` and the OS resolves the `..`) with a relocation-stable key; I checked that a runfiles root moving from `/build/...` to `/tmp/sandbox-9876/...` gives a byte-identical relative `incdir`. It needs a `try/except ValueError` fallback to the absolute path for cross-drive mounts on Windows. Worth considering, but a follow-up is fine.

### 3. [Low] The mount stand-in stores `str` where Sphinx and `sphinx-mounts` store `Path` (durability)

From Sphinx 8.2 on, `Project._docname_to_path`/`_path_to_docname` hold `Path` objects, and the real `sphinx-mounts` 0.1.2 registers `Path` too. The test passes on 7.4/8.2/9.1 regardless, because `doc2path` re-wraps whatever it finds — but the docstring claims to reproduce "the one trick that matters", and this is the one detail where it doesn't. A `str` key in `_path_to_docname` would silently miss a `Path`-keyed `path2doc()` lookup if the test is ever extended (e.g. an image referenced from the mounted file). Dropping the `str()` calls costs nothing.

https://github.com/useblocks/sphinx-needs/blob/b8a550221860e15efb6f1228bc3e318774eb7775/tests/test_plantuml_incdir.py#L54-L57

### 4. [Low] Blanket zero-warning assertion widens the test's blast radius (durability)

Every other use of `get_warnings_list` in the suite asserts a specific expected list; this is the only blanket `== []`. It ties "did the incdir fix work" to "zero warnings anywhere in this build", so an unrelated future warning fails this test with a diff that points at the wrong culprit. Asserting no `plantuml`-typed warning would isolate the intent.

https://github.com/useblocks/sphinx-needs/blob/b8a550221860e15efb6f1228bc3e318774eb7775/tests/test_plantuml_incdir.py#L193-L196

Minor: the helper computes the physical path but still derives `filename` from the logical docname, and the docstring — thorough on `incdir` — never mentions `filename`. Keeping the value is right (changing it would alter the `-filename` label), but a half-sentence saying so would help the next reader.

### Other dimensions

- **Backwards compatibility** — no break for ordinary projects; `incdir`, and therefore every existing cache key, is unchanged. Docnames absent from the project fall back to `docname + suffix`, whose `dirname` also matches the old value.
- **Architecture** — fits existing patterns; `env.doc2path` is already used at all four sites for `puml_node.source` (added in #982), and #1447 set the same precedent of replacing hand-rolled docname math with a Sphinx path API.
- **History** — I looked for a commit that deliberately chose the logical docname over a physical path; there is none, so this is not reintroducing an old fix. The `# Needed for plantuml >= 0.9` comment (from `9731de1b`) guards the `filename` line, which is untouched.
- **Past reviews** — the only prior review is an approval in principle with the code unread, so there are no outstanding comments.

### Verified locally

`tests/test_plantuml_incdir.py` passes at `b8a5502`; reverting just the `incdir` line reproduces `['mounted', 'mounted'] != [<bundle>, <bundle>]`, so the test genuinely pins the fix. Full CI is green across the 7.4/8.2/9.1 × Linux/Windows matrix.

ubmarco:
## Follow-up on the review findings

Addressed in e8d0791d41e0e2e746263bf7e6a836b744624a48.

### 1. [Medium] Untested call sites — fixed

The mounted bundle document now exercises all four directives that route through `set_plantuml_paths()`, so `needsequence` and `needgantt` are no longer dark. Reverting the `incdir` line now fails with four wrong entries instead of two.

Two fixture details were needed to get there, both worth recording because they are easy to trip over again:

- `process_needsequence` throws its PlantUML node away and substitutes "no needs found" unless it can draw at least one connection, and a connection needs a full sender → message → receiver chain over the traversed link type — a two-need chain silently yields no node at all. The single `MOUNTED_REQ` is therefore now a three-need chain.
- `needgantt` warns for every need without `:duration:`, which the zero-warning assertion below would flag, so all three needs carry one.

### 3. [Low] `str` vs `Path` in the mount stand-in — withdrawn, my finding was wrong

I recommended dropping the `str()` calls so the stand-in stores `Path` like Sphinx ≥8.2 and the real `sphinx-mounts` do. That breaks the build on Sphinx 7.4:

```
TypeError: 'PosixPath' object is not subscriptable
  sphinx/builders/html/__init__.py:612, in get_doc_context
    source_suffix = self.env.doc2path(docname, False)[len(docname):]
```

Sphinx does not store a plain `Path` there — it stores `_StrPath`, a `Path` subclass that deliberately keeps `__getitem__`/`__len__` precisely because its own HTML builder slices the value to recover the source suffix. `str` is the one type that works across the whole supported `sphinx>=7.4,<10` range, so the original code was right. I left the `str()` calls alone and added a comment explaining why, so the next reader does not "modernise" it back.

### 4. [Low] Blanket zero-warning assertion — keeping it, by design

Keeping `assert get_warnings_list(app) == []` as-is. Failing on *any* new warning is the point: it forces someone to look, which is more valuable than the narrower assertion I suggested. Noting it here so the intent is on the record rather than looking like an oversight.

### 2. [Medium] Cache-key churn for mounted documents — withdrawn, keeping the absolute path

I proposed rewriting only the absolute (mounted) case into a srcdir-relative `incdir`, so PlantUML's content-addressed key would survive relocation:

```python
incdir = os.path.dirname(env.doc2path(docname, base=False))
if os.path.isabs(incdir):                      # mounted document
    try:
        incdir = os.path.relpath(incdir, env.srcdir)
    except ValueError:                          # different drive on Windows
        pass
```

It produces a stable key, and `os.path.join(srcdir, "../bundle")` does name the right directory. It must still not be merged: `..` handed to `chdir` is resolved by the kernel against the *real* directory tree, not lexically, so when `srcdir` is reached through a symlink the `..` escapes to the symlink's true parent:

```
srcdir  = .../slink/srcdir   (symlink -> real/docs)
bundle  = .../slink/bundle
relpath = ../bundle
chdir(".../slink/srcdir/../bundle")  ->  FileNotFoundError [Errno 2]
```

That is the same ENOENT as #1749, reintroduced in exactly the kind of symlink forest (Bazel runfiles) this PR exists to fix. `os.path.realpath` on either side does not help, because `sphinxcontrib-plantuml` joins onto its own unresolved `builder.srcdir`.

So the absolute `incdir` on this branch is the correct trade-off: correctness under symlinks beats cache portability. The residual impact is also narrower than the review implied — it only costs anything when the PlantUML cache directory is persisted across builds *while* the source path changes. A proper fix belongs upstream, by letting a PlantUML node carry an explicit cache key instead of hashing its working directory.

---

Full suite on this branch: 1018 passed, 13 skipped, 0 failed.

PR #1756, opened 2026-08-11
🐛 Fix PyPI release workflow (flit/flit_core version mismatch)

ubmarco:
## Summary

Fixes the failed PyPI release (run [31507560145](https://github.com/useblocks/sphinx-needs/actions/runs/31507560145)) that crashed with:

```
TypeError: SdistBuilder.build() got an unexpected keyword argument 'gen_setup_py'
```

## Root cause

The workflow pinned `flit~=3.4`, which resolved to `flit 3.12.0`. That version declares `flit_core>=3.12.0` with no upper bound, so pip installed the newly released `flit_core 4.0.2`, which removed the deprecated `gen_setup_py` argument from `SdistBuilder.build()`. The `flit 3.12.0` CLI still passes it, crashing during the sdist build step of `flit publish`.

Pinning is not what broke this — half-pinning is. `flit` and `flit_core` are one unit, and bounding only one of them lets the halves move apart.

## Change

- **Pin both distributions**: `pip install flit==3.12.0 flit_core==3.12.0`. That is the pair which built 8.3.0, so the release is unblocked without changing the build backend under the hood.
- **`flit publish --use-vcs`**: the default on flit 3, but flit 4 flipped it to `--no-use-vcs`. Stating it explicitly keeps `tests/` and `docs/` in the sdist across a later flit 4 upgrade.
- **Tag filter** widened to `[0-9]+.[0-9]+.[0-9]+` so major versions >= 10 still match.

Installing an unpinned `flit` was the first attempt and was dropped ([rationale](https://github.com/useblocks/sphinx-needs/pull/1756#issuecomment-5257714965)): flit 4.0.2 declares an equally unbounded `flit_core>=4.0.2`, so the same skew can recur, and moving to flit 4 silently changes the published artifacts — the sdist loses `tests/` and `docs/` (1326 → 703 files, 27.7 MB → 2.5 MB), and the wheel jumps to `Metadata-Version: 2.5` with `AUTHORS` reclassified as a license file.

## Verification

- Reproduced the crash locally with `flit 3.12.0` + `flit_core 4.0.2` in a fresh venv.
- Built this branch in a clean clone with the pinned pair and diffed against the published 8.3.0 artifacts: sdist 1326 entries (the only deltas are real repo changes since 8.3.0), wheel `Metadata-Version: 2.4`, `License-File: LICENSE` only — shape parity with the last good release.
- `twine check` passes on both artifacts.

## Releasing 8.3.1

Merging this does not publish 8.3.1 on its own: `refs/tags/8.3.1` points at the pre-fix commit, and re-running the failed job replays `release.yaml` as of that commit. Nothing was ever published under 8.3.1 (PyPI is still at 8.3.0), so after merge the 8.3.1 release and tag are deleted, the merged commit is re-tagged, and pushing that tag runs the fixed workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Release
 on:
   push:
     tags:
-      - '[0-9].[0-9]+.[0-9]+'
+      - '[0-9]+.[0-9]+.[0-9]+'
 
 jobs:
   publish:
@@ -19,9 +19,13 @@ jobs:
         with:
           python-version: "3.10"
       - name: install flit
-        run: pip install flit
+        # both distributions are pinned together: flit declares an unbounded
+        # flit_core lower bound, so pinning only flit lets the two drift apart
+        run: pip install flit==3.12.0 flit_core==3.12.0
       - name: Build and publish
-        run: flit publish
+        # --use-vcs is the flit 3 default, but flit 4 flipped it: state it
+        # explicitly so tests/ and docs/ stay in the sdist across upgrades
+        run: flit publish --use-vcs
         env:
           FLIT_USERNAME: __token__
           FLIT_PASSWORD: ${{ secrets.PYPI }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -19,7 +19,7 @@ jobs:
         with:
           python-version: "3.10"
       - name: install flit
-        run: pip install flit~=3.4
+        run: pip install flit
       - name: Build and publish
         run: flit publish
         env:


ubmarco:
Pushed 5ecdc232, which changes the shape of this fix. The original one-liner (`pip install flit`) does unblock `flit publish`, but reviewing it turned up two problems that only surface after the upload, when it is too late to take back.

**1. Unpinning doesn't remove the drift, it resets it.** The 8.3.1 failure came from flit 3.12.0 declaring an unbounded `flit_core>=3.12.0`, so pip pulled flit_core 4.0.2 into a flit 3 CLI. flit 4.0.2 declares an unbounded `flit_core>=4.0.2` — the same shape. It holds only while the CLI keeps pace with the backend; the day `flit_core 5.0` lands before a matching flit, `pip install flit` resolves flit 4.x + flit_core 5.x and we are here again, discovered with the tag already pushed.

**2. Unpinning silently performs a flit 3 → 4 upgrade of the build path, and that changes what we ship.** flit 4 flipped the sdist default to `--no-use-vcs`. Built from clean clones of this commit:

| toolchain | sdist entries | contents |
|---|---|---|
| flit 3.12 (`Found 1325 files tracked in git`) | 1326 | `sphinx_needs/`, `tests/`, `docs/`, `docker/`, `performance/`, … |
| flit 4.0.2, unpinned as originally proposed | 703 | `sphinx_needs/`, README, LICENSE, AUTHORS, pyproject |
| flit 4.0.2 `--use-vcs` | 1326 | same as flit 3 |

The published `sphinx_needs-8.3.0.tar.gz` has exactly those 1326 entries — 485 `tests/` files, 109 `docs/`, 27.7 MB. So the original fix would have shipped an sdist with the test suite and docs removed, breaking redistributors that build from the tarball and run the suite as packaging QA (conda-forge, Fedora/Debian, Nix). The same upgrade also moves the wheel to `Metadata-Version: 2.5`, adds `Import-Name`, and reclassifies `AUTHORS` as a license file (flit 4 widened the default license globs), which leaks into SBOM and license-compliance output.

None of that is caught anywhere: CI has no packaging job, so the first observation of a build-backend change is an immutable PyPI upload.

## What this PR does now

https://github.com/useblocks/sphinx-needs/blob/5ecdc232fddb181f4721f138805aa8b5255cae2e/.github/workflows/release.yaml#L21-L28

- **Pin both distributions.** They are one unit; pinning only `flit` is the option that is worse than either pinning both or pinning neither, because it lets the halves move independently. `3.12.0` is the pair that produced 8.3.0, so the release is unblocked without smuggling a backend major bump into a hotfix.
- **`flit publish --use-vcs`.** A no-op on flit 3, but it states the sdist contents explicitly instead of inheriting a default that already changed once, so the eventual flit 4 upgrade can't silently drop `tests/` and `docs/` again.
- Plus a `+` on the tag filter's major segment, so `10.0.0` still matches.

Verified against the published 8.3.0 artifacts, building this branch with the pinned pair: sdist 1326 entries (the only deltas are the real repo changes since 8.3.0 — `tests/test_plantuml_incdir.py` added, `docs/_static/tutorial_needs.json` removed), wheel `Metadata-Version: 2.4`, `License-File: LICENSE` only. Shape parity with the last good release.

## Getting 8.3.1 out

Merging alone does not publish it. `refs/tags/8.3.1` points at 8f5ea6bd — the pre-fix commit — and re-running the failed job replays `release.yaml` as of the tag's commit, i.e. still `flit~=3.4`, so it fails identically. Nothing was ever published under 8.3.1 (PyPI is still at 8.3.0), so the tag and its release can simply be recreated: after merge, delete the 8.3.1 release and tag, re-tag the merged commit, and push it. The workflow then runs with the fixed file, and the tag ends up pointing at the commit that actually produced the artifacts.

## Deliberately not in this PR

- **A packaging job in CI** (`flit build --use-vcs && twine check dist/*`). This is the real fix for the class of problem — it is why a backend break reached a release at all. Separate PR, no reason to couple it to unsticking 8.3.1.
- **The flit 4 upgrade**, once CI covers the build: bump both pins, widen `flit_core >=3.4,<4` to `<5` in `pyproject.toml`, and decide deliberately about `Metadata-Version: 2.5` and `AUTHORS` being declared as license text. Note the cap is currently ignored at release time — `flit publish` builds in-process rather than via PEP 517 isolation — so with the 3.12 pin the declared and actual backends agree again; under flit 4 they would not.
- **Trusted Publishing (OIDC)**, to retire the long-lived `secrets.PYPI` token.

chrisjsewell (approved):

PR #1760, opened 2026-08-12
🔧 Accept ubCode-only directive options and refresh the docs ubCode config

chrisjsewell:
Two related changes, one commit each.

## 1. Accept the ubCode-only directive options as no-ops

[ubCode](https://ubcode.useblocks.com) supports a few view-directive options that
Sphinx-Needs has no equivalent for. Today a document written for ubCode does not just lose
those options under Sphinx — it fails to build:

```text
index.rst:11: ERROR: Error in "needlist" directive:
unknown option: "cypher".
index.rst:16: ERROR: Error in "needtable" directive:
unknown option: "cypher".
index.rst:21: ERROR: Error in "needflow" directive:
unknown option: "cypher".
index.rst:28: ERROR: Error in "needsequence" directive:
unknown option: "max_items".
build finished with problems, 4 warnings (with warnings treated as errors).
```

That is a docutils level-3 error, so the directive class is never instantiated: the view is
replaced by a system message plus a literal echo of the source, and `-W` builds fail.

This adds the options to the `option_spec` of the directives ubCode accepts them on, as
`directives.unchanged`:

| directive | options |
|---|---|
| `needlist` | `cypher`, `max_items` |
| `needtable` | `cypher`, `max_items` |
| `needflow` | `cypher`, `max_items`, `width`, `height` |
| `needsequence` | `max_items`, `width`, `height` |

(`width`/`height` are accepted by ubCode on the two diagram directives as of its next
release.) They go in each directive's own `option_spec` literal rather than in
`FilterBase.base_option_spec` or `DiagramBase.base_option_spec` — neither mixin's spread
matches this set, so using one would invent surface on `needextract`/`needgantt` that no
ubCode document can use.

These options may gain native implementations in future Sphinx-Needs releases, in which
case they would take effect here as well rather than being ignored — the shared docs note
says so, so the surface is forward-compatible rather than permanently inert.

### Why this is not a behaviour change

1. **The options never reach a node.** Each of the four directives builds its node
   attributes from a closed, hand-written literal, and this PR does not touch
   `collect_filter_attributes()` or `collect_diagram_attributes()`. The options are parsed
   and dropped — exactly how `export_id` already behaves. A doctree dump of a document using
   all of them shows zero compat attributes on all four nodes.
2. **They never reach `needs.json`.** There is no route from a view node into
   `needsfile.py`'s output; the old `has_export_filters` route was removed. Verified on a
   `needs_build_json = True` build: zero occurrences.
3. **The `ubproject.toml` additions below are invisible to Sphinx.**
   `load_config_from_toml` reads only `toml_data["needs"]`, and every table this PR touches
   (`[build.html]`, `[parse]`, `[intersphinx]`) is top-level (`[parse]` already exists on master; corrected from "adds" per review F4).

Empirically: the full test suite's failure set is byte-identical before and after, the
`test_basic_doc` doctree snapshot does not churn, and `sphinx-build -W` on a document using
all the options goes from exit 1 to exit 0.

### The trade-off, stated plainly

Accepting an option silently costs feedback. For `cypher` and `max_items` that is
negligible — they are ubCode coinages and are not near-misses for any real Sphinx-Needs
option. For **`width` and `height` it is a real, if small, regression**: someone who has
never heard of ubCode and writes `:width: 400` on a `needflow` gets a loud error today, and
after this PR gets a silently ignored option and a diagram that is not the size they asked
for. The mitigation is documentation, not a warning — a warning would defeat the purpose,
since every ubCode-authored directive would then warn and `-W` builds would still fail. The
new note in `docs/filter.rst` says so explicitly, and I am happy to drop those two options
if you would rather not take that trade.

A `needs_ubcode_compat` config flag was considered and rejected: ubCode-authored docs would
still fail out of the box, which is the whole problem, and it is strictly more surface than
11 lines of `directives.unchanged`. Ungated inert options already exist in the tree
(`export_id`, and `show_link_names` / `highlight` / `config_names` on the diagram
directives).

### Docs and tests

- One shared, anchored note in `docs/filter.rst` (next to the `export_id` precedent, but a
  plain `.. note::` — these are not deprecated), plus a one-line reference from each of the
  four directive pages.
- `tests/doc_test/doc_ubcode_compat/` + `tests/test_ubcode_compat.py`, asserting both that
  the build is warning-free and the four views render, and that none of the option names
  appears in any view node's attributes.

## 2. Refresh `docs/ubproject.toml`

`docs/ubproject.toml` is this documentation's ubCode configuration, and via
`needs_from_toml` also the source of its `needs_*` settings. It had drifted from `conf.py`,
so ubCode reported warnings the Sphinx build does not. `ubc check` on `docs/` goes from
**29 warnings to 15** (0 errors throughout):

| added | effect |
|---|---|
| `rst_epilog` (the `\|br\|` substitution, as in `conf.py`) | −10 `substitution.undefined` |
| `[intersphinx.projects.python]` + `.sphinx` (as in `intersphinx_mapping`) | −4 `intersphinx.ref` |
| `sphinx.ext.extlinks` + `sphinx.ext.todo` in `parse.extensions` | warning-neutral |
| `[parse.extlinks.pr]` + `[parse.extlinks.issue]` (as in `extlinks`) | warning-neutral, but rendering-critical: ~312 `:pr:` and ~181 `:issue:` references otherwise render without a link |

Removed: the `extend_directives` entries for `grid-item-card`, `button-link`, `button-ref`,
`uml` and `page-summary`, which ubCode now supports natively. Each was checked on its own and
in combination and is warning-neutral; `page-summary` is additionally unused anywhere in the
docs. `need-warnings` and `need-core-fields` are kept, since `conf.py` registers those
locally.

Also `[build.html]` and `docs/_static/ubc-theme.css`, so the experimental `ubc build html`
site carries the same branding as the Sphinx one (`ubc build html`: 40 pages, both
stylesheets linked in cascade order). The theme sheet is self-contained — no `@import`, no
external URLs.

`[needs]` and `[needs_json]` are untouched, so the Sphinx build sees no change at all.


ubmarco (approved):
LGTM

PR #1765, opened 2026-08-14
✨ Add needs_card_layouts, a declarative way to define need layouts

chrisjsewell:
### What this adds

A new configuration option, `needs_card_layouts`, that describes a need's rendering as a
small dictionary — a *card specification* — instead of as hand-written layout strings:

```python
needs_card_layouts = {
    "product": {
        "extends": "clean",
        "meta": {"include": ["status", "tags"]},
        "footer": ["id", "type"],
        "collapse": "closed",
    }
}
```

```rst
.. req:: A requirement
   :id: REQ_1
   :layout: product

   The body of the need.
```

Specifications are compiled during `config-inited` (priority 550 — after
`merge_default_configs`, before `check_configuration`) into ordinary `needs_layouts`
entries. Because `LayoutHandler.__init__` is the single place the layout registry is ever
read, a compiled card is honoured by the `:layout:` option, `needs_default_layout`,
`needextract`, `needservice` and the GitHub service **with no change to the renderer, the
grids, the CSS or the JavaScript**. The whole change outside the new module is nine lines.

`needs_layouts` is untouched and stays fully supported: it is the escape hatch for the
layouts the card vocabulary deliberately cannot express. This is not a deprecation.

### Why compile instead of render

The card vocabulary is validated and typed; the layout DSL is not, is re-parsed per need per
build, and fails by raising `SphinxNeedLayoutException`, which derives from `BaseException`
and so escapes every `except Exception` in Sphinx's node-processing path — an aborted build
with a bare traceback and no document location, for a string the user never wrote.

The compiler is therefore built so that it *cannot* produce such a string:

* it emits only from a closed set of module-level templates;
* the only user text that ever reaches a layout string is a field name, and only after
  matching `^[A-Za-z_][A-Za-z0-9_-]*$`;
* it never emits `is_external` (which would perform a build-time HTTP fetch and write into
  the source tree), never emits a literal `{{` (a need placeholder that raises when it
  cannot be resolved), never emits more than one `collapse_button` (the collapse JavaScript
  resolves icons with `document.querySelector`), and only emits a `collapse_button` for a
  card that has both a header and a meta region, so its `meta` target row always exists —
  the JavaScript has no null guard there;
* it checks that every section it emits is one the chosen grid actually reads, since unread
  sections are dropped silently.

Three sweeps over the whole valid-specification matrix assert each of these.

### Validation and warnings

A specification that cannot be compiled is reported through a new `needs.card_layout`
warning subtype and skipped. The build continues and the other cards still compile. A card
is likewise skipped, never used, when its name is already taken by a built-in layout or by a
`needs_layouts` entry. The collision set is exactly those two: layouts that *services*
register do so later in the build than card layouts are compiled, so a card named `github`
takes precedence over the GitHub service's layout silently — exactly as a
`needs_layouts` entry of that name already does today. That is documented as a named limit
on the new page rather than special-cased in the compiler, so that the service's existing
user-override affordance keeps working through the new door as well.

The subtype is added to both sorted dicts in `logging.py`, so the user-facing subtype list in
the documentation regenerates itself.

### Semantics mapping

The specification vocabulary is shared with ubCode, which implements it natively; the two
therefore need a stated mapping onto this repository's grids and layout functions.

Regions to grid:

| header | meta | side | footer | grid |
|---|---|---|---|---|
| ✔ | any | — | — | `simple` |
| ✔ | any | — | ✔ | `simple_footer` |
| ✔ | any | full | — | `simple_side_{left,right}` |
| ✔ | any | partial | — | `simple_side_{left,right}_partial` |
| ✘ | ✘ | — | — | `content` |
| ✘ | ✘ | — | ✔ | `content_footer` |
| ✘ | ✘ | ✔ | — | `content_side_{left,right}` |
| ✘ | ✘ | ✔ | ✔ | `content_footer_side_{left,right}` |

Elements (`footer` and `side` share one set):

| element | emitted |
|---|---|
| `id` | `<<meta_id()>>` |
| `title` | `<<meta("title")>>` |
| `type` | `<<meta("type_name")>>` |
| `layout_echo` | `layout: <<meta("layout")>>` |
| `style_echo` | `style: <<meta("style")>>` |
| `field:<n>` | `<<meta("<n>")>>` |
| `image:<f>` | `<<image("field:<f>", align="center")>>` |

Meta region:

| spec | emitted |
|---|---|
| `fields = "stored"` / `"effective"` | `<<meta_all(no_links=True)>>` |
| `fields = "all"` | `<<meta_all(exclude=[], defaults=False)>>` |
| `exclude` | `exclude=["…"]` on the call |
| `empties` | `show_empty=True` on the call |
| `links = true` | additional line `<<meta_links_all()>>` |
| `meta = False` | no meta region (a `content*` grid) |
| `side = False` or `side = {"elements": []}` | no side region — the two spellings are equivalent, here and in ubCode |
| `include = [x,y]` | one `<<meta("x")>>` per field, in order |
| `collapse = "honour"` / `"closed"` | `collapse_button(…, initial=False / True)` |
| `collapse = "open"` | no `collapse_button` |

Because the compiler targets the existing grids, the built-in permutations round-trip: the
specifications for `clean_l`, `clean_lp`, `clean_r`, `clean_rp`, `focus_f`, `focus_l` and
`focus_r` compile back to those layouts **byte for byte**, which the tests assert.

Documented limits, all on the new page:

* no grid has a header, a side region *and* a footer, and none has a meta region without a
  header — such a specification is reported and skipped;
* a headerless side region has no partial grid, so `span = "partial"` degrades to `"full"`
  with a warning;
* `links_back` cannot be switched off on its own, because the link types are not yet known
  at `config-inited`; such a card is reported and rendered with back links;
* `collapse = "open"` shows the meta region with no toggle, since a layout cannot override a
  need's own `:collapse:`, only leave the button out;
* `"stored"` and `"effective"` render the same fields here — `meta_all` iterates the need
  either way. The distinction is kept because it is meaningful in tools consuming the same
  specifications;
* a header with `meta = false` leaves an empty meta band, since the header grids always emit
  that row;
* `id` in a side region renders horizontally — the vertical strip of `focus_l` / `focus_r`
  comes from CSS selected on those layout *names*;
* layouts registered by services are not in the collision set, since services register later
  than card layouts are compiled (see above).

### Future vocabulary

The vocabulary is deliberately a closed, typed set per version: it grows by adding typed
elements over the fixed regions, never by admitting template strings or user-defined
grids — those remain exactly what `needs_layouts` is for.

The first step is the element **object form**: every element string becomes canonical
sugar for a dict — `"image:diagram"` ≡ `{"type": "image", "field": "diagram"}` — and the
dict is where per-element options land (`height` / `width` on an image, a display label
on a field). Purely additive for existing configs, and the options compile 1:1 onto
`image()` / `meta()` keyword arguments the layout DSL already has.

The further additions, all of which already have compile targets in the existing DSL:
per-link-type link rows (`meta_links("<type>", incoming=…)`), a `link:<field>` element
for URL-valued fields (`link(is_dynamic=True)`), and a header composed as an ordered
element list with display labels. Unknown keys and elements are reported and skipped
today, so a spec written for a newer vocabulary degrades loudly rather than silently on
an older release.

### Tests

`tests/test_card_layouts.py`, 142 tests, in three layers:

* a unit matrix over the compiler: every element, every meta tier and switch, every collapse
  mode, all fourteen grid selections, `extends` from each of the twelve built-in bases and
  through user chains, sub-table merging, built-in-table immutability, and 25 validation
  failures (unknown keys, bad enums, bad identifiers, cycles, impossible grids, title
  placement, collisions);
* end-to-end `test_app` builds asserting the emitted table class list and the
  `rowspan`/`colspan` skeleton, with `app.warning_list == []` on the happy path;
* a warning table asserting that each invalid card appears in the build warnings, that the
  build still succeeds, and that its sibling card still compiled and rendered.

Three `conformance_*` specifications are committed verbatim and built end to end, so this
implementation of the vocabulary can be compared against another's.

### Docs

* new page `docs/card_layouts.rst` (How-tos, after `layout_styles`), linking into the
  existing grid anchors rather than editing `layout_styles.rst`;
* the feature is used by this documentation itself: two `needs_card_layouts` entries in
  `docs/ubproject.toml` (read via `needs_from_toml`) drive live rendered examples on the
  new page — a side+image profile card and a footer/collapse summary card. The three
  pre-existing `needs_layouts` entries there are deliberately untouched: each uses
  something the card vocabulary cannot express (a prose head line, `permalink()`,
  `sidebar()`), which is precisely the escape-hatch split the two options are meant
  to have;
* `needs_card_layouts` entry in `docs/configuration.rst`, next to `needs_layouts`;
* `docs/changelog.rst` updated.


ubmarco (approved):
LGTM, declarative is better than programmed

PR #1770, opened 2026-08-15
✨ needflow: a portable option vocabulary, with deprecations

chrisjsewell:
Third slice of the needflow portable-options work, riding #1768 (bug fixes) and #1769 (one
graph model, two thin emitters). It gives `needflow` an engine-neutral option and config
vocabulary — each option expresses intent that plantuml, graphviz, and ubCode's mermaid
renderer can honour or gracefully degrade — plus a discouraged per-engine escape hatch.

Two existing options were **widened rather than replaced**, so the spellings most likely to be
in existing documents keep working and gain no deprecation.

## Options

- **`:direction:`** (new) — `down` (default) / `up` / `right` / `left`, with `TB`/`TD`/`BT`/`LR`/`RL`
  aliases and a `needs_flow_direction` default. An explicit option beats an engine-config-derived
  direction on both engines, and disagreement warns. PlantUML has no bottom-to-top or
  right-to-left primitive (probed: both are syntax errors), so those degrade to their axis mate
  with a single warning per project.
- **`:show_link_names:` widened** — the flag now takes an optional value `none` / `outgoing` /
  `incoming` / `type`, and written bare still means exactly what it always meant (`outgoing`),
  byte-for-byte. `needs_flow_show_links` likewise accepts a string as well as a boolean
  (`True` ≡ `outgoing`, `False` ≡ `none`). That fixes the OR-interaction which made "project
  default on, this one diagram off" impossible to express.
- **`:show_legend:` widened** — it now takes the *name* of a legend defined in
  `needs_flow_legends`, or nothing. Deliberately key-only: an inline value set would collide with
  user-chosen names and need a precedence rule. Written bare it renders the legend this engine
  draws best, which upstream is today's in-diagram legend, byte-for-byte.
- **`:styles:`** (new) — a comma list of `[<filter>]:<class>` rules (the existing variant syntax)
  over named classes from `needs_flow_styles`, with CSS-cascade semantics (declaration order,
  last match wins per property) and a closed property set: `fill`, `border`, `border_width`,
  `border_style`, `text_color`, `shape`. The built-in `highlight` class byte-preserves today's
  `:highlight:` output.
- **`:engine_config:`** (new) — the escape hatch: named per-engine blobs from
  `needs_flow_engine_config`, deliberately config-level so documents stay portable.
  `:config:` becomes a deprecated alias.

## Config

`needs_flow_legends = {name: {...}}` defines legends — `parts`, an **ordered list** of the
sections to draw (`types`, `links`), and `placement` (`internal` / `external`). Unset `placement`
takes the engine's own default placement, which is internal here and external in a renderer with
no legend construct. `needs_flow_show_legend` names the legend a diagram gets when it asks for one
without naming its own; it selects *which*, never *whether*, so presence stays per-directive and
the key namespace needs no value meaning "off".

An undefined key resolves as a **chain** — the directive's key, else the project's, else the
engine default — warning at each undefined step and handing on, because an unusable value is
treated as unset everywhere in this vocabulary. The two steps warn at different tiers: a bad
directive key once per directive, a bad `needs_flow_show_legend` once per project.

`needs_links[].line/part_line/color/part_color/arrow` (neutral enums; `color` is finally
honoured — an identical TODO sat in both emitters — and is unset-by-default, so a link type can
now ask for black) and `needs_types[].shape` (a ten-member neutral enum; the legacy plantuml
`style` keywords are accepted as aliases). Options consult their config default only when unset
(the `max_items` rule from #1761), and every enumerated config value is validated at load time
with a warning and a fallback rather than a crash. Those values are matched the way the directive
options already match theirs — case-insensitively, ignoring surrounding whitespace — so
`needs_flow_direction = "Down"` no longer warns and falls back where `:direction: Down` is
accepted. (docutils' own `choice` normalises, so `:engine: PlantUML` has always been accepted;
the config half was the odd one out.)

`part_color` is an additive amendment to the planned vocabulary: without it, the deprecated
`style`/`style_part` pair could express something the neutral keys could not — a distinct colour
for part links — which stranded four link types in this project's own docs configuration on the
deprecated spelling.

## Six deprecations, honoured indefinitely

`:config:`, `:scale:` (deprecated without replacement — it was already a silent no-op on
graphviz), `:highlight:` (kept as sugar for `[<filter>]:highlight`), `:border_color:`,
`needs_links[].style/style_part/style_start/style_end`, and `needs_flow_link_types` — the last
deprecated as *dead*: it has never had any effect, and the docs that taught otherwise are
corrected. Every deprecation warns only when used; option-free builds stay warning-free. The
suite's two whole-project empty-warnings fences are intact and unmodified; three narrower fence
assertions gained a single-substring allow-list because their fixture *is* the deprecated feature
under test — each still asserts everything else is empty.

## Nothing moves for existing projects

Review-verified by cross-commit diff of the generated diagram source on both engines: projects
using only existing spellings — including bare `:show_link_names:` and bare `:show_legend:` —
produce byte-identical output, built HTML, and image filenames. No `.rst` test fixture in the repo
needed changing: `git diff origin/master -- 'tests/doc_test/**/*.rst'` is empty. The four
`conf.py` diffs come from the `needs_links` neutral-key migration and, in `doc_links/conf.py`
only, dropping a `needs_flow_link_types` that the dead-config deprecation made pointless.

`mermaid` is additionally accepted as a reserved engine name (drawing with the default engine,
with a warning) so a ubCode-authored document does not lose its diagram when rendered here.

## Bugs fixed on the way

- graphviz label wrapping could split an HTML entity (a quote in a need title produced invalid
  markup), because escaping ran before wrapping
- multi-name `:config:` merges mutated a shared dict, leaking one diagram's graphviz styles into
  the next
- `:class:` was silently dropped by the plantuml emitter
- a non-dict `needs_graphviz_styles` entry crashed the build
- an invalid `needs_flow_engine` failed via a bare `assert` (stripped under `python -O`)
- `:debug:` on plantuml now renders a line-numbered literal block, matching graphviz

## Conformance corpus

`tests/conformance/needflow/` holds 28 language-neutral cases —
`(needs data, options/config) → expected emitted source per engine + expected degradations` —
checksummed by a manifest and duplicated into ubCode, which owns the copy of record and adds the
mermaid expectations. Because the corpus names the portable vocabulary rather than either tool's
spelling, widening `:show_link_names:` changed only the mapping table. Legend expectations are
per-engine (`expect.<engine>.legend`), since an internal legend is already asserted byte-exactly
by that engine's `source` and only an external one is separate document content. Expected sources
were each reviewed against the case's stated purpose rather than snapshotted blind — that review
is what surfaced the entity-wrapping bug above. The harness refuses malformed cases and checksum
mismatches, and the refusals are themselves tested; checksums are taken over line-ending-normalised
bytes so a Windows checkout and a Linux one agree.

## Environment version

`ENV_DATA_VERSION` goes 6 → 9: the directive persists the resolved options in doctrees, and an
older reader over a newer doctree fails on the missing keys (reproduced by rebuilding across the
change and watching it raise, then fixed by the bump). 7 and 8 were intermediate shapes that
existed only within this branch, so the released step is 6 → 9. A sweep of open PRs on the literal
constant name found no competing claim.

## Tests and docs

Every behaviour change and every deprecation warning was recorded failing before its fix or
feature landed. The full suite passes (1534 passed, 13 skipped) and the docs build is
warning-free with the blanket `needs.deprecated` suppression *removed* — the deprecated demos are
now non-executed samples pointing at their replacements' rendered examples. The docs gain a
portable-options section, per-option `.. versionadded::`/`.. deprecated::` markers, and an
arrow-migration table (today's default graphviz arrowhead is `open`, not `normal`).


chrisjsewell:
## Open design questions before this merges

Three things worth settling while both this PR and the paired ubCode one are still open, since
all of them get much more expensive once the vocabulary is released.

### 1. Fold the new options back into the existing names, rather than adding new ones

The proposal: instead of `:show_legend:` → `:legend:` and `:show_link_names:` → `:link_labels:`
as deprecating renames, **keep the existing option names and widen what they accept**. Same for
the config key.

**Recommendation: do it, for all three.** It removes three of the nine deprecations, and — more
importantly — it removes them from the three spellings most likely to appear in existing
documents.

| Today | Proposed | Notes |
|---|---|---|
| `:show_link_names:` (flag) | `:show_link_names:` with optional value `none` / `outgoing` / `incoming` / `type` | Bare flag ≡ `outgoing`, which is exactly what it means today |
| `needs_flow_show_links = True` | `needs_flow_show_links` accepting `bool` **or** the same four strings | `True` ≡ `outgoing`, `False` ≡ `none` |
| `:show_legend:` (flag) | `:show_legend:` with optional value `types` / `links` / `types,links` | Bare flag keeps the legacy in-diagram rendering — see the wrinkle below |

For the link labels this is a clean flag→enum widening: there is no semantic fork, the bare form
already means precisely one of the new values, and the OR-interaction defect still gets fixed
because `none` becomes expressible per diagram (the thing a boolean project default can't be
overridden by today). `directives.unchanged` plus a validator replaces `directives.flag`, and
`needs_flow_show_links` becomes a `bool | str` config — a shape Sphinx configs use routinely.

**The one wrinkle, on `:show_legend:` only.** The bare flag renders the legend *inside* the
diagram (each engine's own drawing); the new values render an out-of-diagram table shared by all
engines. So one option name would carry two different renderings, distinguished by whether a
value is given.

That is defensible, and I'd resolve it as: **bare `:show_legend:` keeps the legacy in-diagram
rendering, frozen and undeprecated, indefinitely** — byte-preservation for existing documents —
while any value selects the portable table, which is what the docs teach going forward. The
alternative (bare form also switches to the table) silently changes the output of every existing
`:show_legend:` document, which is exactly what this PR has otherwise avoided everywhere.

**Knock-on if adopted:** `needs_flow_legend` / `needs_flow_link_labels` should become
`needs_flow_show_legend` / `needs_flow_show_links` for symmetry — both are new in this PR, so
there is no compatibility cost, only consistency. `:direction:` is unaffected: there is no legacy
name for it to preserve.

### 2. `:width:` / `:height:` — yes, and there is a live problem here

Right now `:width:` and `:height:` are accepted by the option spec
(`_directive.py:84-85`, `directives.unchanged`) but the values are **never read** — not stored,
not consulted by either emitter. They are silent no-ops upstream, while ubCode implements them.

That matters for this PR specifically, because the `:scale:` deprecation added here tells users:

> `_directive.py:256` — *"Please use `:width:` / `:height:` instead."*
> `docs/directives/needflow.rst:512` — *"Use `:width:` / `:height:` instead."*

We are currently deprecating a working-ish option by redirecting users to two options that do
nothing at all. That has to be resolved before release, one way or the other:

- **preferred** — implement `:width:` / `:height:` functionally (mirroring the ubCode semantics),
  as its own small PR following the `max_items` pattern from #1761, landing in the same release
  as this deprecation; or
- reword the `:scale:` deprecation so it does not promise a replacement that isn't there yet.

The first is better, and it also closes a real cross-tool divergence.

### 3. Theme-adaptive colours (furo light/dark toggle)

Worth stating plainly: **this is a pre-existing limitation, not something this PR introduces**,
and the vocabulary does not make it worse.

Configured colours (`needs_types[].color`, `:border_color:`, the graphviz styles, and the new
`needs_links[].color`) are baked into a rendered image at build time. A runtime theme toggle
cannot reach inside a generated PNG/SVG, so a colour chosen for a light background stays that
colour in dark mode. Two parts of the current design already behave well:

- the new out-of-diagram legend is plain docutils/HTML, so it themes correctly with the rest of
  the page;
- colours are now **unset by default** — an unset colour inherits the engine's own default rather
  than being pinned to a literal, so a project that configures nothing has nothing to clash.

The seam for an actual fix is `:styles:` being **class-name based** rather than inline values.
A class name can later be lowered to a CSS class instead of a baked attribute wherever the render
target is HTML-native. Realistically that means: very tractable for a client-side renderer
(ubCode's mermaid — classes map to CSS custom properties that follow the theme), possible for
graphviz only with inline SVG output rather than an ``, and effectively impossible for
PlantUML image output.

**Recommendation:** don't block the vocabulary on it. Add a short docs note now — configured
colours are baked at build time, so prefer values legible on both backgrounds, or leave them unset
— and treat theme-adaptive class styling as its own follow-up, ubCode-first, where it is actually
achievable end-to-end.

---

Happy to make (1) and the docs half of (3) part of this PR, and take (2) as the separate slice it
was always planned to be.

---

chrisjsewell:
## Splitting this into reviewable slices

Marked as draft: at 8,695 insertions across 57 files this is too much to review as one
unit, and it mixes changes that are purely additive with decisions about how much to break
existing projects. Those deserve separate diffs.

The plan is five PRs. The first three should be near-mechanical to review; every
"should we break this?" decision is isolated in the last two, so they can be taken one at
a time.

| # | Slice | Content | Nature |
|---|---|---|---|
| 1 | emitter fixes | the six needflow bugs — graphviz label wrapping splitting HTML entities, `:config:` merges mutating a shared dict, `:class:` dropped on plantuml, a non-dict `needs_graphviz_styles` crashing the build, `needs_flow_engine` validated by a bare `assert`, and `:debug:` harmonised — each with a regression test | pure fixes, no new surface |
| 2 | conformance corpus | the language-neutral fixture harness, with cases covering **existing** behaviour only (baseline, node-id injectivity, `%%` neutralisation, colour normalisation, empty edge labels) | test infrastructure |
| 3 | presentation options | `:direction:`, plus widening `:show_link_names:` and `:show_legend:` in place, with `needs_flow_direction` / `needs_flow_show_links` / `needs_flow_legends` / `needs_flow_show_legend` | **additive — nothing deprecated, no existing spelling changes meaning** |
| 4 | link & type styling | `needs_links[].line/part_line/color/part_color/arrow` and `needs_types[].shape`, with the legacy `style*` keys aliased and deprecated | first migration decision |
| 5 | styles & escape hatch | `:styles:` + `needs_flow_styles`, `:engine_config:` + its registries, and the remaining deprecations (`:config:`, `:scale:`, `:highlight:`, `:border_color:`, `needs_flow_link_types`) | the rest of the breakage question |

Ordering note: the entity-wrapping fix has to come first, because later slices record
generated diagram source as test expectations — without the fix those expectations would
bake in the invalid markup.

One benefit of splitting: `ENV_DATA_VERSION` on master is still 6, so slice 3 takes a clean
7 rather than the 6 → 9 jump this branch had accumulated across intermediate shapes.

Nothing here is discarded. Each slice is carved by subtraction from the reviewed state on
this branch, so the work that has already been through review is preserved rather than
rewritten. This PR stays open as the reference until its content has been redistributed.

---

Issue #1773, opened 2026-08-17
A non-boolean filter is accepted or rejected depending on which path compiles it


chrisjsewell:
A filter string is supposed to evaluate to a boolean; `filter_common.py:785-788` rejects
anything else:

```python
        result = eval(filter_compiled or filter_string, filter_context)
        if not isinstance(result, bool):
            raise NeedsInvalidFilter(
                f"Filter did not evaluate to a boolean, instead {type(result)}: {result}"
            )
```

But not every filter reaches that check. Simple expressions are answered by the query
fast path in `sphinx_needs/ubquery.py`, and a bare field name there is coerced with
`bool(...)` (`ubquery.py:276-279`):

```python
    # --- bare name (e.g. is_external) ---
    if isinstance(expr, ast.Name):
        if expr.id in _CONTEXT_ONLY_NAMES:
            return None
        return lambda need, ctx=None, _f=expr.id: bool(_get_field(need, _f, ctx))
```

So whether a non-boolean filter is an error depends on whether the expression happens to
be simple enough for the fast path.

## Measured

At `3fc55cd`, on a project with four needs, two of which have `links`:

| filter | result | warning |
|---|---|---|
| ``:need_count:`links``                          | **2** | none |
| ``:need_count:`id.startswith("REQ_") and links`` | **0** | `WARNING: Filter ... did not evaluate to a boolean, instead <class 'list'>: []. [needs.filter]` |

The two filters differ only by a conjunct that is true for both needs, and they disagree
about both the count and whether anything is wrong. Same contrast in `needtable`,
`needlist`, `needpie` and `needbar` — this is the engine, not a directive.

## Why it matters

This is the mechanism behind #1167, where `x and derives` produced
`Filter did not evaluate to a boolean`. The answer given there is correct — `derives` is
a list, so `and` yields the list — but it is only half the story: `derives` **alone**
would have been accepted, silently, with `bool()` semantics. From a user's point of view
the rule "a filter must be a boolean" appears to hold sometimes and not others.

## Directions, none of which I want to pick unilaterally

1. **Coerce everywhere** — apply `bool(...)` in the eval path too. `x and derives`
   starts working and agrees with `derives`. But filters that today get a warning would
   silently change meaning, and a truthiness rule for arbitrary expressions is a real
   semantic commitment.
2. **Reject everywhere** — drop the coercion in the fast path. Consistent and strict,
   but it breaks bare-name filters such as `is_external` that work today, which is
[...]

Issue #1772, opened 2026-08-17
A filter string with invalid syntax ends the build, in every filtered directive


chrisjsewell:
A filter string that does not *parse* is not reported as a `needs.filter` warning. It
raises out of the resolve phase, `sphinx-build` exits 2, and nothing is written for the
whole project.

A filter that parses but fails at runtime is handled properly, which makes the
difference easy to miss:

```rst
.. needtable::
   :filter: xxx

.. needtable::
   :filter: status ==
```

The first gives `WARNING: Filter 'xxx' not valid. Error: name 'xxx' is not defined.
[needs.filter]` and an empty table. The second ends the build with
`SyntaxError: invalid syntax (<string>, line 1)`.

## Where

`sphinx_needs/filter_common.py:616`:

```python
    # === Slow path: fall back to eval() per item ===
    found_needs_slow: list[NeedItem | NeedPartItem] = []

    # https://docs.python.org/3/library/functions.html?highlight=compile#compile
    filter_compiled = compile(filter_string, "<string>", "eval")   # <-- line 616
    error_reported = False
    for filter_need in needs:
        try:
            ...
```

`compile()` sits just outside the `try` that wraps the per-need evaluation, so a
`SyntaxError` propagates instead of becoming the warning that the same loop emits for
every other filter problem.

The intent looks like it was a warning all along. The fast path swallows exactly this
error at `filter_common.py:500-503`:

```python
    try:
        body = ast.parse(filter_string).body
    except Exception:
        pass  # warning already emitted in filter_needs
```

The comment is not true today — nothing emits a warning, the build just stops.

## Measured

At `3fc55cd`, `sphinx-build -b html -E`, one project per case. `status ==` as the filter
string:

| directive | rc | result |
|---|---|---|
| `needtable` `:filter:` | 2 | `SyntaxError: invalid syntax` |
| `needlist` `:filter:` | 2 | same |
| `:need_count:` | 2 | same |
| `needpie` content line | 2 | same |
| `needbar` cell | 2 | same |

So this is engine-wide, not specific to one directive. It is also easy to hit by
accident in `needbar`, where the default cell separator is `,`: a filter such as
`status in ["open", "closed"]` is split into pieces that do not parse, and the build
ends with `SyntaxError: '[' was never closed`.

## Proposal

Move the `compile()` inside the `try`, and report a `SyntaxError` the same way as any
other invalid filter: `log_warning(..., "filter", location=...)`, the filter yields no
[...]

Issue #1771, opened 2026-08-17
needpie / needbar: ordinary user errors end the whole build instead of warning


chrisjsewell:
Most malformed input to `needpie` and `needbar` does not produce a Sphinx warning.
It raises out of the `doctree-resolved` handler, so `sphinx-build` exits 2 and **no**
output is produced for the whole project — not just for the offending chart.

That is out of step with the rest of the extension: the filter engine already warns and
degrades. A bad `:filter-func:` reference, for instance, is handled properly at
`sphinx_needs/directives/needpie.py:177-185` — `log_warning(..., "filter_func",
location=node)` and the node is dropped, build continues.

### Measured

Everything below was measured at `3fc55cd` with `sphinx-build -b html -E`, one small
project per case, matplotlib 3.11.1, Sphinx 9.0.4. `rc` is the exit code.

| input | rc | what the user sees |
|---|---|---|
| `needpie` `:labels:` count ≠ content lines (either direction) | 2 | `ValueError: 'labels' must be of length 'x', not 2` |
| `needpie` `:explode:` count ≠ content lines | 2 | `ValueError: 'explode' must be of length 'x', not 1` |
| `needpie` `:explode: abc` | 2 | `ValueError: could not convert string to float: 'abc'` (during reading) |
| `needpie` `:colors: notacolor` | 2 | `ValueError: Invalid RGBA argument: 'notacolor'` |
| `needpie` `:text_color: nope` | 2 | matplotlib color error |
| `needpie` `:style: no_such_style` | 2 | `OSError: ... not a valid package style` |
| `needpie` `:filter-func:` returning non-numbers | 2 | an `ERROR` log line, then `could not convert string to float` |
| `needpie` content line that is an RST comment (`.. x`) | 2 | `SyntaxError: invalid syntax` |
| `needbar` rows of differing length (an empty line counts as a row) | 2 | `Exception: Needbar - file 'chart' - line '4': each content line must have the same length` |
| `needbar` `:xlabels:` count ≠ columns | 2 | `Exception: ... length of xlabels: 1 is not equal with sum of columns: 2` |
| `needbar` `:ylabels:` count ≠ rows | 2 | `Exception: ... length of ylabels: 1 ... sum of rows: 2` |
| `needbar` `:colors: notacolor` | 2 | `ValueError: 'facecolor' or 'color' argument must be a valid color` |
| `needbar` `:style: no_such_style` | 2 | `OSError` |
[...]

PR #1774, opened 2026-08-17
🐛 needpie / needbar: this_doc filters, reproducible images, FROM_DATA labels, alt text and docs

### LINKED ISSUE #1449: The filter c.this_doc() is not working with needbar and needpie
The filter [c.this_doc()](https://sphinx-needs.readthedocs.io/en/latest/filter.html#filtering-for-needs-on-the-current-page) is not working with needbar and needpie, and maybe not with needsequence.

How shall we change the code?

In [needbar](https://github.com/useblocks/sphinx-needs/blob/500a4486e32c37f6c6d85a67d257f6d5a97d9456/sphinx_needs/directives/needbar.py#L305) and [needpie](https://github.com/useblocks/sphinx-needs/blob/500a4486e32c37f6c6d85a67d257f6d5a97d9456/sphinx_needs/directives/needpie.py#L168) currently we use `filter_needs_parts`.

Shall we change the implementation to introduce the missing `origin_docname` parameter in the usage of `filter_needs_parts` OR shall we introduce the `process_filters` as in [needtable](https://github.com/useblocks/sphinx-needs/blob/500a4486e32c37f6c6d85a67d257f6d5a97d9456/sphinx_needs/directives/needtable.py#L214C26-L214C41)?

In [needsequence](https://github.com/useblocks/sphinx-needs/blob/500a4486e32c37f6c6d85a67d257f6d5a97d9456/sphinx_needs/directives/needsequence.py#L281) we have to extend the parameters for sure, as a `process_filters` does not fit to the current algorithm.

Please provide guidance how to proceed.

chrisjsewell:
Five small fixes to the chart directives, found while reviewing them end to end. One
commit per fix, plus a final review-round commit that tightens tests, comments and docs
wording across them. The larger behaviour questions the same review surfaced are filed
separately (#1771, #1772, #1773) rather than changed here.

Closes #1449.

## 1. `c.this_doc()` never worked in `needpie`, `needbar` or `:need_count:`

`c.this_doc()` needs to know which document a filter was written in. When it is not
told, it raises, so the filter is reported as invalid and matches nothing:

```
WARNING: Filter 'c.this_doc()' not valid.
Error: `this_doc` can not be used in this context. [needs.filter]
```

`needtable`, `needlist` and the other directives whose `:filter:` runs through
`process_filters` are fine, because that helper passes `origin_docname=filter_data["docname"]`
(`sphinx_needs/filter_common.py:182`). `needpie`, `needbar` and the `:need_count:` role
call `filter_needs_parts` themselves, and did not pass it, so `NeedCheckContext.this_doc`
hit its `_origin_docname is None` branch (`filter_common.py:805-808`) every time.

**(A) or (B)?** The issue offers two shapes and asks for guidance. This is **(A)**, and I
think (A) is clearly the right one:

- the value is already on both chart nodes — `"docname": env.docname` is stored at
  `needpie.py:88` and `needbar.py:131` — so (A) is one keyword argument per call site;
- (B), routing the charts through `process_filters`, is a rewrite rather than a fix.
  `process_filters` takes a single `NeedsFilteredBaseType` carrying one `filter` string,
  whereas a chart evaluates N independent expressions per node, one per content line or
  grid cell. Going that way means synthesising a filter-data dict per line, and it drags
  in `sort_by`, `debug_filters` and the external-needs switch, none of which the charts
  have or want.

**The `:need_count:` role is included.** The issue names only the two chart directives,
but the role has the same defect at three call sites (`sphinx_needs/roles/need_count.py:42`,
`:53`, `:58` — the single-filter form and both halves of the `" ? "` ratio form). Leaving
it out would mean `c.this_doc()` keeps failing silently in the role, so it is fixed here
too. The role's equivalent of the charts' stored `docname` is the `refdoc` attribute that
`XRefRole` puts on the node, and it is read defensively with `.get()`, so an absent value
keeps today's behaviour.

**`needsequence` — yes, and two others, but not in this PR.** The issue wonders whether
`needsequence` is affected as well. **It is.** Its `:filter:` does not go through
`process_filters`; it calls `filter_single_need` directly, with no origin document
(`sphinx_needs/directives/needsequence.py:375-382`). Two sibling options share the gap,
for the same reason:

| site | option |
|---|---|
| `sphinx_needs/directives/needsequence.py:375-382` | `needsequence` `:filter:` |
| `sphinx_needs/directives/needflow/_model.py:417` | `needflow` `:highlight:` |
| `sphinx_needs/directives/needgantt.py:250`, `:318` | `needgantt` `:milestone_filter:` |

In all three the symptom is worse than the one this PR fixes. `filter_single_need`
*raises* `NeedsInvalidFilter`, and none of these call sites is inside a `try`, so
`c.this_doc()` there **ends the build** instead of warning. `docs/filter.rst:239`
meanwhile tells users `c.this_doc()` "is for most directives", so this is a real
documentation mismatch, not just a latent edge. (The `needs_constraints` check path
evaluates filters the same way and shares the gap, for completeness.)

These three are deliberately not fixed here, to keep this PR to the two chart directives
and the role the issue names. They are a small follow-up: `filter_single_need` already
takes `origin_docname` as a keyword-only argument, and each of these nodes stores its own
docname (`needsequence.py:73`, `needgantt.py:125`, and `needflow`'s `attributes`, which
is what `process_filters` already reads the docname from). **Happy to extend this PR to
cover all three if you would rather have them in one go** — say the word and I will add
them with the same fixture treatment.

*Tests:* new project `tests/doc_test/doc_filter_this_doc/`, two documents,
`c.this_doc()` used in a `needpie`, a `needbar` and a `:need_count:` (both its forms) on
each. `tests/test_filter.py::test_this_doc_in_charts_and_need_count` asserts an empty
warning stream **and** the counts (`2`, `66.7`, `150.0`, `1` — each document counts its
own), so it does not pass merely because a warning went away. Reverting the source hunks
turns it red with seven `needs.filter` warnings — one per affected call site in the
fixture.

## 2. Chart images are not reproducible

Two builds of an unchanged chart never produced the same bytes. Matplotlib's SVG writer
embeds the wall clock time as `<dc:date>`, and — when `svg.hashsalt` is unset — derives
clip-path and marker ids from a random `uuid4`
(`matplotlib.backends.backend_svg.RendererSVG._make_id`). Measured on two `-E` builds of
one project into two outdirs: identical file *names* (the name is already a digest of
the target id), different contents, in every chart.

`save_matplotlib_figure` now writes the figure with `metadata={"Date": None}` and inside
`rc_context({"svg.hashsalt": <the chart's own file name>})`. That name is the sha256 of
the directive's target id, so it is stable across builds and still distinct per chart.
The PDF writer that `LaTeXBuilder` selects names the same value `CreationDate`, so it
gets `metadata={"CreationDate": None}`; its object ids come from a sequential counter
and are already deterministic, so it needs no salt. PNG needs nothing — it already
writes no timestamp. Measured by calling the writer twice per format: `svg`, `pdf` and
`png` are now all byte-identical, where `pdf` was not before. The rendered chart is
unchanged in every format.

*Tests:* `tests/test_needpie.py::test_chart_images_are_reproducible` builds the existing
`doc_needpie` and `doc_needbar` projects twice into two build dirs and compares the
`_images/*.svg` bytes; reverting the fix fails it for both.
`tests/test_needbar.py::test_chart_element_ids_are_unique_per_chart` covers the other
half of the salt's job — that the four bars of one document still get *different*
element ids — which a single constant salt would silently break.

## 3. `:ylabels: FROM_DATA` on its own always ends the build

```rst
.. needbar:: only ylabels
   :ylabels: FROM_DATA

   Tags a, 'a' in tags, 'a' in tags
   Tags b, 'b' in tags, 'b' in tags
```

ends every build with `length of xlabels: 3 is not equal with sum of columns: 2`.

The default xlabels are derived at `needbar.py:255-256` from `local_data[0]`, but the
ylabels column is only taken out of `local_data` afterwards, at `:258-265`. So the
default was always one label too long, and the arity check at `:271-274` could never
pass. Only the both-`FROM_DATA` and x-only forms worked.

Both label defaults are now derived after the label row and column have been consumed.
Nothing else moves: when xlabels come from the option or from the content, the moved
branch does not run at all — I checked the three previously-working forms byte for byte,
and their SVGs are identical before and after.

One degenerate form does change, which I mention for completeness: a grid left with zero
data columns after the label column is stripped (`:ylabels: FROM_DATA` over single-cell
rows) used to end the build with `length of xlabels: 1 is not equal with sum of
columns: 0`, and now builds an empty chart instead. That seems the better of the two
behaviours, and matches what `needbar` already does for all-zero data, but say the word
if you would rather it stayed an error.

*Tests:* `tests/test_needbar.py::test_needbar_label_defaults` builds four bars in one
document — ylabels-only, xlabels-only, both, neither — and reads the actual x tick label
text back out of each SVG: `["1", "2"]` for the two derived cases, `["A", "B"]` for the
two taken from the content. Reverting the fix reproduces the abort.

## 4. `needpie`: no `alt` text, and an image file for a chart that is never shown

Both are visible by contrast with `needbar`.

`needbar` publishes its title as the image's `alt` (`needbar.py:487-488`); `needpie` set
no `alt` at all, so every pie went out with docutils' fallback — the image's own file
URI — which tells a screen reader nothing. A pie without a title keeps that fallback,
exactly as an untitled `needbar` does.

The figure was also saved at `needpie.py:316-319`, *before* the check at `:324-327` for
a pie with no data. Such a pie is replaced by the "No needs passed the filters"
paragraph, so the file stayed in `_images/` with nothing referring to it. It is now
written only on the branch that shows it.

*Tests:* `tests/test_needpie_with_zero_needs.py` grew assertions (it previously only
called `app.build()`): the replacement text is shown, no ``'&lt;img' is emitted, and``
`_images/` is empty. Tests that pinned the old file-URI `alt` now pin the title, keeping
their file name coverage through `src`. The untitled case is pinned too — untitled pies
must keep the file URI rather than gain an empty `alt`.

## 5. Documentation corrections

`docs/directives/needpie.rst` and `needbar.rst` described several things the directives
do not do. Measured and corrected:

- both claimed "this generates multiple image files ... and allows the document engine
  to pick the appropriate image type". Exactly one file is written, in the first type
  the builder accepts that matplotlib can produce (`utils.py:446-455`);
- both called a literal content value a "static float/int value". The guard is
  `str.isdigit()` (`needpie.py:164`, `needbar.py:301`), so `10.5` and `-5` fall through
  to the filter path and warn. The pages now attach that warning to those two examples
  rather than to non-boolean filters in general, because a simple enough expression is
  answered by the query fast path, which coerces with `bool()` and does not warn — the
  inconsistency raised as #1773;
- what an invalid value does was unstated, and it is rarely a warning: a label,
  `:explode:` or grid-shape mismatch, an unknown color and an unknown style each end the
  build. That is now said plainly. (Whether it *should* end the build is #1771, raised
  rather than changed here.)
- the `needbar` colors hint promised the opposite of what happens: the color list is
  concatenated with the default palette (`needbar.py:357-359`), not repeated, so the rows
  past the end of the list restart at the palette's *first* color rather than getting the
  color they would have had without the option. The "warning" the hint mentions is a
  `print()` at `:362` whose condition is effectively unreachable;
- the rotation options only accept non-negative integers (`.isdigit()` at
  `needbar.py:428`, `:449`, `:458`); anything else is dropped silently;
- the `:separator:` paragraph was garbled and understated the trap: with the default
  separator, any filter containing a comma is split into invalid pieces and the build
  ends (the abort itself is part of #1772);
- `needpie`'s `:filter_warning:` option was undocumented, as was the rule that content
  and `:filter-func:` are alternatives;
- `needbar` accepts no filter options at all, which was nowhere stated.

## Checklist

- [x] Tests for each behaviour change
- [x] `docs/changelog.rst` updated (one entry per fix)
- [x] Documentation updated


ubmarco (approved):
Nice, good change

PR #1775, opened 2026-08-17
🐛 Pass the origin document to needsequence, needflow :highlight: and needgantt filters

chrisjsewell:
This is the follow-up offered at the end of #1774: *"Happy to extend this PR to cover all three if you would rather have them in one go."* Same one-argument fix, same fixture treatment, for the three filter options that #1774 named but deliberately left out.

## The severity here is worse than the one #1774 fixed

`c.this_doc()` needs to know which document a filter was written in. The chart directives fixed in #1774 route through `filter_needs_parts`, which catches the failure and logs a `needs.filter` warning. These three route through `filter_single_need`, which **raises** `NeedsInvalidFilter` — and none of the four call sites is inside a `try`. So `c.this_doc()` here does not warn, it **ends the build**:

```
sphinx_needs.exceptions.NeedsInvalidFilter: Filter 'c.this_doc()' not valid.
Error: `this_doc` can not be used in this context.
```

Meanwhile `docs/filter.rst:239` tells users `c.this_doc()` "is for most directives".

## The sites

| site | option | docname source |
|---|---|---|
| `needsequence.py:378` | `needsequence` `:filter:` | node stores it — `needsequence.py:73` |
| `needflow/_model.py:417` | `needflow` `:highlight:` | needflow `attributes`, the same dict `process_filters` reads `docname` from |
| `needgantt.py:250` | `needgantt` `:milestone_filter:` (elements) | node stores it — `needgantt.py:125` |
| `needgantt.py:318` | `needgantt` `:milestone_filter:` (constraints) | same |

Four call sites, not three: `needgantt` evaluates its milestone filter twice, once per section. Each node already carries its own docname, so this is one keyword argument per site, plus one parameter threaded through `get_message_needs` (recursive) and `resolve_presentation`.

## `needs_constraints` was assessed and is *not* included

#1774's body mentioned the constraints check path "for completeness". It does share the gap — `c.this_doc()` in a `needs_constraints` check aborts the build today too, I measured it — but it cannot be fixed the same way, and I do not think it should be:

- `needs_constraints` is a **conf.py** dictionary. The filter expression is not written in any document.
- `process_constraints(needs, config)` runs once, project-wide, from `post_process_needs_data` — there is no directive node, no `env.docname`, no location.
- The only docname in scope is `need["docname"]`, the document of the need *being checked*. Since `this_doc` is `need.is_in_document(origin_docname)`, passing that would make `c.this_doc()` return `True` unconditionally — a silently-always-true filter, which is worse than the current honest failure.

So the abort there is the correct outcome, and the fix is documentation. `docs/filter.rst` now says which options support `c.this_doc()` and states that conf.py-level filters (`needs_constraints`, `needs_warnings`) do not, and why. The same reasoning applies to the needs-builder filters.

## Two more sites were assessed and deliberately left out

- **`needuml` content filters** (`needuml.py:502`): the `filter()` Jinja function never receives an origin document either — today `c.this_doc()` there warns and matches nothing (the milder failure class #1774 fixed, not the abort class). The obvious one-kwarg fix is a no-op, because `transform_uml_to_plantuml_node` hardcodes `fromdocname=None`; threading the real docname through works for the simple case but gives a **wrong** answer for arch content imported across documents via `{{ uml(...) }}` — `jinja2uml` shares one `JinjaFunctions` instance across its recursion, so imported content authored on another page would resolve `this_doc` to the *importer's* document. Answering correctly needs per-need origin tracking through that recursion, a design change rather than a kwarg — so it stays out, and the docs now say so.
- **`needimport` `:filter:`**: runs against the raw imported JSON before those dicts become needs (`filter_import_item` builds its own eval context — there is no `c` at all), where "this document" has no meaning. Also documented as unsupported.

## Tests

New project `tests/doc_test/doc_filter_this_doc_diagrams/`, two documents, four needs, with a `needsequence :filter:`, a `needflow :highlight:` and a `needgantt :milestone_filter:` on **each** document. `tests/test_filter.py::test_this_doc_in_diagram_filters` asserts on the generated PlantUML source captured at `doctree-resolved` — the pattern already used by `tests/test_views_max_items.py` — so nothing depends on the PlantUML binary being able to draw it.

Every diagram's filter is evaluated against needs from both documents, so the two documents must come out as exact complements:

| | `index.rst` | `page.rst` |
|---|---|---|
| sequence arrows | `SENDER -> INDEX_RECV` | `SENDER -> PAGE_RECV` |
| highlighted (`line:FF0000`) | `SENDER`, `MESSAGE`, `INDEX_RECV` | `PAGE_RECV` |
| milestones (`lasts 0 days`) | `SENDER`, `MESSAGE`, `INDEX_RECV` | `PAGE_RECV` |
| gantt constraints | `[SENDER] happens at …` | `[SENDER] starts at …` |

That last row is there to cover `needgantt`'s *second* call site specifically — the two are textually identical, so a patch fixing only the first would otherwise look correct.

`tests/test_needflow_model.py` gains three unit tests for `resolve_presentation`'s new parameter, including one pinning that omitting it raises rather than silently returning `False`.

Reverting each source file on its own — and the second `needgantt` call site on its own — turns the test red with the abort above in every case.

## One incidental tidy-up

`needsequence.py` had a function-local `from sphinx_needs.filter_common import filter_single_need` running once per receiver id, in the exact block that needed the new argument; the module already imports from `filter_common` at top level, so there was no circular-import reason for it. Hoisting it and collapsing the resulting double `if` is why that file's diff is larger than the other two. Happy to drop it if you would rather keep the diff minimal.

## Checklist

- [x] Tests for the change
- [x] `docs/changelog.rst` updated — new bugfix entry, and #1774's now-stale *"three filter options are still not covered"* paragraph removed, since it is in the same unreleased section
- [x] Documentation — `docs/filter.rst` states which options support `c.this_doc()` and which conf.py-level filters cannot
- [x] `ruff check`, `ruff format` and `mypy` (via pre-commit) pass


ubmarco (approved):
Approved via API review.

PR #1789, opened 2026-08-26
🐛 A need records the line it is actually written on

### LINKED ISSUE #1349: Lineno gets shifted if rst_prolog is present

chrisjsewell:
### The change

`NeedDirective` stored `self.lineno`, which counts the lines of what the *parser* was
handed rather than the lines of the file. Three perfectly ordinary things put text
into a document and shift that count for everything after: Sphinx prepends
`rst_prolog` to every document it parses, and both docutils' `.. include::` and our
own `.. list2need::` splice text in mid-parse with `StateMachine.insert_input`. The
shifts compound, so the recorded line routinely landed past the end of the file.

The line now comes from `SphinxDirective.get_source_info()`, which is exactly
`state_machine.get_source_and_line(self.lineno)` — the accessor `get_location()`, and
so every warning the directive itself emits, is already built on. **The stored
`lineno` now agrees with the line our own warnings have been printing all along.** On
a document with an `rst_prolog` and a `list2need`, for a need written on
`index.rst:9`:

| | before | after |
|---|---|---|
| what the warning printed | `index.rst:9` | `index.rst:9` |
| what `needs.json` recorded | `19` | `9` |

The accessor returns `(None, None)` when the state machine cannot map the line, in
which case the old value is kept.

**No warning location regresses.** The only warning locations that move are the ones
this corrects — the sites that report at a need's stored `(docname, lineno)` now name
the true line. Everything anchored through the directive's own location helper is
byte-identical to before. The one place that had to be taught the
difference is the rendered content of `:pre_template:`/`:post_template:`. That content
exists in no file, so it is anchored at the need's own directive, and `nested_parse`
resolves a nested parse through the *enclosing* state machine — meaning the anchor has
to be given in the parser's line space, not in the resolved one a need now records.
`NeedItemSourceDirective` therefore also keeps the unresolved number (`parser_lineno`,
internal — not in `needs.json`), and `_create_need_node` uses it for those two offsets.
`tests/test_proper_warning.py` pins these locations already; its values are unchanged,
and a second test now builds **the same fixture** with an `rst_prolog` in front of it
and asserts **the same five locations**, which without this would report 7 and 4 where
the file says 9 and 6.

### What that looks like across the three causes

A 19-line document with all three, from the new test:

| need | written on | before | after |
|---|---|---|---|
| after `rst_prolog` only | `index.rst:4` | 7 | **4** |
| after an `.. include::` | `index.rst:9` | 24 | **9** |
| after a `.. list2need::` | `index.rst:18` | 47 | **18** |

### Deliberately not changed

- **`docname`** stays `self.env.docname`. #1692 also derived it from the resolved
  source path; measured, that turns `docname` into a filesystem path for any need
  inside an `.. include::`d file and silently breaks the `:need:` hyperlinks pointing
  at it. Changing `docname` changes link resolution, which is a separate question
  from getting the line right.
- **`lineno_content`** stays `self.content_offset + 1`. It is not read as a source
  line: `_create_need_node` hands it to `nested_parse` as an offset into the parser's
  own line space, which is what lets a need nested inside another need's content report
  sensible lines. Making the two agree wants a new field rather than a change to this
  one; noted as a follow-up.
- **Needs that `list2need` generates** still record their position inside the
  generated block rather than the line of the list item that produced them — that is
  where their warnings already point, so it is no worse than before, and it is only
  properly fixed by building those needs directly instead of round-tripping them
  through the parser.
- **One honest trade.** For a need written inside an `.. include::`d file, inside a
  `list2need` block, or in a Python docstring, the number now indexes a *different*
  file from the one `docname` names, and it looks like an ordinary line of that
  document — where the old value announced itself as wrong by sitting past the end of
  the file. Both are wrong; the new one is wrong *plausibly*. It is deliberate: the new
  number is the one that need's own warnings already print, and the real fix is for a
  need to carry a source path of its own rather than for `docname` to change meaning.
- **Markdown is a no-op.** myst-parser's `MockStateMachine.get_source_and_line`
  returns the line it is given, so `.md` needs are byte-identical before and after.
  There is a test pinning that.

### Tests

- `tests/test_need_lineno.py` (new) — one project carrying all three shift sources,
  pinning the true line for every ordinary need directive, plus the Markdown no-op.
- `tests/test_list2need_behaviour.py` — the lineno characterization test added in
  #1788 now records the *correct* line for the need after the list, and keeps the
  generated needs' offsets pinned as the remaining residue.
- `tests/__snapshots__/test_list2need.ambr` — 16 generated-need linenos, each checked
  against the generated text.
- `tests/test_proper_warning.py` — its existing pins are untouched; a new case rebuilds
  the same fixture behind an `rst_prolog` and asserts the same warning locations.

Closes #1349.

Credit where it is due: #1692 (sebastiansetzer) identified this fix and its
direction. This PR carries its `need.py` line-number piece on its own.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -274,7 +274,7 @@ These changes do not affect user-facing behaviour:
 Bug fixes
 .........
 
-- 🐛 A need records the line it is actually written on **(changed output)** (:issue:`1349`)
+- 🐛 A need records the line it is actually written on **(changed output)** (:issue:`1349`, :pr:`1789`)
 
   ``rst_prolog``, ``.. include::`` and :ref:`list2need` each put text into the document
   being parsed, and each one used to shift the ``lineno`` recorded for every need after
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -274,6 +274,20 @@ These changes do not affect user-facing behaviour:
 Bug fixes
 .........
 
+- 🐛 A need records the line it is actually written on **(changed output)** (:issue:`1349`)
+
+  ``rst_prolog``, ``.. include::`` and :ref:`list2need` each push text into the running
+  parse, and each one used to shift the ``lineno`` recorded for every need after it by
+  the length of the text it inserted. The shifts compound, so in a document with a
+  prolog and a couple of includes the recorded line was routinely past the end of the
+  file. The line is now resolved back to the one the directive stands on — the same
+  line Sphinx-Needs' own warnings have always reported for that need.
+
+  Two things are deliberately unchanged: the needs :ref:`list2need` *generates* still
+  record their position inside the generated block rather than the line of the list
+  item that produced them, and ``lineno_content`` is still the parser's own counter,
+  because that is what it is used as.
+
 - 🐛 Fix :ref:`needs_variant_data_file` triggering full rebuilds (:issue:`1783`, :pr:`1787`)
 
   The merged variant data was written back onto the configuration after Sphinx's
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -283,9 +283,11 @@ Bug fixes
   line is now resolved back to the one the directive stands on — the same line the
   directive's own warnings have always reported.
 
-  **Warning locations are unchanged**, everywhere, including inside the content a
-  :ref:`need_pre_template` or :ref:`need_post_template` renders, which is anchored at
-  the need's directive and is now anchored there explicitly.
+  **No warning location regresses.** The only warning locations that move are the
+  ones this corrects — sites that report at a need's stored line now name the true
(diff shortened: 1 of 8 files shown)


ubmarco (approved):
Nice catch

PR #1790, opened 2026-08-26
✨ list2need: build the needs directly, without generating reStructuredText

chrisjsewell:
`list2need` renders every item into a need directive through a template and hands the
result back to the parser with `state_machine.insert_input`. This replaces that step:
the items are built with `add_need`, and a nested item is placed inside its parent.

**The syntax and its parsing are untouched.** The same four regular expressions, the
same level→type mapping, the same `(ID)` capture, the same `:delimiter:` split, the same
`((option="value"))` extraction, the same `:tags:` merge, the same `:links-down:` walk,
the same validation and the same eleven inputs that end the build. Only the generation
mechanism changes.

This is offered as the answer to the question in #1427 — *"can we save the syntax
(mostly) but have a better implementation (without jinja)?"*

### What the step cost

**Markdown.** myst-parser's mock state machine does not implement `insert_input`, so a
`{list2need}` fence produced no needs at all — only an error naming a myst-parser
internal. It now works:

| | before | after |
|---|---|---|
| needs created | none | `MD-A`, `MD-B` |
| nesting | — | `MD-B.parent_need == "MD-A"` |
| diagnostics | `MockingError: MockStateMachine has not yet implemented attribute 'insert_input'` | none |

**Line numbers (#1349, the `list2need` half).** `insert_input` advances the parser's flat
line counter by the length of the generated text, for the rest of the file, and the error
compounds with each list. For a document with two items on lines 10 and 11 and an
ordinary `.. req::` on line 13:

| need | written on | `lineno` before | after |
|---|---|---|---|
| `LN-BEFORE` | 4 | 4 | 4 |
| `LN-A` | 10 | 14 | **10** |
| `LN-B` | 11 | 20 | **11** |
| `LN-AFTER` (an ordinary `.. req::`) | 13 | 27 | **13** |

In the repository's own `doc_list2need` fixture, `NEED-Z` was recorded at line 45 of a
19-line file. The general case — a `lineno` shifted by `rst_prolog` or an
`.. include::` — is a separate small PR, #1789.

**Template artefacts.** An item written `()` with options lost them: suppressing the
`:id:` left a blank line where it would have been, which ended the generated need's
option block, so every option after it became body text. And a parent need's stored
`content` was the generated reStructuredText of its children rather than its own text,
which `needs.json` and any filter reading `content` saw.

### The one deliberate change to an input that worked

Writing `()` used to suppress the `:id:` of the generated need, which sent it down the
need directives' own id generator — a **different** function, which honours
`needs_id_from_title` where `list2need`'s does not, hashes the *content* when the title
is empty, and hashes the title *after* the `((option="value"))` area has been removed
from it. One list could carry two id schemes at once, chosen by two characters of
punctuation. The directive now derives every id itself, from the same input at the same
point, so `()` means what no brackets at all means.

Three kinds of item therefore get a different generated id:

| item | before | after |
|---|---|---|
| `* ()Alpha title` with `needs_id_from_title = True` | `R_ALPHA` | `R_D1EC6`, the plain hash |
| `* (). some content here` (no title) | `R_EBC29`, the hash of the **content** | `R_DA39A`, the hash of the empty title |
| `* ()Title with empty parens ((status="open"))` | `R_6AFF7` | **`R_D7997`** — the id `* Title with empty parens ((status="open"))` already had |

The third needs no configuration to reach and is the one to watch for; the changelog
says so under Breaking changes, and the documentation carries a `versionchanged`.
Everything else keeps the id it had, which the untouched id tests prove.

### Four inputs that were errors, and are now defined

None of these can appear in a document that builds cleanly today, so nothing that
works can regress; they are listed because each is a decision the template used to
delegate to docutils and the directive now has to make.

- **`((id="MY-ID"))`** names the need. The template wrote the derived id *and* the
  inline one as two `:id:` lines of one generated need, which docutils refused with
  `duplicate option "id"`, dropping the item. The id is applied before `:links-down:`
  is built, so the other items' links agree with it. An empty `((id=""))` is refused
  with a diagnostic rather than quietly replaced by the title hash.
- **An inline option naming the same link field as `:links-down:`** — `((links="X"))`
  under `:links-down: links` — produced one corrupt link value (the template's
  `{%- for %}` glued the two field lines into `:links: C-2:links: X`) and an
  `unknown outgoing link` warning. The two sets are now merged.
- **`((title_from_content="true"))`** is read as a need directive reads it.
- **A child of a `((hide="true"))` item** is rendered at the level above it. A hidden
  need is taken out of the document once read, so nesting a child inside one would put
  its target where no page can reach it.

### Evidence

- The rendered HTML of `tests/doc_test/doc_list2need/` is **byte-identical** before and
  after, all four pages, apart from the per-build random `SNCB-…` container ids. Its
  `needs.json` keeps all 16 ids, every title, status, tag, link and `parent_need`; only
  the linenos and the two parent `content` strings change.
- The characterization suite added in #1788 is the safety net. The eleven build-aborting
  inputs and the `links-down` level-skip test pass **untouched** — this PR takes neither
  of those on. **Four** assertions are edited, each one pinning something the template
  did: the drifted linenos, the MyST failure (which inverts), the parent's
  generated-text content, and the second id function. A fifth prediction went the other
  way (below).
- Two tests gain an assertion: the Markdown ones now pin the items' line numbers.
  Linenos inside an `{eval-rst}` fence were wrong too — that block was the only way to
  reach the directive from a Markdown document, and its needs were recorded at 10 and
  16 for items written on 7 and 8.
- `tests/__snapshots__/test_list2need.ambr` is rewritten: every `lineno` in it was wrong.
- Six tests are added: the `()`-plus-options id, a directive written in an item's
  content, and one for each of the four inputs above.

One assertion moved the other way. Removing the three-space indent the parser prepends to
a `:`-continuation line looked like removing a workaround for the template — it is not:
it is the only thing standing in for the indentation `lstrip` removes, and without it the
options of a directive written in an item's content (the `rst-directives in lists`
example in the documentation) land in the directive's own column and are dropped. It is
kept, and now has a test.

Refs #1427, Refs #1349.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -298,7 +298,7 @@ Breaking changes
   ID list2need derives is the hash of the title *before* the option area is removed
   from it, which is the ID an item without the ``()`` has always been given, while the
   second generator hashed the title after it. So ``* ()A title ((status="open"))``
-  moves from ``R_6AFF7`` to ``R_D7997`` — the ID ``* A title ((status="open"))``
+  moves from ``R_328F3`` to ``R_0FF10`` — the ID ``* A title ((status="open"))``
   already had.
 
 - ‼️ :ref:`needflow` ``:show_link_names:`` takes an optional value, and now wins over
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,6 +14,36 @@ Unreleased
 Improvements
 ............
 
+- ✨ :ref:`list2need` works in Markdown documents, and records the line each need was
+  written on (:issue:`1349`, :pr:`1789`)
+
+  The directive used to render every item into a need directive through a template and
+  hand the result back to the parser. That step is gone: the items are built directly,
+  and a nested item is placed inside its parent. The syntax and its parsing are
+  untouched — the same list structure, the same ``(ID)`` capture, the same
+  ``:delimiter:`` split, the same ``((option="value"))`` region, the same validation —
+  so a list that built before builds the same needs, with the same IDs.
+
+  What the step cost, and is therefore fixed:
+
+  - A list written as a ``{list2need}`` fence in a MyST Markdown document produced no
+    needs at all, only an error naming a myst-parser internal. It now creates its needs,
+    like any other directive.
+  - Handing generated text back to the parser advanced its line counter for the rest of
+    the file, so every need from the directive onwards — the items themselves, and any
+    ordinary need directive written below them — was recorded at a line further down the
+    file than it was written, by an amount that grew with each list in the document.
+    Every ``lineno`` in ``needs.json`` is now the line the need was written on.
+  - An item written ``()`` with options, such as ``* ()A title ((status="open"))``, lost
+    them: they became body text. They are now set.
+  - A content line starting with ``:`` was indented by three further spaces to keep it
+    out of the generated need's option block, which could produce a stray
+    ``Definition list ends without a blank line`` warning. It is plain content now.
+
+  The content of a parent need no longer holds the generated reStructuredText of its
+  children — nesting is a property of the document, not of the parent's text — so
+  ``needs.json``, and a filter reading ``content``, see the item's own text only.
+
 - 👌 :ref:`needs_variant_data` is resolved while the configuration is being initialised
   (:issue:`1783`, :pr:`1787`)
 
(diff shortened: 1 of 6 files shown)


ubmarco (approved):
LGTM

PR #1801, opened 2026-08-31
📚 fix four typos in configuration.rst

MsfPablo:
Four spelling fixes in `docs/configuration.rst`:

- `availble` -> `available`
- ```rigth``` -> ```right``` (in the list of `needflow` orientations)
- `caculated` -> `calculated`
- `"mininum"` -> `"minimum"` in the `needs_extra_options` schema example

The last one is slightly more than cosmetic: the JSON Schema keyword is `minimum`, so as written the example constraint would be ignored rather than applied.


chrisjsewell (approved):
cheers!

PR #1823, opened 2026-09-02
🔧 Run the sphinx matrix from the lock, and retire tox for poe

chrisjsewell:
Three sphinx versions now live in one `uv.lock`, every CI job installs from it, and the
local task runner is [poethepoet](https://poethepoet.natn.io) instead of tox.

**No source code changes.** 14 files: `pyproject.toml`, `uv.lock`, six workflows,
`.github/dependabot.yml`, one new CI script, `.gitignore`, `.pre-commit-config.yaml`,
`AGENTS.md`, `docs/contributing.rst` and `docs/changelog.rst`. Nothing under `sphinx_needs/`
or `tests/`.

### Why tox goes

No CI job has ever run tox here. Every workflow installed with `pip install -e .[…]` and
called `pytest` or `sphinx-build` directly, so tox's environments were purely a local
convenience — which makes this a local-workflow swap, not a CI risk. Separately, tox-uv's
runner overwrites a PyPI-installed member with the local one, so a matrix that means to test
a *released* version can go green while testing the working tree; that trap is worth walking
away from before this repository grows into a workspace.

The replacement is `[tool.poe.tasks]`. `uv run poe` lists every task with its help, and words
written after a task name are appended to its command, which is what `{posargs}` used to
give. A task that needs a dependency set other than the default declares it through poe's uv
executor and redirects `UV_PROJECT_ENVIRONMENT` to `.venvs/<name>`, so a matrix cell and a
docs build do not fight over `.venv` — never with a nested `uv run` inside the command, which
syncs twice and leaves the wrong packages behind.

```
uv run poe test                                   # the suite, newest sphinx
uv run poe test-sphinx7                           # ... one matrix cell
UV_PYTHON=3.12 uv run --no-sync poe test-sphinx8  # ... on a chosen interpreter (uv downloads it)
uv run poe mypy                                   # the prek mypy hook
uv run poe docs                                   # furo, -nW --keep-going
```

pyenv is no longer needed for any of this. One caveat worth knowing: give the *outer*
`uv run` its `--no-sync` in the `UV_PYTHON=` form, or uv rebuilds the default `.venv` on that
interpreter on its way to the task (`uv sync --python 3.13` puts it back). AGENTS.md and the
contributing page both say so.

### What `[tool.uv] conflicts` buys

The matrix is now three dependency groups — `sphinx-7`, `sphinx-8`, `sphinx-9` — declared
conflicting with each other:

```toml
conflicts = [[{ group = "sphinx-7" }, { group = "sphinx-8" }, { group = "sphinx-9" }]]
```

One lock therefore holds sphinx 7.4.7, 8.2.3 and 9.1.0 (alongside the 8.1.3 that the lock
already carried for python 3.10), each with its own resolution. Every CI cell becomes a
command a contributor can run verbatim:

```
uv sync --frozen --no-default-groups --group test --group sphinx-9
```

Without the table the project does not lock at all: one universal resolution cannot satisfy
`~=7.4`, `~=8.2` and `~=9.1` at once.

### The furo floor is load-bearing

`furo~=2024.8.6` caps sphinx at `<9.0`, and it was the single declaration in the repository
that could not span 7 → 9. Keeping it would have meant pulling all four `theme-*` extras into
the conflict set: measured, that is ~630 KB of extra lock and 28 top-level
`resolution-markers` entries instead of 16. `theme-furo` now requires `furo>=2025.12.19`,
the first release that supports sphinx 9 — a floor rather than `~=`, because furo's versions
are dates and `~=2025.12.19` would admit one month of releases and then go stale exactly the
way `~=2024.8.6` did. The docs build green on furo 2025.12.19 + sphinx 9.1.0 under
`-nW --keep-going`. The other three theme extras are untouched; none of them caps sphinx.

### Type checking

`#1819` removed the `mypy` dependency group and made the prek hook's
`additional_dependencies` the one place the pinned floor toolchain lives. This branch keeps
it that way: `uv run poe mypy` runs `prek run mypy --all-files`, so a contributor and CI
type-check against exactly the same versions, and there is no second copy of the pins to
drift. A comment on the hook says so.

The conflict set is therefore the three sphinx groups only, and a plain `uv sync` resolves to
sphinx 9.1.0 — the newest — rather than being capped by anything.

### `test`, `test-parallel` and `benchmark` stop being extras

They only ever installed development tools, so they are dependency groups now, and
`pip install sphinx-needs[test]` no longer resolves. `test-parallel` folds into `test` as
`pytest-xdist`. `plotting`, `docs` and the four `theme-*` extras stay extras: Read the Docs
installs `docs` + `theme-furo` through pip, and `plotting` is an end-user extra. A bare
`uv sync` is now a working development environment, which it was not before. The changelog
records the removal and the furo floor.

### CI installs from the lock

Every job that installs the project uses `astral-sh/setup-uv@v10.0.1` with its cache enabled
and `uv sync --frozen`; `actions/setup-python` and `pip install --upgrade pip` are gone,
`js_test.yml` loses a hand-rolled pip cache, and the redundant `coverage xml` step goes
(`--cov-report=xml` already writes it). The cache key carries the python version and the
group, because the lock file it keys on is now identical across cells while the installed set
is not.

Every command after a sync is `uv run --no-sync`. This is not cosmetic: a bare `uv run`
re-syncs the default groups, and in `tests-no-mpl` that would reinstall the matplotlib the
job has just uninstalled and quietly stop testing anything.

All 13 Core cells stay, including the three Windows ones and the two excluded python 3.10
combinations. Only the axis name changes, from `sphinx-version: ["7.4","8.2","9.1"]` to
`sphinx-group: ["sphinx-7","sphinx-8","sphinx-9"]`.

Locking CI does give something up: the cells no longer resolve fresh on every run, so a bad
new release of pytest, myst-parser or sphinx will not break CI on the day it ships. That
signal is already covered — `.github/dependabot.yml` runs the uv ecosystem monthly, batching
minor and patch updates into one pull request and sending majors one each, and every one of
those re-runs the full matrix against the refreshed lock.

With one deliberate exception, added here: **dependabot now ignores major and minor updates
of `sphinx` alone.** The `sphinx-7/8/9` groups pin one series each *on purpose* — they are
the matrix, and which three series are tested is a decision taken by hand, together with the
CI cells and the `python_version` markers that go with them, not something a lock refresh
should do. Patch updates inside a series still flow, and nothing else is ignored.

### Four checks that fail loudly if this is broken later

- The lint job asserts that syncing `sphinx-7` and `sphinx-9` together is *refused*. If
  `[tool.uv] conflicts` is ever dropped and the tree made to lock again, this fails.
- It asserts that a plain `uv sync` resolves to sphinx ≥ 9, so a new pin in a
  non-conflicting group cannot silently cap the contributor default.
- Every test cell asserts that the sphinx it installed satisfies its group's own
  requirement, read out of `pyproject.toml` by `.github/scripts/check_sphinx_cell.py`. A
  group whose python marker excludes the cell's interpreter contributes nothing, the
  environment falls back to the project's `sphinx>=7.4,<10`, and the sync still exits 0 — so
  without this a cell could test another series, green. Checking against the specifier
  rather than against the group's name means `~=8.2` refuses the 8.1.3 that python 3.10
  falls back to, and that the check follows the matrix when the pinned series move. It is a
  script rather than an inline `python -c` so that the Windows cells carry no shell-quoted
  assertion; the only `-c` left in a cell is the one-line version print above it.
- `tests-no-mpl` asserts `import matplotlib` *fails* after the uninstall, before pytest runs.

All four were verified in both directions — passing on this branch, failing on a
deliberately broken tree.

### The lock diff

102 package entries, +409/−321 lines, +78 KB. Expect: sphinx gains 7.4.7 and 9.1.0 (8.1.3
and 8.2.3 were already there), myst-parser ×2, markdown-it-py ×2, furo bumped, `poethepoet`
and `pastel` added, and `tox`, `tox-uv`, `tox-uv-bare`, `virtualenv`, `filelock`, `distlib`,
`cachetools`, `pyproject-api`, `python-discovery`, `tomli-w` and `ruff` removed — `ruff`
because prek runs ruff-check and ruff-format at the identical 0.16.5 pin, so the separate
group was a second place for that pin to drift. `docutils` stays on one version. Most of the
byte growth is the conflict markers uv writes on every dependency edge of a package that now
exists in more than one version.

### Wall time

Locally, a warm-cache `uv sync` of a matrix cell takes ~0.3 s against ~8 s cold, versus
`setup-python` + `pip install --upgrade pip` + a full `pip` resolution and editable build on
a runner. **That predicts roughly 25–40 s saved per cell — several minutes of runner time per
run over 18 jobs — but it is an estimate from local timings, not a measurement.** The pytest
run itself is unchanged and dominates each cell, so the slowest cell's wall clock barely
moves; the win is runner minutes, cold-start latency and determinism. The first run on this
PR is the only honest confirmation. `setup-uv`'s `cache-python` (off by default) is the
obvious follow-up if interpreter downloads turn out to show up in that first run.

### What a contributor loses and gains against tox

Two things are genuinely lost. Tox's factor syntax generated the whole
`py{310,312}-sphinx{7,8,9}` matrix for free and `tox -p` ran it in parallel with a summary;
poe has no matrix construct, so each sphinx cell is a hand-written task, the interpreter is
`UV_PYTHON`, and running every cell locally is a sequence or a shell loop. In practice the
local sweep was rarely used (`envlist = py310` ran one cell by default; CI owned the matrix),
but it is a real capability drop. And tox's `basepython` pinned the docs environments to
3.12; the docs tasks currently rely on a documented `UV_PYTHON` note instead (poe's uv
executor has a per-task `python` option that restores this, if wanted).

Everything else is equivalent or better. Per-environment dependency sets become *locked*
groups instead of unlocked ad-hoc installs, so a contributor's `poe test-sphinx8` is
byte-for-byte the CI cell — something tox never gave. `{posargs}`, `setenv` and
`commands_pre` map onto argument pass-through, `env` and `sequence`. Environment recreation
is a 0.3 s idempotent sync rather than `tox -r`. Interpreters are downloaded by uv rather
than needing pyenv. Ad-hoc experiments that tox's `deps =` allowed still exist outside poe
(`uv run --with sphinx==9.0.0 pytest …`), just not locked.

### Why this is easier for coding agents

This was built and reviewed by agents, and the difference was visible. `uv run poe` gives an
agent every task with a help line in one command, instead of tox env names to guess out of
a config string. The locked environment means an agent tests in exactly CI's environment,
which removes a whole class of "green locally, red in CI" round trips. Failures are plain
pytest output, not a wrapper's. The two subtleties agents did trip on during the build are
now written into `AGENTS.md`: pass `--no-sync` to the outer `uv run` when setting
`UV_PYTHON` (otherwise uv rebuilds the default `.venv` on that interpreter), and editing
`pyproject.toml` requires committing the re-lock the `uv-lock` hook produces. The one thing
that made agents stumble before it was fixed — an old uv first on `PATH` being picked up by
poe's executor — is why `uv` itself is now pinned in the `dev` group.

### What to watch

The three Windows cells go through `uv sync` and `uv run --no-sync` for the first time. Every
step goes through `uv run` rather than a hard-coded `.venv/bin/...` path, which is what would
otherwise break on `.venv\Scripts`, and none of this was reproducible off a Windows runner.
The cell *names* change with the axis rename; branch protection here is wired to the `check`
(alls-green) job rather than to individual cell names, so nothing needs re-pointing — worth
confirming once against the repository settings.

Unaffected by this branch: dependabot #1811 (`actions/cache` → v6) still applies, to
`benchmark.yaml`'s benchmark-history cache; only `js_test.yml`'s pip cache goes away here.


ubmarco (approved):
Follow-up question rather than a blocker: are the tests anywhere executed against the
**built** package, so packaging problems — a missing data file, say — would surface?

I had assumed tox gave us that for free, since it does build an sdist and install the
wheel by default. But our `[testenv]` sets `usedevelop = true`, and every env inherits
it, so it has been `pip install -e .` all along. CI does the same today
(`pip install sphinx~=X -e .[test]`), and `uv sync` likewise installs the project
editable. So this PR changes nothing on that axis — I only want to record that the axis
was never armed, in case retiring tox looked like it was losing something here.

How much it matters for us: less than in a setuptools project, because `flit_core` packs
the whole `sphinx_needs/` directory and there is no `package-data` list to forget. I
built the wheel from master to check — 621 non-`.py` entries, `libs/html/**`, `css/**`,
`templates/*.html`, `images/feather_png/**`, `needsfile.json`, and no `__pycache__`.

The one place it could still bite is the sdist: `release.yaml` publishes with
`flit publish --use-vcs`, so a data file that exists in the working tree but was never
`git add`ed ships in the wheel and is missing from the sdist, and nothing would notice.
The closest thing to a packaging check we have is the Read the Docs build, which does a
real `pip install .` — and that is not a PR gate.

One thing worth knowing if we ever add such a gate: `uv sync --no-editable` on its own is
not enough. I tried it — the install is genuinely non-editable, but `tests/__init__.py`
makes pytest prepend the repo root to `sys.path`, so the source tree shadows the
installed copy and the run stays green while still testing the working tree.

If we want the coverage later, the cheap version is a wheel-contents assertion (diff the
wheel's file list against `git ls-files sphinx_needs`); the thorough version is a smoke
job that installs the built wheel into a clean environment and runs one small
`sphinx-build` from outside the repo. Either seems fine as a separate issue.

chrisjsewell:
Recorded the built-package question as #1829, with both options as you framed them.

PR #1831, opened 2026-09-03
✨ needpie / needbar: filter, status, tags and types scope the chart

chrisjsewell:
`needpie` and `needbar` now accept `:filter:`, `:status:`, `:tags:` and `:types:`. On a chart these options are a
**scope**: the set of needs the chart counts over, resolved once for the whole chart, with every content line or grid
cell then counted within it. A static number is never affected, and a chart without any of the options counts over
the whole project exactly as before.

```rst
.. needpie:: Open requirements by tag
   :status: open
   :types: req
   :labels: safety, other

   "safety" in tags
   "safety" not in tags
```

Each slice now counts only open requirements instead of every need in the project. The same options apply to every
cell of a `needbar`, and a `:filter:` given as an option is not split by the cell `:separator:`, so
`:filter: status in ['open', 'closed']` works as written.

The scope holds exactly what the same four options select on a `needlist` — ANDed, any-of within each option,
`:types:` matching the directive name or the type title, `:filter:` evaluated last over needs and parts — and a test
pins that equality. With the ubCode-only `cypher` option this completes the portable pairing: ubCode reads the query,
a Sphinx build ignores it and reads these options, so writing both makes the two tools count over the same needs.

Behaviour worth knowing, each pinned by a test:

- A `:filter-func:` on `needpie` receives the scoped needs, as the same view type as before (a new public
  `NeedsAndPartsListView.filter_id_complete` selects exactly by `id_complete`).
- A scope that selects nothing keeps each directive's existing empty state: a pie of only filter lines shows its
  `:filter_warning:` text, a bar chart draws all zeros.
- An invalid scope `:filter:` warns once per chart (`needs.filter`) and selects nothing. A scope restricts what a
  line counts, not what its filter can see: the `needs` variable inside a content line is unchanged.
- `c.this_doc()` works in a scope `:filter:`.

Not added: `sort_by`, `export_id` and `show_filters`; `filter-func` and `filter_warning` stay unavailable on
`needbar`. Known gaps: `needs_filter_max_time` and `needs_debug_filters` do not cover a chart's scope, and a syntax
error in a scope filter ends the build like any other filter string (#1772).

Docs: both directive pages gain a `common filters` section with a paired example, `docs/filter.rst` lists `needbar`
and rewrites the ubCode compatibility note for both charts, and the changelog gains an entry (with #1818's amended so
the two agree). Companion ubCode PR: https://github.com/useblocks/ubcode/pull/3262.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,9 +24,12 @@ Improvements
   none of the four counts over the whole project exactly as it did before.
 
   A scope holds what the same four options select on a view directive such as
-  :ref:`needlist`, so a chart and a list written with the same options cover the same
-  needs. On :ref:`needpie` a ``:filter-func:`` is handed the needs of the scope, so a
-  scope restricts what such a function counts from, not what it returns.
+  :ref:`needlist`, so a chart and a list written with the same options select the same
+  needs. On :ref:`needpie` a ``:filter-func:`` is handed the needs of the scope, as the
+  same view an unscoped chart hands it, so a scope restricts what such a function counts
+  from, not what it returns. The view gains one public method for that,
+  ``NeedsAndPartsListView.filter_id_complete``, which selects exactly the given needs and
+  parts by ``id_complete``.
 
   Neither directive gains an empty state it did not have: a pie of only filter lines
   whose scope selects nothing shows its ``:filter_warning:`` text, exactly as it does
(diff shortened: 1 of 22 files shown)


ubmarco (commented):
LGTM

PR #1839, opened 2026-09-03
🔧 Restructure the repository into a uv workspace (packages/sphinx-needs)

chrisjsewell:
### What this does

Turns this repository into a [uv workspace](https://docs.astral.sh/uv/concepts/projects/workspaces/):
the `sphinx-needs` distribution moves to `packages/sphinx-needs/`, with its importable code
under `src/`, and the repository root becomes a project that is never built or published
(`sphinx-needs-workspace`, `[tool.uv] package = false`, no `[build-system]`). The root
depends on every member, owns the dependency groups they share, and holds policy — the
lock, the lint/format/type-check/pytest configuration, the task definitions, the hooks and
the workflows. **Nothing about the published package changes**: same wheel contents (verified
file-for-file against PyPI 8.5.0's 706 entries), same sdist composition, same version, same
metadata apart from five README image URLs that had to be repointed, same release mechanics
(`flit publish --use-vcs`, one new `working-directory:` line).

```
sphinx-needs/
├── pyproject.toml           # workspace root: members, shared groups, conflicts, ruff, ty, pytest, poe
├── uv.lock                  # one lock for the workspace
├── .pre-commit-config.yaml  # one hook config
├── .readthedocs.yml         # stays here, under this name (see below)
├── .github/  docker/  scripts/  codecov.yml  context7.json
├── AGENTS.md  CLAUDE.md  README.md  LICENSE   # CLAUDE.md imports AGENTS.md, then adds the Claude Code cloud setup
├── .claude/settings.json    # attribution off: no Co-Authored-By, PR footer or session link
└── packages/
    └── sphinx-needs/
        ├── pyproject.toml   # [project] and extras -- the groups are the root's
        ├── src/sphinx_needs/
        ├── tests/  docs/  performance/  package.json
        └── README.rst  LICENSE  AUTHORS  AGENTS.md  CLAUDE.md
```

### What changes for contributors

Two commands:

```bash
uv sync --frozen   # every package, plus the shared development and test dependencies
uv run poe         # every task, with its help
```

* **No group has to be named.** The test tooling is a dependency group of the workspace
  root, and the root's default `dev` group includes it, so a bare `uv sync` is already an
  environment the suite runs in.
* **Dependency groups moved to the root, and kept their pre-move names** (`test`,
  `benchmark`, `sphinx-7/8/9`, `typing`). A group cannot be composed across the root/member
  boundary in either direction, so a shared group can only live in one place; the root is
  the place a plain `uv sync` reaches. The package's `pyproject.toml` is now `[project]` and
  its extras, nothing else.
* **Task names:** repository-wide tasks stay bare (`lint`); package tasks end in the
  package's short name — `test-needs`, `docs-needs`, `typecheck-needs`,
  `test-needs-sphinx8`, `benchmark-needs`. Task names keep the `-needs` suffix because they
  act on the package; the *group* names are the root's and are unprefixed.
* **A path you pass to a package task is package-relative.** Package tasks run with
  `packages/sphinx-needs` as their working directory, so it is
  `uv run poe test-needs tests/test_basic_doc.py`, not the repository-relative path. The
  tasks themselves carry `--ignore=performance` rather than naming `tests`, because a path
  in the task's own command would be *added* to the one you pass instead of replaced by it.

`AGENTS.md` at the root lists the rest, including the incantations that look right and are
not (`uv sync --all-groups`, `uv sync --package sphinx-needs`,
`cd packages/sphinx-needs && uv sync`).

### Why the root has a name

The root is a project rather than a virtual one, and that is not cosmetic. In a uv
workspace `[tool.uv] conflicts` requires a `package =` qualifier on every entry, and a
virtual root has no name to give — so a virtual root cannot declare its own dependency
groups conflicting at all. It has to, because the shared `test` group and the sphinx matrix
must resolve in the *same* conflict space: split across root and member, `myst-parser`
locks to two versions that only the member's conflict extras tell apart, and a bare
`uv sync` fails with `Found multiple packages matching myst-parser`. Naming the root also
means `uv sync` syncs *that project*, which is why `dependencies = ["sphinx-needs"]` lists
every member — that line is what keeps a bare sync installing them. The root is never built
and never published: no `[build-system]`, `package = false`, and `uv build --all-packages`
skips it and builds only `sphinx-needs`.

One consequence worth knowing: `--extra` resolves against the *project*, so the root
re-exports each of the member's extras (`docs = ["sphinx-needs[docs]"]`). A new extra on a
member needs a line there before any task can ask for it.

### Rebasing your PR after the directory move

```
git fetch origin
git rebase -X find-renames=15% origin/master
```

`-X find-renames=15%` is important: without it, files that have drifted since you branched
come back as "deleted by us" conflicts at their OLD path, and resolving those the obvious
way silently drops your change.

Do NOT use `git rebase --apply`, `git am`, or a downloaded `.patch` — those do no rename
detection at all, and a NEW file you added under `tests/`, `docs/` or `sphinx_needs/` will
land at the old path with no conflict and no warning.

Then:
* **content conflicts** (`UU`) are already at the new `packages/sphinx-needs/…` path —
  resolve them as usual;
* **`CONFLICT (file location)`** on a file you added: git has already put it at the right
  new path, just `git add` it;
* **`pyproject.toml`** — the ONLY special case. The root `pyproject.toml` is now the
  *workspace* file and has nothing to do with your change. Run
  `git checkout --ours pyproject.toml && git add pyproject.toml`, then apply your change
  by hand to `packages/sphinx-needs/pyproject.toml`.
* **`rename/delete`** (only if your PR deletes or adds files the move touched) — `UD` are
  files you deleted that the move renamed, `UA` are files you added:

      git status --porcelain | awk '/^UD /{print $2}' | xargs git rm --
      git status --porcelain | awk '/^UA /{print $2}' | xargs git add --

### How to review this

With rename detection the diff is small: **1407 files — 1375 pure renames, 7 renames that
also change content, 18 edits and 7 new files, so 32 files to actually read.** Without it,
the same diff is **2789 files and 276 000 lines** — which is what `git log -p`, a `.patch`
download, or any client that does not pass `-M` will show you. Use:

```bash
git diff -M origin/master...HEAD --stat     # the 1407/32 view
git log --stat -M --follow <path>           # a file's history across the move
```

Exactly one file under `tests/` changes content: `tests/schema/test_generate_schema.py`
located its snapshot directory by counting three parents up from itself, which used to reach
the repository root and now reaches `packages/sphinx-needs` — one level short of `src/`. It
resolves the directory through the package instead. Nothing else under `tests/` or
`src/sphinx_needs/` is touched.

The first commit is a **pure `git mv`**: 1384 files, `0 insertions(+), 0 deletions(-)`, every
rename at 100% similarity. GitHub renders those as one line each with no diff body, so the
whole of the move itself is nothing to read. The seven commits are: the rename, the
`pyproject.toml` split, the path/name fix-ups, `uv lock`, the `CLAUDE.md` shims
(`@AGENTS.md` at the root and in the package — Claude Code reads `CLAUDE.md`, not
`AGENTS.md`), the root `AGENTS.md` additions from the cloud-agent trials of this branch
(machine prerequisites and network hosts, the interpreter-decides-the-sphinx trap, rough
runtimes, where `ty` lives, and that `benchmark-needs` is red for a pre-existing reason,
#1840), and the split of the Claude-specific findings out of `AGENTS.md` into `CLAUDE.md`
(the cloud environment recipe and the Cypress download workaround) plus
`.claude/settings.json`, which turns off the attribution trailers cloud sessions add.

Two things that look like regressions and are not:

* **`README.md` replaces `README.rst` as the front page.** flit requires the package's
  `readme` file to live inside the package, so `README.rst` moved with it; the new root
  `README.md` is a router to the package.
* **`git log <file>` (GitHub's per-file *History* button) now shows only the commits since
  the move.** `git log --follow` recovers the full history and `git blame` follows the
  rename with no flag at all.
* **Codecov's per-file history restarts at the merge.** Every path becomes
  `packages/sphinx-needs/src/sphinx_needs/…`, so Codecov sees new files. `codecov.yml`, the
  `--cov=sphinx_needs` flag (a module name, not a path) and the upload step are unchanged,
  and `coverage.xml` still lands at the repository root.

### Two checks that go red on this PR and self-heal on merge

* **Docker-Image** — the Dockerfile's tagged install now carries
  `#subdirectory=packages/sphinx-needs`, which is correct for every tag cut from this layout
  but wrong for `NEEDS_VERSION: 8.5.0`, a tag that predates it. From now on `NEEDS_VERSION`
  must name a tag cut *after* this move; the next release bumps it (as every release does)
  and the check goes green. Note what the `pull_request` run does and does not prove: it
  proves the `docker/**` path filter fires and builds this literal version — it never
  exercises the `pre-release` branch of the Dockerfile at all. Without the fix, the breakage
  would instead be discovered *while publishing* 8.6.0.
* **Docs-Linkcheck** — three GitHub deep links in `docs/schema/index.rst` now point at
  `master/packages/sphinx-needs/tests/…`, which 404s until this merges. Leaving them at the
  old path would have been the silent failure: they check out fine today and break on every
  docs PR afterwards.

Neither is part of the required `check` gate.

### Companion changes outside this repository

* **`useblocks/sphinx-mounts`** — its `test` dependency group pins
  `sphinx-needs @ git+https://github.com/useblocks/sphinx-needs.git@master`, which stops
  installing the day this merges. Its own TODO ("swap for `sphinx-needs>8.3.0` once
  released") is already satisfied — #1750 shipped in 8.3.1 — so the fix is to drop the pin.
  **This should land before this PR.**
* **`useblocks/ubcode`** — `scripts/check_fixture_drift.py` points at
  `<upstream>/tests/schema`; it needs `packages/sphinx-needs/tests/schema` (best: probe both,
  since it clones the default branch). `workflow_dispatch`-only, so it can follow.
* **`useblocks/awf`, `useblocks/awesomesphinx`** — bare `git+https://…/sphinx-needs` in their
  docs requirements; low priority, PyPI or `#subdirectory=` fixes them.
* **`AlexanderLanin/features`** (a third-party devcontainer feature) installs
  `git+…/sphinx-needs` for `VERSION=dev` — worth a courtesy issue.

### Deliberately NOT changed

* **the release backend** — still `flit publish --use-vcs`, one new `working-directory:`.
  `uv build` would resolve flit 4, change `Metadata-Version` and drop `tests/` and `docs/`
  from the sdist; that belongs to the release arc, not to a move.
* **the tag scheme** — still `[0-9]+.[0-9]+.[0-9]+`. Prefixed tags matter once a second
  distribution shares the repository.
* **the version** — still dynamic, read from `src/sphinx_needs/__init__.py`.
* **`py.typed`** — a public API commitment, and its own pull request.
* **`requires-python`, the extras, the ruff ruleset, and anything under `sphinx_needs/`** —
  untouched.


ubmarco (approved):
Let's do this!

chrisjsewell:
## Dev-workflow validation from a Claude Code web session

I'm an agent running in a Claude Code cloud session, asked to exercise every development activity on this branch and report what a fresh web agent trips over. Commit `d39739a`, Ubuntu 24.04 container, default interpreter **Python 3.11.15**.

**Short version: the layout works. Every task resolves, runs from the documented directory, and takes the documented arguments.** Three things fail here, and none of them are the layout — two are the container's network policy and one is a bug that predates the move.

### What worked

| Activity | Result |
| --- | --- |
| `uv sync --frozen` | 2.8s (warm uv cache) |
| `uv run poe` | all 18 tasks listed |
| `uv run poe lint` | 7/7 prek hooks, 15s |
| `uv run poe typecheck-needs` | pass; creates `.venvs/typing` on first run |
| `uv run poe test-needs` | **1740 passed, 12 skipped, 2 deselected** — 7m47s |
| `uv run poe build-needs` | 2s — 2.5 MB sdist + 2.8 MB wheel |
| `uv run poe smoke-needs` | 6s, 6/6 checks on the built wheel |

Targeted runs all behave, with paths relative to `packages/sphinx-needs` exactly as documented — a single test id, a single file, a directory (`tests/schema`), a whole module, and `-k` expressions. CI's other invocation style (from the root, with `packages/sphinx-needs/tests/...` paths) works too, so both styles are available.

All three matrix cells run: `UV_PYTHON=3.10 … test-needs-sphinx7`, `3.11 … sphinx8`, `3.12 … sphinx9`. The container ships 3.10/3.11/3.12/3.13 and uv downloads others on demand, so nothing is blocked there. The Python-3.12 gate behaves precisely as `AGENTS.md` warns: `tests/schema/test_generate_schema.py` gives 8 passed on 3.12 and 8 *silently* skipped on 3.11.

The three CI lint-job canaries all behave: the type-gate probe fails naming the probe, syncing `sphinx-7` + `sphinx-9` together is rejected by the conflicts set, and `check_typing_floor.py` reports `sphinx 7.4.7 / docutils 0.20.1 / types-docutils 0.20.0.20240406`.

**Special mention for the "Incantations that do NOT work here" section** — all four reproduced exactly as written (`uv sync --all-groups`, root `uv build`, `uv sync --package sphinx-needs` leaving no pytest/poe/prek, and `cd packages/sphinx-needs && uv run pytest` erroring on `performance/`). That section saved me from three separate wrong turns; it's the most useful thing in the file.

### What didn't

**1. `poe docs-needs` — container network policy, not the docs.** The session proxy returns 403 for `docs.python.org` and `www.sphinx-doc.org`, intersphinx can't fetch either inventory, and under `-nW` the ~68 resulting unresolved references become errors. Suppressing `ref,intersphinx` on top of `conf.py`'s own `suppress_warnings` list leaves exactly the two "failed to reach any of the inventories" warnings and nothing else — so the docs are clean, they just can't be *verified* clean without those two hosts. (`needservice.rst` also hits a blocked `api.github.com`.)

**2. `poe test-needs-js` — impossible here.** `npm install cypress` dies with `ECONNRESET`; `download.cypress.io` is 403 as well. Nothing the repo can do about it.

**3. `poe benchmark-needs` — 12 failed, 13 passed, and this one is a real pre-existing bug.**

```
FAILED tests/benchmarks/test_querying.py::test_filter_single_need_simple[100]  (+11 more)
E   AttributeError: Mock object has no attribute 'variant_data_proxy'.
    Did you mean: 'variant_data_file'?
```

`config.variant_data_proxy` is assigned dynamically at `src/sphinx_needs/needs.py:702` rather than declared as a `NeedsSphinxConfig` field, so it isn't present on the `Mock(spec=NeedsSphinxConfig)` at `tests/benchmarks/test_querying.py:113`. Broken since #1787 (2026-08-22 — confirmed with `git log --follow -S variant_data_proxy`), i.e. well before this branch. CI never catches it because the benchmark job runs `-k '_time or _memory'`, which collects **3 of the 25** tests; the 12 failures are all outside that filter. Worth its own issue rather than anything on this PR.

### One flake worth knowing about

On my *first* full-suite run I had a docs build and a wheel build going concurrently, and `tests/test_needflow.py::test_doc_build_html[plantuml]` failed its `assert warnings == ""`. It passes alone, passes as a whole module (120 tests), and the clean serial re-run was green (the 1740 above). The plantuml/java path appears load-sensitive, and a zero-warnings assertion has no tolerance for it — relevant if the suite is ever run with `-n auto` or alongside anything else heavy.

### Suggestions for `AGENTS.md`

The file is already unusually good — the directory table, the naming rule, the package-relative-paths warning and the negative-incantations section all did their job without me having to read `pyproject.toml` first. Five gaps I actually hit, roughly in order of how much time each cost:

**1. The default interpreter silently decides which sphinx you test.** This is the big one. On this container `python3` is 3.11, so a bare `uv sync` resolves **sphinx 8.2.3**, not 9.x — meaning `poe test-needs` is quietly a *second sphinx-8 run*, and CI's "the default environment resolves to the newest sphinx" canary fails locally for that reason alone. It also means the Python-3.12-gated schema test skips by default. Both disappear with `UV_PYTHON=3.12`, which I verified gives sphinx 9.1.0 in the default env *and* un-skips those 8 tests, while explicit per-cell `UV_PYTHON` still overrides it. Suggested addition near the commands block:

> The default environment's sphinx follows whichever interpreter uv picks: on a machine whose default python is < 3.12 a bare `uv sync` caps at sphinx 8.2, so `test-needs` silently repeats the sphinx-8 cell and the 3.12-gated tests skip. Export `UV_PYTHON=3.12` (or newer) to make the default environment the newest-sphinx cell that CI's canary expects.

**2. Two tasks need specific hosts reachable.** The prerequisites paragraph covers `java`, `dot` and Node — which was accurate and necessary, `dot` really is absent here — but not network. Worth extending it: `docs-needs` needs `docs.python.org` and `www.sphinx-doc.org` for intersphinx or it fails under `-nW` with dozens of unresolved references that look like a docs regression, and `test-needs-js` additionally needs `download.cypress.io` to install cypress at all. In a sandboxed environment that's several minutes of misdiagnosis.

**3. Say that `benchmark-needs` is currently red.** It's listed alongside the other tasks with no caveat, so an agent told to "check every task" will run it, get 12 failures, and reasonably suspect itself or this branch. One parenthetical — "(currently failing on `test_querying.py`, unrelated to the layout — see #xxxx; CI only runs `-k '_time or _memory'`)" — prevents that entirely.

**4. `ty` isn't on `PATH`.** The ty paragraph says `cd packages/sphinx-needs && ty check`, `ty check packages/sphinx-needs` and an editor all give the gate's result — all three of which I confirmed true — but `ty` ships only in the `typing` group, so on a fresh checkout there's no `ty` command until `poe typecheck-needs` has created `.venvs/typing`, and even then it's `.venvs/typing/bin/ty`, not on `PATH` and not in `.venv`. Worth naming that path, since the paragraph currently reads as though `ty` is just available.

**5. Expected runtimes.** For an agent this is load-bearing, not trivia: it decides whether to background a command and what timeout to give it. Something as compact as *lint 15s · typecheck <1s · build 2s · smoke 6s · docs ~1m40s · full suite ~8min · benchmark ~4min* would stop agents either backgrounding a 15-second lint or foregrounding an 8-minute suite into a timeout.

Two smaller ones: `docs-needs` carries `--keep-going`, so it prints a summary line and exits 1 — piping it through `tail` throws the exit code away and it reads as success (this bit me); and the load-sensitivity above is worth a sentence next to the `java`/`dot` prerequisites, since "run the suite serially" is the kind of thing that only shows up as a mystery failure.

### Cloud-environment configuration

For anyone else running this repo in a web session, the container needs one package and benefits from one variable:

```bash
apt-get install -y graphviz     # dot is absent; needflow tests FAIL rather than skip
```

`java`, Node, npm and Python 3.10–3.13 are all pre-installed, and `prek`/`poethepoet` come from the root `dev` group via `uv sync`, so nothing else has to be provisioned. Setting `UV_PYTHON=3.12` as an environment variable fixes item 1 above. `docs.python.org`, `www.sphinx-doc.org` and `download.cypress.io` are not on the default network allowlist and have to be added explicitly for `docs-needs` and `test-needs-js`.

Happy to open the benchmark-mock issue, or push the `AGENTS.md` additions above to this branch, whichever is more useful.

---

chrisjsewell:
## Re-run after the container + `AGENTS.md` changes

Second Claude Code web session, this time on commit `9c3aa8f` (so with the `AGENTS.md`
additions in place) and on the reconfigured environment. **Two of the three previous
failures are gone; the third is unchanged and expected. One new, more precise diagnosis.**

The environment now reads:

* **Allowed domains** — `docs.python.org`, `www.sphinx-doc.org`, `api.github.com`,
  `download.cypress.io`, plus the default package-manager list
* **Setup script** — `uv tool install pre-commit --with pre-commit-uv` / `prek` /
  `poethepoet`, `cargo install cargo-insta --locked`, `apt-get install -y graphviz`
* **Environment variables** — none set

### Results, against the previous run

| Task | Previous run | This run |
| --- | --- | --- |
| `uv sync --frozen` | 2.8s | 2.0s |
| `uv run poe` | 18 tasks | 18 tasks |
| `lint` | 7/7, 15s | 7/7, 21s |
| `typecheck-needs` | pass | pass |
| `test-needs` | 1740 passed, **12 skipped** | **1748 passed, 4 skipped**, 8m21s |
| `build-needs` | 2s | 1.9s |
| `smoke-needs` | 6/6, 6s | 6/6, 5.4s |
| **`docs-needs`** | **failed**, ~68 unresolved refs | **exit 0, zero warnings, 1m28s** |
| `test-needs-js` | blocked | still blocked (see below) |
| `benchmark-needs` | 12 failed, 13 passed | 12 failed, 13 passed, 4m27s |

**`docs-needs` is the big one.** With `docs.python.org` and `www.sphinx-doc.org`
allowlisted, intersphinx resolves, and the build is clean under `-nW --keep-going` —
`build succeeded`, exit 0, not a single warning. So the docs really were clean; last time
they just couldn't be verified. `api.github.com` being allowed also clears the
`needservice.rst` fetch.

**`graphviz` in the setup script works** — `dot` 2.43.0 is on `PATH` from session start, so
the needflow tests run rather than failing. Spot-checked on the oldest cell too:
`UV_PYTHON=3.10 … test-needs-sphinx7 tests/test_needflow.py` gives 120 passed.

**The `UV_PYTHON=3.12` paragraph does its job.** A bare `uv sync` here still gives Python
3.11.15 / sphinx 8.2.3, exactly as the new text warns; exporting `UV_PYTHON=3.12` gives
3.12.3 / sphinx 9.1.0. That is the whole of the `test-needs` delta above — the 8 extra
passes are `tests/schema/test_generate_schema.py`, which silently skipped last time. Since
it is not set as an environment variable, it still has to be exported by hand; worth adding
`UV_PYTHON=3.12` to the Environment variables box if you want the default session to be the
cell CI's canary expects.

**`benchmark-needs` reproduces identically** — 12 failed / 13 passed, same
`AttributeError: Mock object has no attribute 'variant_data_proxy'`. Expected: #1840 is
filed and `AGENTS.md` now says so, which is exactly what stopped me suspecting the branch
this time.

### The one thing still broken, and why the allowlist entry didn't take

`npm install cypress` still dies with `ECONNRESET`. The cause is more specific than "the
host is blocked": `download.cypress.io` **is** now reachable (302), but it is only the
redirector. The binary lives on **`cdn.cypress.io`**, which is not on the list:

```
$ curl -I 'https://download.cypress.io/desktop/16.0.0?platform=linux&arch=x64'
HTTP/2 302
location: https://cdn.cypress.io/desktop/16.0.0/linux-x64/cypress.zip

$ curl https://cdn.cypress.io/
curl: (56) CONNECT tunnel failed, response 403
```

The npm package itself installs fine; it is the postinstall binary download that fails, and
because the proxy rejects the tunnel rather than returning a status the script can read, it
surfaces as `ECONNRESET` — which looks like a flaky download, not a policy block. **Adding
`cdn.cypress.io` should be all `test-needs-js` needs.** I have pushed a one-paragraph
`AGENTS.md` fix for this (`a25ac1f`), since the file currently says only
`download.cypress.io`.

### Two notes on the setup script

Not problems, just dead weight:

* `uv tool install prek` and `uv tool install poethepoet` are redundant — both are in the
  root `dev` group, so a bare `uv sync` already puts `prek` and `poe` in `.venv`. `uv tool
  install pre-commit --with pre-commit-uv` is redundant in a stronger sense: this repo runs
  `prek`, not `pre-commit` (`poe lint` is `prek run --all-files`).
* `cargo install cargo-insta --locked` has nothing to do with this repo — no `Cargo.toml`,
  no Rust, and snapshots are syrupy. It is also the slowest line in the script.

Dropping all four would leave `apt-get install -y graphviz`, which is the only line that
this repository actually needs.

### `AGENTS.md` accuracy check

All five additions from last time hold up against a cold re-read:

* the interpreter/sphinx trap — verified in both directions
* the runtimes — `lint` 21s vs "15 s", `docs-needs` 1m28s vs "under 2 min", `test-needs`
  8m21s vs "7–8 min", `benchmark-needs` 4m27s vs "4 min". All close enough to size a
  timeout with
* `ty` at `.venvs/typing/bin/ty`, absent from `PATH` and from `.venv` — exactly right, and
  it is created by the first `typecheck-needs` as described
* `benchmark-needs` red with the #1840 reference — saved the misdiagnosis it was meant to
* `docs-needs` exiting 1 under `--keep-going` — still true, though it did not bite this
  time since the build was clean

The only inaccuracy was the cypress host, now fixed.

---

chrisjsewell:
## `test-needs-js` is fine in the new layout

With `cdn.cypress.io` allowed I got it running: **2 passed in ~18 s**, from a genuinely clean slate (no `node_modules`, empty `~/.cache/Cypress`). Nothing in the harness needed changing — `tests/conftest.py` resolves `spec_pattern`, `supportFile` and `cypress.config.js` relative to its own file, so the directory move never touched them.

## …but `npm install cypress` still can't complete on its own

The allowlist change worked (no more `ECONNRESET` from the 302). What's left is a separate problem: the transfer is reset a few MB from the end, and cypress's installer **hides** it rather than causing it. Reproduced with cypress's own HTTP client (`@cypress/request` + `request-progress`) against the same URL through the same proxy:

| client | transfer | result |
|---|---|---|
| node, drain to `/dev/null` | 250 MB | 3/3 `ECONNRESET` at 95.8–97.3% |
| node, pipe to file (what the installer does) | 250 MB | 3/3 `ECONNRESET` at 95.7–98.9% |
| node, `Range: 0-10MB` / `0-50MB` | 10 / 50 MB | complete |
| node, `Range: 0-100MB` | 100 MB | `ECONNRESET` at 90.9% |
| curl, HTTP/2 **and** forced HTTP/1.1 | 250 MB | complete, ~97 MB/s |

So it isn't a byte cap, a policy denial (nothing logged against `cdn.cypress.io`), the HTTP version, or consumer backpressure — the drain-to-`/dev/null` row rules that last one out.

<details>
<summary>Why it surfaces as <code>Corrupted download</code> instead of a reset</summary>

Three things in `downloadFromUrl` (`node_modules/cypress/dist/cli-*.js`) line up:

```js
.on('error', (e) => {
    if (e.code === 'ECONNRESET')
        return; // sometimes proxies give ECONNRESET but we don't care
    reject(e);
})
```

The real error is discarded on purpose. Then node fires `end` *after* `aborted`, so the installer's completion gate — `Bluebird.all([writeStream 'close', response 'end'])` — resolves on a truncated file:

```
[2.3s] EVENT response "aborted"  bytes=243286464
[2.3s] EVENT response "error"    ECONNRESET aborted
[2.3s] EVENT response "end"      bytes=243286464
[2.3s] EVENT writeStream "close" size=243286464
```

The only integrity check is after the fact, against the `x-amz-meta-size`/`x-amz-meta-checksum` headers the CDN sends. Hence the misleading label, the percentage that moves every attempt, and no resume or retry anywhere in the path — which is why retrying `npm install cypress` never converges.

</details>

## Suggested change to the recipe in `AGENTS.md`

The branch already carries this note (the *why* paragraph above it is accurate as committed); this refines the commands. The version I pushed makes you fill in a `<version>` placeholder, still starts the doomed 250 MB download, and dirties `package.json`. Verified replacement for the code block and the paragraph after it:

```bash
cd packages/sphinx-needs
CYPRESS_INSTALL_BINARY=0 npm install cypress --no-save --no-package-lock
V=$(node -p "require('cypress/package.json').version")
curl -fL -C - --retry 3 --retry-all-errors -o /tmp/cypress.zip \
  "https://download.cypress.io/desktop/$V?platform=linux&arch=x64"
CYPRESS_INSTALL_BINARY=/tmp/cypress.zip npx cypress install --force
```

> `CYPRESS_INSTALL_BINARY=0` stops the doomed download before it starts; the version then
> comes from the package npm actually resolved, so the binary cannot mismatch it, and
> `download.cypress.io/desktop/<version>` redirects to the right platform path on its own.
> `--no-save --no-package-lock` leaves the working tree clean, so there is nothing to revert
> afterwards. Cypress checksums whatever zip it is handed, so a bad resume is still caught.

That also lets the note drop its closing sentence about reverting `package.json` and `package-lock.json`.

## For a web-container setup script

The same five lines are all that's needed — no repo-side setup, no hook. `java`, `dot` and `node` are already in the image. Two notes if it goes into a container setup script:

- It costs ~250 MB and well under a minute, for a path that is 2 tests and that CI runs in its own workflow (`js_test.yml`), so pre-warming every container may not be worth it versus running the block on demand.
- During the run, `api.cypress.io`, `cloud.cypress.io` and `redirector.gvt1.com` each get a 403 from the proxy. That's cypress telemetry and an Electron component check — **harmless, the tests pass regardless**. Worth knowing so nobody chases them as the cause of a failure.

---

chrisjsewell:
## Correction: the workaround does not need to be repo-specific

My previous comment put a `cd packages/sphinx-needs` recipe in the container setup, which wrongly couples a generic environment to this repo. That was the wrong split. The fragile 250 MB part is the **binary**, and cypress caches that globally in `~/.cache/Cypress/<version>/` — only the cheap `npm install` is repo-local. So the two halves separate cleanly:

**Generic, for a container setup script — nothing in it refers to this repository:**

```bash
V=$(npm view cypress version)
curl -fL -C - --retry 3 --retry-all-errors -o /tmp/cypress.zip \
  "https://download.cypress.io/desktop/$V?platform=linux&arch=x64"
CYPRESS_INSTALL_BINARY=/tmp/cypress.zip npx --yes cypress@$V install --force
rm -f /tmp/cypress.zip
```

**Repo-side: nothing new at all.** With that cache warm, the ordinary documented command works unaided — cypress's postinstall finds the binary and skips the download:

```console
$ rm -rf node_modules && npm install cypress     # no env vars, no workaround
added 171 packages
$ uv run poe test-needs-js
2 passed, 1752 deselected in 19.97s
```

Verified in both directions: seeded the cache from `/tmp` with the repo untouched, then wiped `node_modules` and ran the plain install — green, and `git status` clean.

### So `AGENTS.md` needs less than I suggested, not more

Please **disregard the recipe in my previous comment**. Since the workaround belongs to the environment, the note only has to tell a future agent how to recognise the failure and where the fix lives — no commands, no placeholder version, no "revert `package.json`" step. Something like:

> The Cypress tests (`-m jstest`, which `test-needs` excludes) additionally need Node, and
> `npm install cypress` — run it in `packages/sphinx-needs`, where the `package.json` is —
> pulls a ~250 MB binary from `download.cypress.io`, which 302s to `cdn.cypress.io`; an
> allowlist naming only the first gets an `ECONNRESET` from the postinstall script, not a
> 403. If it instead reports `Corrupted download` with a computed size short of the expected
> one, the transfer is being truncated a few MB from the end and the installer is hiding it:
> it swallows `ECONNRESET` deliberately, and node fires `end` after `aborted`, so a partial
> file counts as a finished download and only the trailing `x-amz-meta-checksum` comparison
> notices. There is no resume and no retry, so retrying never converges. The binary is cached
> globally in `~/.cache/Cypress/<version>/`, so this is a machine-setup problem, not a repo
> one — seed that cache with a downloader that resumes and the plain `npm install` then skips
> the download entirely.

### One residual coupling, worth naming

The seed resolves `npm view cypress version` — latest at container build time. Nothing here pins cypress (`package.json` carries no dependency on it), so a later `npm install cypress` resolves the same latest and the two agree. **If cypress ever gets pinned in `package.json`, the seed has to name that version** or the postinstall will try to fetch the pinned one and hit the same reset. That's a value to keep in step, not a procedure — but it is the one thing that would re-couple the environment to the repo, so it is worth remembering if pinning is ever added.

---

PR #1844, opened 2026-09-04
⬆️ Raise the Python floor to 3.11, and pin the workspace interpreter

chrisjsewell:
## Why now

Python 3.10 reaches [end of life](https://devguide.python.org/versions/) in **October 2026**,
so the supported floor moves to **3.11**. `sphinx-mounts` already came down to 3.11
(useblocks/sphinx-mounts#45), so the two repositories now agree.

`requires-python = ">=3.11,<4"` moves in **both** manifests at once — the workspace root and
`packages/sphinx-needs` — because the root's own comment on that line says they move together.
As of this PR something checks that (see *The pin and its fences*).

## What a user on Python 3.10 experiences: nothing breaking

`pip` has honoured `Requires-Python` since 9.0 and PyPI serves it in the JSON API, so
`pip install -U sphinx-needs` on 3.10 resolves to **the newest release that still accepts
3.10 — 8.5.0 — and installs it without an error**. It is a version pin, not a failure.

Read out of the wheel this branch builds:

```
Requires-Python: >=3.11,<4
Classifier: Programming Language :: Python :: 3.11
Classifier: Programming Language :: Python :: 3.12
Classifier: Programming Language :: Python :: 3.13
Classifier: Programming Language :: Python :: 3.14
Requires-Dist: sphinx>=7.4,<10
Requires-Dist: requests-file>=2.1,<4.0
Requires-Dist: requests~=2.32
Requires-Dist: jsonschema-rs>=0.37.1,<0.53.0
Requires-Dist: sphinx-data-viewer~=0.1.5
Requires-Dist: sphinxcontrib-jquery~=4.0
Requires-Dist: typing-extensions>=4.14.0
Requires-Dist: minijinja~=2.15
…
Requires-Dist: sphinx-syntax-example~=0.2.0 ; extra == "docs"
```

Two visible consequences for an installed package, both in that block:

- **`tomli` is gone from the dependencies.** `needs_from_toml` reads the standard library's
  `tomllib`, which Python 3.11 added, so the conditional import in `needs.py` collapses to a
  plain `import tomllib`.
- **the `docs` extra's `sphinx-syntax-example` no longer carries a python marker**, so the
  extra installs in full wherever it resolves.

## The CI matrix: two excludes become one, and it is not the one you would guess

The core matrix is now `["3.11", "3.12", "3.13", "3.14"] × [sphinx-7, sphinx-8, sphinx-9]`,
and the whole `exclude:` block is a single rule: **`ubuntu-latest × 3.11 × sphinx-9`**.

The old comment said *"Sphinx 8.2 only supports py3.11+"*. That was true and it justified the
`3.10 × 8.2` exclude, which is now moot — 3.11 is the floor. It never explained the second
exclude, and the real rule is stricter: **sphinx 9.1 requires Python 3.12**. So the `sphinx-9`
dependency group keeps its `python_version >= '3.12'` marker while `sphinx-8` loses its one.

What that costs on 3.11 is worth spelling out, because it is *not* an install failure:

```
$ UV_PYTHON=3.11 uv sync --frozen --no-default-groups --group test --group sphinx-9
$ UV_PYTHON=3.11 uv run --no-sync python -c "import sphinx; print(sphinx.__version__)"
8.2.3                                                     # exit 0 — the group contributed nothing
$ UV_PYTHON=3.11 uv run --no-sync python .github/scripts/check_sphinx_cell.py sphinx-9
::error::sphinx-9 asks for sphinx~=9.1, but 8.2.3 is installed     # exit 1
```

The marker leaves the group empty, the environment falls back to the project's own
`sphinx>=7.4,<10`, the sync succeeds, and only `check_sphinx_cell.py` notices that a job
called `sphinx-9` is testing 8.2.3. Excluding the cell is the fix; the comment in `ci.yaml`
now says exactly this. On 3.12 the same two commands give sphinx 9.1.0 and exit 0, which is
why the exclude is scoped to 3.11 rather than to the group.

Windows keeps three corner cells and the floor stays among them (`3.11 × sphinx-7`,
`3.12 × sphinx-8`, `3.14 × sphinx-9`); the JS matrix's floor cell moves 3.10 → 3.11; the
`smoke-needs` job — the one that installs the built wheel at the declared floor — moves to
3.11; `release.yaml` moves off 3.10 to 3.12 (it only runs flit for a pure-python wheel, so it
now matches every other utility workflow). 11 ubuntu + 3 windows = 14 cells, one more than
before, all of them ubuntu and therefore free in wall-clock terms.

## `tomli` is still in `uv.lock`, and that is correct

Grep the lock and you will find it. sphinx-needs' own edge is gone; what remains is
**`coverage[toml]`'s**:

```
[package.optional-dependencies]
toml = [{ name = "tomli", marker = "python_full_version <= '3.11' or …" }]
```

Note the `<=`, not `<`: under PEP 440 `3.11.14 > 3.11`, so that marker is true only for
CPython **3.11.0 exactly**. On 3.11.14 the synced test environment contains no `tomli` at all:

```
$ UV_PYTHON=3.11 uv sync --frozen --no-default-groups --group test --group sphinx-7
$ uv pip list --python .venv/bin/python | grep -i tomli
                                                    # nothing
$ .venv/bin/python -c "import sphinx_needs.needs"   # fine — stdlib tomllib
```

## `UP042`: the two `(str, Enum)` classes become `StrEnum`, and here is the audit

`[tool.ruff]` sets no `target-version`, so it follows `requires-python`; moving the floor
turns on 8 findings. Five are `UP035` (`typing_extensions.NotRequired`/`Required` →
`typing`) and one is the `UP036` that removes the `tomllib` version block — all mechanical.
The two `UP042`s are not, so they were audited rather than assumed:

`StrEnum.__str__` returns the member's **value**; a `(str, Enum)` mixin's returns
`ClassName.MEMBER`. That is the entire behavioural difference, and it only bites where a
member is *formatted*. Neither class is ever formatted:

- **`ExtendType`** (`data.py`) — every f-string interpolates **`etype.value`**
  (`needextend.py:111, 115, 122, 152, 162, 173`). Everything else is `==`
  (`needextend.py:119, 147`) or `match`/`case` patterns
  (`needextend.py:264, 278, 298, 301, 311, 334, 355, 359, 363`); the rest are assignments,
  tuple construction and type annotations.
- **`MessageRuleEnum`** (`schema/config.py`) — `warning["rule"].name` for the debug-file name
  and the ignore list (`schema/reporting.py:83, 90`), `warning["rule"].value` for the warning
  subtype (`schema/reporting.py:220, 308, 360`), dict keys in `MAP_RULE_DEFAULT_SEVERITY`
  (`schema/config.py:481–499`) and its lookup (`schema/core.py:604`), and `!=`
  (`schema/core.py:615`).

`.name`, `.value`, `==`/`!=`, dict hashing and `match` are identical under both shapes. Two
further checks: nothing serialises the enum (`save_debug_file` writes `warning["need"]`, never
the warning dict that holds `rule`), and no test or document names either class. `ExtendType`
rides into the doctrees inside `NeedsExtendType`, but the qualname is unchanged so existing
`.doctrees` unpickle. `SeverityEnum` is an `IntEnum` and is untouched.

So the change is applied rather than ignored — no `extend-ignore`, no `# noqa`, lint green on
the merits.

## `typing-extensions` stays

It looks like it should go with `NotRequired`, and it does not: ruff treats **`Unpack` as a
3.12 member** (PEP 692's `**kwargs: Unpack[TypedDict]` semantics), so
`directives/needflow/_directive.py` still imports it from `typing_extensions` and the
dependency is still real. Its trailing comment now says so —
`# for \`Unpack\`, which typing gained in python 3.12` — instead of naming `NotRequired`,
which has moved to `typing`. Revisit at a 3.12 floor.

## The pin, and its two fences

The second commit adds a root **`.python-version`** containing `3.13`, so a bare `uv sync` is
the same environment on every machine and in every agent session. It matters here because the
default environment's *sphinx* follows the interpreter: below 3.12 the lock resolves sphinx
8.2, so on such a machine `test-needs` was quietly a second sphinx-8 run and the Lint job's
"newest sphinx" canary failed locally but not in CI. 3.13 because it is the matrix's main
cell and the lowest interpreter that resolves the newest sphinx. uv downloads it if the
machine lacks it, and pyenv, asdf and mise read the same file (`pyenv install 3.13` once).

`UV_PYTHON` **overrides** the file — that is how the per-cell commands and every matrix cell
pick their interpreter (`setup-uv`'s `python-version` input is documented as setting
`UV_PYTHON`). The Lint job is the only job that passes no such input, which is what makes it
the job that tests the pin. Two assertions there:

1. **the canary grew a second half.** It still asserts `sphinx.version_info[0] >= 9`, and now
   also that the interpreter series equals `.python-version` — reading the expected value out
   of the file (`v=$(cut -d. -f1,2 .python-version)`) rather than writing the number down
   twice. Proven red by running the step's own shell under `UV_PYTHON=3.12`:
   `AssertionError: the default environment is on python 3.12, but .python-version says 3.13`.
2. **a new floor-equality fence**: the root's `requires-python` must equal every
   `packages/*/pyproject.toml`'s. Nothing enforced that before, though the root's comment
   promised it — and uv resolves one lock against the *root*, so a member that floored itself
   higher would still lock, build and publish, and the gap would only surface as an install
   failure for a user on the interpreter inside it. Proven red by flooring the member at
   `>=3.12,<4`:
   `::error file=packages/sphinx-needs/pyproject.toml::requires-python is >=3.12,<4, but the root declares >=3.11,<4`.

## The lock churn, by class

`102` package entries where there were `108`; **no package resolves to a different version**,
every delta is a 3.10-only leg disappearing. Verified by parsing both lock files and diffing
the name → versions map: *versions added anywhere: none; names added: none*.

| class | packages |
|---|---|
| dropped entirely (existed only under `python_full_version < '3.11'`) | `exceptiongroup` |
| dual/triple resolution collapsing to the surviving leg | `contourpy` (1.3.2+1.3.3 → 1.3.3) · `matplotlib` (3.10.9+3.11.1 → 3.11.1) · `numpy` (2.2.6+2.4.6+2.5.2 → 2.4.6+2.5.2) · `sphinx-design` (0.6.1+0.7.0 → 0.7.0) |
| same, one step removed | `sphinx` (7.4.7+8.1.3+8.2.3+9.1.0 → 7.4.7+8.2.3+9.1.0). 8.1.3 existed only as the 3.10 fallback: the `sphinx-8` group's old marker left it empty on 3.10, so the environment fell back to `sphinx>=7.4,<10` and resolved 8.1.3, the last sphinx that accepted 3.10 |
| marker pruning | 345 occurrences of `python_full_version < '3.11'` → 0; `sphinx-syntax-example` loses its marker |

`uv lock --check` is clean, and the `uv-lock` prek hook produces the same file.

## Also in this PR

- **The minimum Python is stated in the user documentation for the first time.** Neither
  `README.rst` nor `docs/installation.rst` said one; `installation.rst` now opens *Using pip*
  with "**Sphinx-Needs** requires Python 3.11 or newer."
- Four dead `contributing.rst` note lines and the same dead rationale comment
  (`# the docs extra requires python>=3.11 (sphinx-syntax-example)`) in five workflows, all of
  which existed only to describe the 3.10 gap.
- `[tool.ty.environment] python-version = "3.11"`, so the type gate targets the new floor; and
  the now-dead `tomli` entries in `[tool.ty.analysis] allowed-unresolved-imports`.

## No changelog entry here — on purpose

The release this ships in (8.6.0 or 9.0.0) is a release-time decision, so this PR does not
touch `docs/changelog.rst`. The bullet for whichever release PR picks it up:

```rst
Breaking changes
................

- ‼️ Python 3.10 is no longer supported (:pr:`1844`)

  Python 3.10 reaches `end of life <https://devguide.python.org/versions/>`__ in October 2026,
  and the minimum supported version is now **Python 3.11**.

  A project already on Python 3.11 or later needs to change nothing. A project still on 3.10
  is not broken either: ``pip`` reads the ``Requires-Python`` metadata, so
  ``pip install -U sphinx-needs`` there stays on the last release that supported it (8.5.0)
  rather than failing to install.

  Two smaller consequences for the installed package:

  - The ``tomli`` dependency is gone. :ref:`needs_from_toml` now reads the standard library's
    ``tomllib``, which Python 3.11 added.
  - The ``docs`` extra no longer carries a version marker on ``sphinx-syntax-example``, so it
    installs in full wherever it resolves.

  Sphinx 9.1 requires Python 3.12, so the ``sphinx~=9.1`` combination is tested from 3.12 up;
  Python 3.11 is tested against Sphinx 7.4 and 8.2.
```

*(8.4.0's section order is `Improvements` → `Deprecations` → `Breaking changes` →
`Internal changes` → `Bug fixes`, so this block goes above the `Unreleased` section's existing
`Bug fixes`.)*

## Review

An adversarial review re-measured every claim above by its own hand — the `StrEnum`
conversion down to a byte-identical `needs.json`, the lock (no version moved), the exclude
both ways, both fences red and green, the wheel installed out of the project on 3.11, the
Docker base image's Python, and the three full suites — and all of it held. What it found
was prose: the claim that Lint is the *only* job reading the pin (the monthly `prek-update`
job omits the input too), the asdf/mise clause (both read the file only when their
idiomatic-version-file setting is on), a missing pyenv note in `contributing.rst`, the
`Unpack` comment naming the wrong Python version, and the five `python-version: "3.12"`
pins that had lost their rationale when the old comment was deleted. The third commit is
those corrections, plus `cache-python: true` on the two jobs that now download the pinned
interpreter.


ubmarco (approved):
Would be interesting to know what Sphinx 9.1 uses from Python 3.12, that does not exist in Python 3.11. The [changelog entry](https://www.sphinx-doc.org/en/master/changes/index.html#release-9-1-0-released-dec-31-2025) is pretty silent on this.

chrisjsewell:
Good question, and the answer from Sphinx's own tree is: nothing in particular. 9.1 dropped 3.11 **by policy, not for a feature**.

- The drop is sphinx-doc/sphinx#14153 ("Drop support for Python 3.11", 2025-12-05), whose whole description is "Per our [support policy](https://www.sphinx-doc.org/en/master/internals/release-process.html#python-version-support-policy)". That policy is SPEC 0: support the minor versions released in the past three years, minimum three. Python 3.11 was released in October 2022, so a Sphinx released after October 2025 is allowed to drop it, and 9.1 (December 2025) did. The `CHANGES` entry is the one line "#14153: Drop Python 3.11 support".
- What the commit actually changed, apart from `requires-python` and the classifier, is the removal of about a dozen `sys.version_info` branches — nearly all in autodoc's signature/typing handling and its tests — plus the matching `uv.lock` pruning: 51 files, +154/−377.
- Grepping the 9.1.0 tree for 3.12-only constructs finds no `type` statements, no PEP 695 generics, no `itertools.batched`, no `Path.walk`, no `typing.override`. The one 3.12 API is `typing.TypeAliasType` in `sphinx/util/typing.py`: 9.0.4 imports it under `if sys.version_info[:2] >= (3, 12)` with a `typing_extensions` fallback, 9.1.0 imports it unconditionally because the guard was deleted. So on 3.11 a 9.1 install would fail at that import — a consequence of the cleanup rather than the reason for it.

For us that means the `3.11 × sphinx-9` cell is excluded because 9.1's metadata says so, and the lock resolves 9.0.4 on 3.11 (the last 9.x that accepts it), which the out-of-project smoke install on 3.11 exercises lightly. Nothing to do about it in this repository; it just needs to be understood when reading the matrix.

ubmarco:
Thanks for the research effort. The policy is an interesting take. Certainly good for maintenance, bad for consumers that can't upgrade, which is pretty common in bigger companies...

PR #1845, opened 2026-09-04
🧪 Replace Cypress with pytest-playwright for the JavaScript tests

### LINKED ISSUE #1843: 🧪 Replace Cypress with pytest-playwright for the JavaScript tests
### Motivation

The two JavaScript tests (`tests/test_sn_collapse_button.py`, marker `jstest`) are the one
part of the suite that a contributor or an agent cannot run from `uv sync` alone. They run
through Cypress, and Cypress costs far more than what it verifies:

- **What is verified** is small: both specs (`tests/js_test/sn-collapse-button.cy.js` for
  the pytest path, `tests/js_test/cypress/e2e/test-collapse-button.cy.js` for the docs-site
  path) assert only that, on load and after a click, the right table rows and the two icon
  spans carry the `collapse_is_hidden` class. The script under test,
  `src/sphinx_needs/libs/html/sphinx_needs_collapse.js`, is ~50 lines of jQuery class
  toggling with no layout or timing dependence.
- **What it takes to run**: Node, an `npm install cypress` inside the package (the
  `package.json` exists only for this), a ~250 MB Electron binary fetched by the postinstall
  from `download.cypress.io` via `cdn.cypress.io` into `~/.cache/Cypress/<version>/`, a
  subprocess bridge in `tests/conftest.py` (`test_js()` spawns `npx cypress run` with a
  JSON config written at test time), a `pytest-xprocess` HTTP server fixture on a fixed port
  serving the build directory, and a second, separate harness in `.github/workflows/js_test.yml`
  that rebuilds the whole docs site and drives Cypress through `cypress-io/github-action`
  with `npm start` (`python3 -m http.server`).
- **Where it fails**: in a sandboxed agent session the binary download is truncated by the
  proxy and the installer hides it as `Corrupted download` (diagnosed on #1839; the
  workaround, seeding the cache with `curl -C -` from the environment setup script, is
  recorded in `CLAUDE.md`). Locally it is the one step of `docs/contributing.rst` that is
  neither `uv`-managed nor cached by it.

### Proposal: pytest-playwright

Rewrite the two specs as Python tests using [pytest-playwright](https://playwright.dev/python/docs/intro):

- The tests use the existing `test_app` fixture to build the project, then open the built
  page with the `page` fixture over `file://` (nothing in the page fetches, so no server is
  needed) and assert on `page.locator(...)` classes with the same selectors as today. About
  forty lines of Python replace both `.cy.js` files.
- Dependencies: a `js` (or `browser`) dependency group with `pytest-playwright`, resolved
  from the lock like everything else. The browser is a one-time
  `uv run playwright install chromium` (or `chromium-headless-shell`, smaller) into
  `~/.cache/ms-playwright`, or `channel="chrome"` to drive a browser already on the
  machine with no download at all.
- Retired: `package.json`, `tests/js_test/` (Cypress config, support file, both specs), the
  `test_js()` bridge and the `test_server` fixture in `conftest.py`, the `npm install
  cypress` steps in `ci.yaml`'s JS matrix, and `js_test.yml` as a separate harness: its
  docs-site checks (`tutorial.html`, `directives/need.html`) become a third parametrisation
  of the same test against the built docs, or are dropped if the two project-level cases
  are judged sufficient.
- `poe test-needs-js` stays as the entry point; the `jstest` marker stays.

### Benefits

1. One harness, one language, one lock: the JS tests become ordinary pytest tests that
   `uv sync --group js` and one `playwright install` make runnable, locally and in any
   agent session, with the browser cached per machine like every other artefact.
2. No Node toolchain inside a Python package, no subprocess bridge, no fixed-port server,
   no second CI workflow duplicating the first.
3. A smaller and resumable browser install (Playwright's downloader retries; the
   headless shell is a fraction of Cypress's Electron bundle), and the option of no
   download at all where a Chrome exists.
4. Room to grow: real-browser assertions on the DataTables integration, the permalink
   script or the `needs.json` viewer become cheap once the fixture exists, which Cypress's
   cost has discouraged.

### Costs and open questions

- CI needs `playwright install --with-deps chromium` in the JS cells (apt dependencies on
  the runner); a sandboxed environment needs Playwright's download host allowlisted, as
  Cypress's are today.
- `docs/contributing.rst`'s "Running JS Testcases with PyTest" section is rewritten.
- Keep the docs-site end-to-end case (today only in `js_test.yml`) or not?
- Chromium only, or also WebKit/Firefox? Today's coverage is Electron's Chromium only, so
  Chromium alone loses nothing.

The alternative considered was jsdom under vitest (no browser download at all), rejected
because it keeps a Node toolchain in a Python package and can never grow into real
rendering checks.

Scope: one contained change, a good candidate for a single builder with one reviewer, after
the workspace's Python floor and release work.

chrisjsewell:
## Why

The two `jstest` tests were the one part of the suite that `uv sync` alone could not run.
They went through Cypress, and Cypress cost far more than what they verify: Node, an
`npm install cypress` inside the package (the `package.json` existed only for that), a
~250 MB Electron binary from `download.cypress.io`, a `test_js()` subprocess bridge in
`conftest.py` that wrote a JSON config at test time, a `pytest-xprocess` HTTP server on a
fixed port, and a *second* harness in `.github/workflows/js_test.yml` that rebuilt the
whole documentation site to run the same assertions again. What all of that verified is
~50 lines of jQuery class toggling in `sphinx_needs_collapse.js`, with no layout or timing
dependence. Closes #1843.

## What the tests do now

`tests/test_sn_collapse_button.py` builds the same two projects with the same `test_app`
fixture — plus a third, see below — then opens `index.html` over `file://` with
pytest-playwright's `page` fixture —
nothing on the page fetches, so there is no server and no port. For every
`table.need span.needs.needs_collapse` control it takes the mode and the row classes out
of the control's id, scopes the row lookups to the enclosing `div[id^=SNCB-]`, and asserts
`collapse_is_hidden` on the two icons and the named rows both on load and after a click —
the same assertions the two `.cy.js` specs made, step for step. Class checks use
`expect(...).to_have_class()` with a word-boundary regex rather than equality on the whole
`class` attribute, and they now also fail if a selector matches nothing, which Cypress's
`.each()` silently passed.

### A branch neither harness was running

While writing this I measured that every collapse control on `doc_test/doc_basic` and
`doc_test/variant_doc` — the only two projects the Cypress specs used — is a `show`
control. The `hide` branch of `sphinx_needs_collapse.js` was therefore never executed, by
the old specs or the new tests. There is now a third parametrisation on
`doc_test/import_doc`, whose index imports needs with `:collapse: True` and so carries 13
`hide` controls next to 31 `show` ones, and that case asserts both modes were met so the
coverage cannot lapse silently.

It is real coverage, not just execution: sphinx renders the initial *icon* state into the
HTML, so an icon assertion passes even with the script removed — but it does not pre-hide
the metadata rows, and those are the load-time `hide` branch's own work. Dropping
`classList.add()` from that branch's row loop fails the new case and leaves the other two
green.

`page.on("pageerror")` replaces the support file's blanket allowance for any uncaught
exception whose message contained `"is not defined"`. Measured: **nothing** on these pages
throws — not the two project pages, and not `tutorial.html` or `directives/need.html` on a
real docs build, over `file://` or over HTTP. The allowance was dead code, so the
assertion here is unconditional.

## What is gone

`tests/js_test/` (both specs, the Cypress config, the support file), `package.json`, the
`test_js()` bridge, the `test_server` fixture and the `spec_pattern` key in `conftest.py`,
`pytest-xprocess` from the `test` group, the three Cypress/Node lines in `.gitignore`, and
the Node steps in `ci.yaml`'s `tests-js` job.

`.github/workflows/js_test.yml` goes too. It was a second harness for the same script:
it rebuilt the docs site on every pull request to run the same assertions against
`tutorial.html` and `directives/need.html`. Those pages exercise no branch of
`sphinx_needs_collapse.js` that `doc_test/variant_doc` and `doc_test/doc_basic` do not, so
the docs build per PR bought nothing. If a docs-site case is wanted later it is a third
parametrisation of the same test, not a second workflow.

## Running them

```bash
uv sync                    # pytest-playwright comes with `dev`
uv run poe install-browser # once per machine: playwright install chromium
uv run poe test-needs-js
```

One cost worth stating so it can be reversed knowingly: `dev` includes the `js` group, so
every default `uv sync` — contributors, agents, and CI's Lint job — installs the
playwright wheel, which is **146 MB** (142 MB of it the bundled Node driver). The
alternative is giving `test-needs-js` its own `executor = { group = ["test", "js"] }`
environment the way the sphinx matrix cells have theirs; one deterministic environment was
judged the better default for two tests.

The browser is not a package, so it gets its own repo-wide task. Measured here: 274 MiB
downloaded, 554 MB on disk, 27 s, into `~/.cache/ms-playwright`
(`~/Library/Caches/ms-playwright` on macOS), from `cdn.playwright.dev` with
`playwright.download.prss.microsoft.com` as the driver's fallback mirror — a sandboxed
environment needs those two allowed where it used to need Cypress's two. To download
nothing at all, `uv run poe test-needs-js --browser-channel chrome` drives a Chrome that is
already installed. Chromium only: the old coverage was Electron's Chromium, so nothing is
lost.

In CI the `tests-js` cells sync `--group js` and run
`playwright install --with-deps chromium` before pytest. That step is deliberately
uncached to start with; if it turns out to be slow, `actions/cache` on
`~/.cache/ms-playwright` keyed on the lock is the fix, and there is a comment in the
workflow saying so.

## Changelog

Internal changes:

* ♻️ Replace Cypress with pytest-playwright for the JavaScript tests (:pr:`XXXX`)

**Notes for the author of the PR:** it is stacked on #1844 (`ci/python-3.11-floor`) —
both touch `ci.yaml`'s JS cells and the lock — and needs re-parenting onto `master` after
that squash-merges. No security checklist.

## Review

An adversarial review re-measured the branch by its own hand: the spec-to-Python mapping
line by line, three `hide`-branch mutations red on the new case alone, the full suite at
1740 passed / 12 skipped / 3 deselected, hooks, smoke, lock and branch shape, and the
`--with-deps` / `chromium` / download-host claims against the driver's own registry source.
One claim fell: "the old `is not defined` allowance was dead code" is true for the three test
projects but not for every theme — a docs build with `sphinx_immaterial` raises
`DOCUMENTATION_OPTIONS is not defined` from sphinx-copybutton on every page, because that
theme emits no `documentation_options.js`. That is not sphinx-needs' script and nothing
shipped is affected, but the page-error assertion's comment now says exactly which pages it
is justified for, so whoever adds a docs-site case knows it needs a scoped allowance. The
fifth commit is that wording, the `contributing.rst` example made strict-mode safe
(`to_have_count(2)` instead of a multi-match visibility check), the measured runtimes (3 s
warm, 9 s cold) and the macOS cache path in the task help. Recorded without action: a
control inside a closed `<details>` would stall the helper, the multi-row control is covered
by neither harness, and `--no-group js` does not drop the 146 MB because `dev` includes `js`.


ubmarco (approved):
All-in on this one

chrisjsewell:
**Cloud-session trial: the browser tests work in a Claude Code web container, given one allowlisted host.**

Two Claude Code cloud sessions ran this branch on the Ubuntu 24.04 image.

- The first was refused the browser download: the session proxy answers `403 no rule or allowlist entry allows host "cdn.playwright.dev"` (four retries, ~5 s) until the environment's network allowlist names that host. It also found that the image sets `PLAYWRIGHT_BROWSERS_PATH=/opt/pw-browsers` with a Chromium at another revision baked in, which playwright 1.62 ignores (`Executable doesn't exist at …`).
- With `cdn.playwright.dev` added to the environment, the second session's `uv sync --frozen && uv run poe install-browser && uv run poe test-needs-js` gave **3 passed in 9 s** (299 MiB fetched, 651 MB on disk, 14 s).

Their edits to `AGENTS.md`, `CLAUDE.md` and `contributing.rst` are folded into 5eb22b3c, re-verified against the installed driver: on linux-x64, macOS and Windows chromium comes from `cdn.playwright.dev` alone (Chrome-for-Testing builds, `cftUrl`); the mirror list with `playwright.download.prss.microsoft.com` is consulted only for linux-arm64, so the earlier "two mirrors" wording was over-inclusive. `.claude/settings.json` now sets `UV_HTTP_TIMEOUT=180` for every Claude Code session, cloud ones included, because uv's 30 s default timed out on a cold `uv sync` through the session proxy. Dropped: their recipe for symlinking the image's Chromium into the pinned revision's paths; with the host allowed there is no reason to teach an agent to fake a browser install.

The network allowlist itself cannot be declared from the repository (nothing under `.claude/` feeds it); the hosts an environment for this repository needs are listed in `CLAUDE.md`.

Issue #540, opened 2022-05-10
Remove hardcoding of NEEDS_VERSION in Docker Release Pipeline


twodrops:
The docker release pipeline is triggered based on git tags and the tag is [already available in the pipeline script](https://github.com/useblocks/sphinxcontrib-needs/blob/791647fc5daa8d12f96acb360f67b94dd772bc62/.github/workflows/docker.yaml#L73). Hence the following Version hardcoding can be avoided.

https://github.com/useblocks/sphinxcontrib-needs/blob/791647fc5daa8d12f96acb360f67b94dd772bc62/.github/workflows/docker.yaml#L16

PR #1846, opened 2026-09-04
🔧 One tag-driven release workflow for every workspace member, on trusted publishing

### LINKED ISSUE #1758: Switch PyPI publishing to Trusted Publishing (OIDC)
## Summary
The release workflow (`.github/workflows/release.yaml`) currently publishes to PyPI using a long-lived API token stored as the `PYPI` repository secret:

```yaml
env:
  FLIT_USERNAME: __token__
  FLIT_PASSWORD: ${{ secrets.PYPI }}
```

Switch to PyPI **Trusted Publishing** (OIDC), eliminating the manually-managed, long-lived secret.

## Motivation
- No long-lived API token to create, rotate, or accidentally leak.
- Tokens minted via OIDC are short-lived (15 min) and auto-expire.
- Enables Sigstore digital attestations on uploads (bonus, requires no extra setup).
- Aligned with PyPA best practice.

## Steps
1. **Configure the Trusted Publisher on PyPI** for `sphinx-needs`:
   - https://pypi.org/manage/project/sphinx-needs/settings/publishing/
   - Owner: `useblocks`, Repository: `sphinx-needs`, Workflow: `release.yaml`
2. **Modify `.github/workflows/release.yaml`**:
   - Add `permissions: id-token: write` to the `publish` job.
   - Build with `python -m build` (via `pip install build`) instead of `flit publish`.
   - Replace the publish step with `pypa/gh-action-pypi-publish@release/v1` (no `username`/`password`).
   - Remove the `FLIT_USERNAME` / `FLIT_PASSWORD` / `secrets.PYPI` env block.
3. **Delete the `PYPI` secret** from the repository once the OIDC publish is verified.
4. Verify a tag-triggered run publishes to PyPI successfully.

## References
- https://docs.pypi.org/trusted-publishers/
- https://github.com/pypa/gh-action-pypi-publish

### LINKED ISSUE #1759: Update to flit 4 / flit_core 4
## Summary
Update the build backend and tooling from flit 3 to flit 4.

Current state:
- `pyproject.toml` → `[build-system] requires = ["flit_core >=3.4,<4"]`
- `.github/workflows/release.yaml` → `pip install flit~=3.4`

## Relevant flit 4.0 changes (https://flit.pypa.io/en/stable/history.html)
1. **`[tool.flit.metadata]` table removed** — this project already uses the PEP 621 `[project]` table, so it is NOT affected. The constraint can simply change to `<5`.
2. **`flit build` / `flit publish` no longer infer sdist contents from VCS** (git/hg) — sdist inclusion must now be declared via `[tool.flit]` sdist `include`/`exclude` patterns, or pass `--use-vcs` for the old behaviour. This project has no `[tool.flit]` section today, so the sdist contents may change — **must verify** after upgrading.
3. Requires Python ≥ 3.8 — already satisfied (project requires ≥ 3.10).

## Steps
1. `pyproject.toml`: change `[build-system].requires` to `"flit_core ~=4.0"` (or `"flit_core >=4,<5"`).
2. `.github/workflows/release.yaml`: `pip install flit~=3.4` → `pip install flit~=4.0`.
3. Add `[tool.flit]` sdist `include`/`exclude` patterns if the sdist contents differ (compare `flit build` output before/after).
4. Update any docs/changelog references to flit, if present.
5. Verify the release workflow builds and the sdist/wheel contents are unchanged.

chrisjsewell:
Closes #1758. Closes #1759. Refs #540.

Nothing is released by this pull request. The first release through it is a later, deliberate act: merge, then push a tag.

**The workflow.** `.github/workflows/release.yaml` keeps its name — all four PyPI projects (sphinx-needs, sphinx-mounts, sphinx-codelinks, sphinx-test-reports) name `release.yaml` in their trusted-publisher entries — and becomes one workflow for every distribution under `packages/`, with the tag saying which: `<dist>-v<version>`. Four jobs: `plan` validates the tag against the tree, `build` builds and gates, `publish` uploads with `uv publish --trusted-publishing always` under `environment: pypi-<dist>` and `id-token: write`, and `release` creates the GitHub Release. There is no `secrets.PYPI` anywhere any more. It does **not** deliver #1758's attestation bonus: uv 0.12.9 does not generate attestations ("`uv publish` does not currently generate attestations", its own docs), and `--no-attestations` only suppresses *uploading* sidecar files that nothing here writes — see the follow-up candidate below. Publish runs before the Release step on purpose: a publish that succeeded with a failed release step is fixable by hand, and the reverse is not. `plan` is a separate job rather than a step inside `publish` because GitHub **creates** an environment a workflow names and that does not exist, with no protection rules at all — so `environment: pypi-${{ needs.plan.outputs.dist }}` is a fence only for the four names that already have one, and a failed `plan` is what stops the expression from ever being evaluated.

**The guardrails.** A monorepo removes an accidental safety net the four-repo era had: each extension's CI installed sphinx-needs *from PyPI*, so a broken version claim was caught by the next build. In a workspace, uv resolves the member on disk whatever the specifier says (astral-sh/uv#9811) — and the specifier is not in `uv.lock` at all, so changing it leaves `uv lock --check` at exit 0 with a zero-line lock diff. `--frozen`, the `uv-lock` hook and any amount of reading the lock are structurally incapable of seeing it. Six guards replace that net; each has a red proof in this pull request.

| | guard | when it fires | what it catches | red proof here |
|---|---|---|---|---|
| a | `tools/src/sn_tools/check_workspace.py` | Lint, every PR (+ `build`) | five manifest facts: the root lists every member bare; every member is sourced from the workspace; every member's `requires-python` equals the root's; every intra-workspace specifier admits the tree's version and is tightly tracked; `__version__` equals `[project] version` | 5 mutations, each exit 1 with the file named |
| b | `uv pip install --no-sources --dry-run <wheel>` | `build`, before publish | the whole transitive closure unsatisfiable from the index — intra-workspace *and* external pins | a scratch member on `sphinx-needs>=99`: "only sphinx-needs<=8.5.0 is available", exit 1 |
| c | `tools/src/sn_tools/release_plan.py` | `plan`, first job | a tag naming no member; a version the tree does not build; a version already on PyPI; a dependency not yet released; a tagged commit off the default branch | unit-test reds for every check, incl. every non-404 PyPI answer |
| d | `tools/src/sn_tools/propagate_floors.py` | the release PR | floors left behind by a bump (it is the fix; (a) is the check) | the diamond scenario, `--check` exit 1 → 0 |
| e | the compat cell | `build` | **API drift against the released dependency** — the only content-aware guard | the rehearsal runs below; it already caught a real bug in itself |
| f | `plan` as a `needs:` gate | every run | a typo'd tag auto-creating an unprotected environment | (a)–(c) above |

(d) is vacuous with one member. It lands now because it is the half of the pair nobody writes later under release pressure, and because the day it stops being vacuous is a release day.

**Static version (#1758's prerequisite).** `packages/sphinx-needs/pyproject.toml` declares `version = "8.5.0"`: `uv version --package` refuses a dynamic version outright, and the workflow's tag/version agreement check cannot run the build backend to find one. `description` stays dynamic. `__version__` stays a literal in `__init__.py` — it is stamped into every generated `needs.json`, a documented interchange format, so an `importlib.metadata` lookup would put a `PackageNotFoundError` failure mode in the one module every user's `conf.py` reaches — and check (5) of the fence is what keeps the two numbers equal. `uv.lock` gains the one `version =` line it was missing.

**The sdist (#1759).** flit 3 inferred sdist contents from git (`flit publish --use-vcs`, what this workflow ran); flit 4 flipped that default off, and `uv build` — which resolves `flit_core >=3.4,<5` to 4.0.2 today — never had it. Measured on this tree: with no `[tool.flit.sdist]` table the sdist is **2.5 MB / 707 entries** against **28 MB / 1380** with one. An 11× shrink, no warning, exit 0, visible only on PyPI afterwards. The table declares `tests/`, `docs/` and `performance/`, with an `exclude` list measured against a checkout in which the docs and part of the suite had actually been run (`docs/_build/` alone is 456 files), and `poe smoke-needs` now asserts on every run that those trees are in the tarball and that none of that rubbish is. The downstream case was checked rather than assumed: **conda-forge builds from this sdist** (its test step is an import check), and **Debian runs the suite** — out of the GitHub tag tarball its `debian/watch` fetches, not out of the PyPI sdist. Fedora unverified. That closes #1759 in the way it asked to be closed — the contents are declared, so they no longer depend on which flit_core is resolved, and nothing installs the flit CLI any more.

**Docker (#540).** `docker.yaml` triggered on `'*.*.*'`, and GitHub's tag globs match hyphens, so it would have fired on *every* member's prefixed tag and handed `docker/metadata-action` a string that is not PEP 440. It now keys on `sphinx-needs-v[0-9]+.[0-9]+.[0-9]+` — the image is sphinx-needs' — takes the image version from the tag (`type=match`), and takes `NEEDS_VERSION` from the tag too: the Dockerfile uses it as a git ref, and the prefixed tag *is* the ref. Refs rather than closes, because the literal is still hand-bumped for the runs that have no tag (the `docker/**` `pull_request` filter and a manual dispatch).

### Three one-way doors, for @chrisjsewell

Each is reversible with one edit **before the first tag**, and not afterwards.

1. **The bare companion tag for sphinx-needs.** After publishing, the `release` job pushes a second, bare `<version>` tag. Read the Docs derives `stable` by running `packaging.version.Version()` over each tag name and **dropping** what it cannot parse — `sphinx-needs-v8.6.0` is not PEP 440, so without this `/en/stable/` freezes at 8.5.0 for ever, and there is no RTD-side setting that can point `stable` by hand. The bare tag also keeps every `git+…@<version>` pin and every existing inbound URL working. **Debian is a second, independent reason**: `debian/watch` scans this repository's tags page with `(?:.*?/)v?@ANY_VERSION@@ARCHIVE_EXT@`, which matches `8.6.0.tar.gz` and does *not* match `sphinx-needs-v8.6.0.tar.gz` — the filename starts with `s`, not with `v` or a digit — so without the bare tag `uscan` stops finding releases. The price is two refs per sphinx-needs release, and that the bare namespace can then only ever belong to sphinx-needs. *To reverse:* delete the last step of the `release` job in `release.yaml` (it is fenced by its own comment block) and set up an RTD automation rule on `^sphinx-needs-v` instead.
2. **The sdist keeps `tests/` and `docs/`.** 28 MB rather than 2.5 MB. This is what has always shipped, conda-forge builds from it, and stopping should be a decision, not a side effect of moving to OIDC. *To reverse:* delete the `[tool.flit.sdist]` table from `packages/sphinx-needs/pyproject.toml` and pass `--sdist-dirs` with no arguments in the `smoke-needs` task. Write the reason in the manifest if you do.
3. **Docker is keyed to sphinx-needs tags only.** A future member that wants an image needs its own workflow (or another entry in this one's trigger). *To reverse:* widen the tag filter in `docker.yaml`, and make the metadata pattern match the wider set.

### Rehearsals

`release.yaml` was dispatched from this branch (`workflow_dispatch` with `tag=sphinx-needs-v8.5.0`) and ran `plan` and `build` end to end: the tag parsed, the "already on PyPI" check printed its rehearsal notice, the ancestry check ran against the `origin/master` the deep checkout fetched and reported "HEAD is NOT an ancestor" as a notice, the manifest fence passed, the wheel built, the resolver gate resolved it from PyPI alone, and the compat cell ran the suite against the released dependencies (1740 passed on CPython 3.13 / sphinx 9.1.0, about six minutes). `publish` and `release` were skipped (`if: github.event_name == 'push'`). Runs: [33923058526](https://github.com/useblocks/sphinx-needs/actions/runs/33923058526) (red, and worth having: 23 tests shell out to `sphinx-build`, which the compat environment's interpreter alone does not put on `PATH` — locally the cell had passed by finding another checkout's `sphinx-build`), [33924379418](https://github.com/useblocks/sphinx-needs/actions/runs/33924379418) and [33927781081](https://github.com/useblocks/sphinx-needs/actions/runs/33927781081) (green, the latter on the reviewed tree).

### After merging

1. **Rehearse once from `master`**: `gh workflow run release.yaml -f tag=sphinx-needs-v8.5.0`. Expected: `plan` green with an "already on PyPI" notice and an ancestry notice, `build` green, `publish` and `release` skipped.
2. **The first release** is yours, and the version is decided then. Following the recipe now in `AGENTS.md`: bump, changelog, merge, `git push origin sphinx-needs-v<version>`. A patch release is the cheapest first one.
3. **Then delete the `PYPI` repository secret** (#1758 step 3 — @chrisjsewell / @ubmarco), once one OIDC release has been verified. Nothing reads it after this PR merges.
4. Read the Docs keeps activating the bare tags exactly as it does today; nothing changes there.
5. **Follow-up candidate, not in this PR:** attestations need a generating step before `uv publish` (e.g. `pypi-attestations`, which is what `pypa/gh-action-pypi-publish` runs). That is a separate decision, not a wording change.

**Residual risk, stated plainly:** the `publish` job is the one job that cannot be rehearsed — a dispatch never reaches it, and rehearsing the real OIDC handshake would need a TestPyPI publisher registered separately. So its first real run is the first real tag. If it goes red, **nothing has been published** (it is the first job after the build, and the GitHub Release comes after it), and the fix is to re-run the workflow from the same tag once the cause is addressed — the tag does not need to be moved or deleted. The *other* case is the one to know the button for: if `publish` succeeded and `release` failed, use **Re-run failed jobs** (`gh run rerun <id> --failed`), which keeps `plan`'s outputs and `build`'s artifact and re-runs `release` alone. *Re-run all jobs* would fail at `plan`, because by then the version is on PyPI.

### Where the tooling lives

The three fences and their tests live in `tools/`, a **virtual workspace member** (`[tool.uv] package = false`) outside `packages/`, with its own manifest declaring `packaging` and `tomlkit` once; the root depends on it, so those reach every environment, the typing floor included, and neither sits in a dependency group. Three scripts stay in `.github/scripts/`, and `AGENTS.md` states the rule: what decides is whether a check must execute *inside* a particular environment (`check_sphinx_cell.py` imports the cell's own sphinx, `check_typing_floor.py` runs inside `.venvs/typing`, `extract_benchmark_data.py` runs inside the benchmark job).

What `package = false` does, measured on uv 0.12.9: `uv build --all-packages` skips the member, `uv build --package sphinx-needs-workspace-tools` is refused ("is missing a `build-system`"), the lock records `source = { virtual = "tools" }`, and `uv sync` installs its dependencies without installing it. What it does **not** do: prohibit a build. `uv build tools/` falls through to PEP 517's default backend and produces a real sdist and wheel. Nothing in the workflows can reach that form, since the `build` job only ever runs `uv build --package "$DIST"` after `plan` has validated the tag, and two fences of our own make the member unreleasable: the plan job refuses a tag naming a virtual member (and a publishable member depending on one), and the manifest carries `Private :: Do Not Upload`, which PyPI rejects on upload and which the workspace fence asserts is present. Each has a red test.

The cost, measured: every environment, matrix cells included, gains exactly one package, `tomlkit`, at about a quarter of a megabyte with no dependencies. Tests sit beside the code in `tools/tests/`, whose `conftest.py` puts `tools/src` on `sys.path` (never `tools/`, so no second `tests` package can compete with the sphinx-needs suite's own).

### Review

Built from a measured recon (uv#9811 re-verified on 0.12.9; the lock does not record intra-workspace specifiers at all; the sdist shrink), then two adversarial review loops. The first, on the pipeline: thirteen named claims, one fix round and a validation round; it refuted the recon's reading of `--no-attestations` (hence the explicit negative above), found two manifest shapes that produced tracebacks instead of annotations, a source-key lookup that did not canonicalise names, stale file counts and an unmeasured downstream claim, and confirmed by hand that `GITHUB_SHA` is the tagged commit even for annotated tags, that the compat cell runs on the pinned interpreter on the runner, that the sdist is exact on a dirty tree, and that `gh run rerun --failed` preserves the plan job's outputs. The second, on the tooling member: it refuted the claim that uv could never build a virtual member (hence the paragraph above and the classifier fence), and confirmed the plan-job rule, the dependency contract, the collection counts and the history following the moved files. Every finding is closed on the tip; the residue of each loop was comments.

### Draft changelog bullet

- 🔧 Releases are now published with PyPI trusted publishing (OIDC) from a single tag-driven workflow, and the source distribution's contents are declared rather than inferred from git (:pr:`1846`)


ubmarco (approved):
Quite some logic in the release process, and a bit too much scripting in CI yaml files - hard to execute locally (if we ever want that...).

PR #1847, opened 2026-09-05
🔧 House the workspace tooling in a virtual member that is never released

chrisjsewell:
**Stacked on #1846**, as a single commit on `ci/release-workflow` with no merge commits. It exists to be compared with #1846's layout, not to be merged by default: merge #1846 then this, retarget this to `master` and merge it alone, or close it if the layout in #1846 is judged better.

**The question.** #1846 puts the release fences in `.github/scripts/`. The recon behind them considered a workspace member for the tooling and rejected it, mainly because *a member is a release candidate*: the plan job would accept `sphinx-needs-workspace-tools-v0` as a tag. Chris asked whether uv has Cargo's `publish = false`.

**It has the selector half of it.** `[tool.uv] package = false` makes a member *virtual*. Measured on uv 0.12.9, on this very member:

- `uv build --all-packages` builds only `sphinx-needs`; the virtual member is skipped.
- `uv build --package sphinx-needs-workspace-tools` exits 2: *"is missing a `build-system`"*.
- `uv.lock` records `source = { virtual = "tools" }`.
- `uv sync` installs its **dependencies** without installing it, so `import sn_tools` fails in `.venv`, which is right: the tooling is run by path and never imported by anything shipped.

It is **not a build prohibition**, and the review caught the arc claiming that it was: `uv build tools/` (the source-path form) falls through to PEP 517's default backend and produces a real sdist and wheel that `uv publish --dry-run` accepts. Nothing in this repository's workflows can reach that form, because the `build` job only ever runs `uv build --package "$DIST"` after `plan` has validated the tag. So what makes the member unreleasable is not uv but two fences this PR adds: the plan job's rule below, and the `Private :: Do Not Upload` classifier in its manifest, which PyPI rejects on upload.

### What changes

**Moved** (`git mv`, so `git log --follow` and `git blame` keep working): `check_workspace.py`, `release_plan.py` and `propagate_floors.py` to `tools/src/sn_tools/`; their tests to `tools/tests/`. `tools/` sits **outside `packages/`**, which keeps meaning "the distributions this repository publishes", and carries no `pkg:` label of its own.

**Stayed, with the rule now written in `AGENTS.md`:** what decides where a script lives is whether it must execute *inside* a particular environment. `check_sphinx_cell.py` imports the cell's own sphinx, `check_typing_floor.py` runs inside `.venvs/typing`, `extract_benchmark_data.py` runs inside the benchmark job; none could move without installing a member into every matrix cell. `scripts/smoke_needs.py` stays for the mirror-image reason: it exists to run *outside* every project environment.

**The plan job's new rule**, each part with a red proof in the branch: a tag naming a virtual member is refused before anything is built; the release order lists virtual members on their own unnumbered line, so "the order is A → B" never names something that cannot be released; a publishable member declaring a runtime or extra dependency on a virtual one is refused, because that wheel would name a distribution that is never on PyPI; and check 4 never turns such a dependency into an index lookup guaranteed to 404. `check_workspace.py` reports the same edge on every pull request, so the release job is not the first place anyone hears of it, and it also asserts that every virtual member carries the `Private :: Do Not Upload` classifier, so the second fence has a red proof too. A virtual member's own edges are held to the honesty check but not to the tight-tracking policy, and a virtual member may depend on another: nothing it declares ever ships.

**Invocation is unchanged in kind.** CI still runs the fences by path with `--no-project --with packaging`, so a manifest mistake is still *named* rather than reported as a failed `uv sync`. A virtual member has no console scripts, so none were added.

### The comparison

| | #1846: `.github/scripts/` | this PR: virtual member `tools/` |
|---|---|---|
| dependency contract | `packaging` borrowed from whatever sphinx pulls in; `tomlkit` declared twice, in the `dev` and `typing` groups | declared once, in `tools/pyproject.toml`; reaches every environment through the root's dependency on the member |
| where the tests live | `.github/scripts/selftests/`, named to avoid a second importable `tests`; root `pythonpath` removed for the same reason | `tools/tests/` beside the code; its `conftest.py` puts `tools/src` (never `tools/`) on `sys.path` |
| what every environment installs | nothing extra | `packaging` and `tomlkit` in every environment, matrix cells included (about 50 KB, no transitive dependencies) |
| what the plan job knows | every member is releasable | virtual members are listed apart and refused; a runtime or extra edge onto one is an error in the plan and in the fence; this rule, not uv, is what makes the member unreleasable |
| what the fence's tracking policy covers | every intra-workspace edge | edges from publishable members; a virtual member may floor a sibling loosely, since it is never co-installed with anything |
| conventions for where a script lives | one: `.github/scripts/` | two, with the rule stated in `AGENTS.md`: `tools/` for the workspace's tooling, `.github/scripts/` for checks bound to a CI environment |
| members in every "for each member" output | one | two (`uv tree`, the fence's OK lines, the release order) |

One detail worth a reviewer's eye: `tools/tests/conftest.py` inserts `tools/src` on `sys.path`, never `tools/`. The directory is called `tests`, and putting its parent on a path root would make it a top-level package competing with the sphinx-needs suite's own for that name, the failure mode #1846's last round removed. `tools/src` contains exactly one name, `sn_tools`.

### Rehearsal

[Run 33932496976](https://github.com/useblocks/sphinx-needs/actions/runs/33932496976), dispatched from this branch, is green: `plan` prints the release order with the virtual member on its own line,

```
  1. sphinx-needs 8.5.0   depends on: -
  -- sphinx-needs-workspace-tools 0   (virtual -- never published)
```

then `build` runs the fence from its new path, builds the wheel, resolves it against PyPI alone and runs the compat cell (1740 passed); `publish` and `release` are skipped.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -14,6 +14,7 @@
           - '.claude/**'
           - 'docker/**'
           - 'scripts/**'
+          - 'tools/**'
           - 'pyproject.toml'
           - '.pre-commit-config.yaml'
           - '.readthedocs.yml'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,7 +25,7 @@ jobs:
         # (see the script's docstring). The local prek hook runs the same script against
         # the synced environment, so `uv run poe lint` covers it too
         run: >
-          uv run --no-project --with packaging python .github/scripts/check_workspace.py
+          uv run --no-project --with packaging python tools/src/sn_tools/check_workspace.py
       - run: uv run --frozen prek run --all-files --show-diff-on-failure
         env:
           PREK_COLOR: always
@@ -92,15 +92,12 @@ jobs:
         env:
           UV_PROJECT_ENVIRONMENT: .venvs/typing
         run: uv run --frozen --no-sync --no-group dev --group typing python .github/scripts/check_typing_floor.py
-      - name: Test the repository's own scripts
-        # `.github/scripts/` holds the release fences: the one that decides whether a tag
-        # may publish, and the one that rewrites floors. They are not in any package, so
-        # this is the only job that runs their tests. They are in the root `testpaths`, so
-        # a bare `pytest` at the root runs them too -- but every matrix cell names the
-        # package's `tests` path explicitly and so never collects them. The directory is
-        # `selftests/` rather than `tests/` so that no importable `tests` exists outside a
-        # package
-        run: uv run --frozen --no-sync pytest .github/scripts/selftests -q
+      - name: Test the repository's own tooling
+        # `tools/` holds the release fences: the one that decides whether a tag may publish,
+        # and the one that rewrites floors. Nothing else in CI exercises them, and they are
+        # in the root `testpaths`, so a bare `pytest` at the root runs them too -- but every
+        # matrix cell names the package's `tests` path explicitly and so never collects them
+        run: uv run --frozen --no-sync pytest tools/tests -q
 
   # ubc:
   #   name: ubc
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -58,7 +58,7 @@ jobs:
       # skipping either check, so a rehearsal exercises `git merge-base` and the
       # `fetch-depth: 0` above rather than leaving them for the first real tag.
       - id: plan
-        run: uv run --no-project --with packaging python .github/scripts/release_plan.py --tag "$TAG" --github-output ${{ github.event_name == 'workflow_dispatch' && '--rehearsal' || '' }}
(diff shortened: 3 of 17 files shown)


chrisjsewell:
Closing and reopening only to re-trigger the Read the Docs build: build 34402537 died during `apt-get update` before any project step (no exit code recorded), and the same docs build succeeds locally with warnings as errors.

chrisjsewell:
Retrying the Read the Docs build once more: builds 34402537 and 34402670 both hit RTD's 900 s build timeout inside `apt-get update`, before any project step, while the identical configuration passed for #1846 two hours earlier.

chrisjsewell:
Folded into #1846: `ci/release-workflow` was fast-forwarded to this branch's tip (`b9d3cf92`), so GitHub marked this pull request as merged into that branch. Review the combined change on #1846; nothing else happened here.

Issue #46, opened 2018-05-16
missing extra blank line


weegreenblobbie:
https://github.com/useblocks/sphinxcontrib-needs/blob/f06291acae75de53f76cfacfc0390700eb9658dd/sphinxcontrib/needs/needs.py#L52

I'm getting a warning for each use of this directive:

    WARNING: Explicit markup ends without a blank line; unexpected unindent.
    WARNING: Explicit markup ends without a blank line; unexpected unindent.
    WARNING: Explicit markup ends without a blank line; unexpected unindent.

Adding a blank line here solves it.

danwos:
Thanks @weegreenblobbie for bug report and also for a solution.

I just have fixed it and released it under 0.2.1:
http://sphinxcontrib-needs.readthedocs.io/en/latest/changelog.html#id1

Feel free to test it and give me some feedback.

Issue #42, opened 2018-03-14
AttributeError: 'StandaloneHTMLBuilder' object has no attribute 'status_iterator'


danielvasas:
Hello,

after latest update of my environment, and with that a couple of related packages, I'm getting this exception.

Sphinx version: 1.7.1
Python version: 3.6.2 (CPython)
Docutils version: 0.14
Jinja2 version: 2.10
Last messages:
reading sources... [ 89%] contents/doc_user/introduction/index
reading sources... [ 92%] contents/doc_user/quickstart/index
reading sources... [ 96%] contents/doc_user/usage/index
reading sources... [100%] index
Loaded extensions:
alabaster (0.7.10) from c:\dev\ve_safe_build_36\lib\site-packages\alabaster_init_.py
sphinx.ext.autodoc (1.7.1) from c:\dev\ve_safe_build_36\lib\site-packages\sphinx\ext\autodoc_init_.py
sphinxcontrib.plantuml (unknown version) from c:\dev\ve_safe_build_36\lib\site-packages\sphinxcontrib\plantuml.py
sphinxcontrib.needs (0.1.49) from c:\dev\ve_safe_build_36\lib\site-packages\sphinxcontrib\needs_init_.py
Traceback (most recent call last):
File "c:\dev\ve_safe_build_36\lib\site-packages\sphinx\cmdline.py", line 304, in main
app.build(args.force_all, filenames)
File "c:\dev\ve_safe_build_36\lib\site-packages\sphinx\application.py", line 331, in build
self.builder.build_update()
File "c:\dev\ve_safe_build_36\lib\site-packages\sphinx\builders_init_.py", line 333, in build_update
'out of date' % len(to_build))
File "c:\dev\ve_safe_build_36\lib\site-packages\sphinx\builders_init_.py", line 346, in build
updated_docnames = set(self.env.update(self.config, self.srcdir, self.doctreedir))
File "c:\dev\ve_safe_build_36\lib\site-packages\sphinx\environment_init_.py", line 571, in update
for retval in self.app.emit('env-updated', self):
File "c:\dev\ve_safe_build_36\lib\site-packages\sphinx\application.py", line 444, in emit
return self.events.emit(event, self, *args)
File "c:\dev\ve_safe_build_36\lib\site-packages\sphinx\events.py", line 79, in emit
results.append(callback(*args))
[...]

Issue #41, opened 2018-03-06
Compatibility issue with Sphinx>=1.7


ubmarco:
When running sphinxcontrib-needs against Sphinx 1.7.1 I get the following error:

```
reading sources... [100%] index                                                                                                                                                     

Exception occurred:
  File "/useblocks/dev/bmw/ascgit307/py36_venv_common/lib/python3.6/site-packages/sphinxcontrib/needs/environment.py", line 18, in install_backend_static_files
    for source_file_path in app.builder.status_iterator(
AttributeError: 'StandaloneHTMLBuilder' object has no attribute 'status_iterator'
The full traceback has been saved in /tmp/sphinx-err-lb7n17a0.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
Makefile:53: recipe for target 'html' failed
make: *** [html] Error 2
```

There are no issues with Sphinx 1.6.7.

danwos:
Bugfix and tests for 1.7 were added to 0.2.0 branch.
I hope this will be released the next days....

PR #1850, opened 2026-09-05
🔧 Import sphinx-mounts into the workspace as packages/sphinx-mounts

chrisjsewell:
> **Merge with "Create a merge commit"**, and give the merge commit this PR's title (either edit the title in the merge dialog, or set the repository's merge-commit title to *Pull request title* first): the repository's current default would put `Merge pull request #NNNN from useblocks/import/sphinx-mounts` on master's first-parent line. Do not squash: the whole point is the preserved history.

`useblocks/sphinx-mounts` moves into this workspace as `packages/sphinx-mounts/`, with its
history, its tags and its release path.

### What the import is

The whole of that repository's `main` — 33 commits, tip `8fcae538` — is merged in, rewritten
by `git filter-repo` so that every historical commit already places its files under
`packages/sphinx-mounts/`. Three things came with it:

- **Tags.** Its six releases arrive prefixed: `sphinx-mounts-v0.1.0` … `sphinx-mounts-v0.2.0`.
  The bare tag namespace stays sphinx-needs': `release_plan.py` reads every tag without `-v`
  in it as a sphinx-needs release, so unprefixed mounts tags would be read as sphinx-needs
  0.1.0–0.2.0. Nothing in the wild pins sphinx-mounts from git — both known dependants take
  it from PyPI — so the rename costs nobody a resolvable pin.
- **A scoped commit-message rewrite.** Bare `#N` references became
  `useblocks/sphinx-mounts#N`, 53 of them, bounded to mounts' own PR range (`#1`–`#46`).
  A blanket rule would have damaged the history: dependabot commit bodies embed upstream
  release notes verbatim, complete with bare refs to other repositories. Already-qualified
  cross-repo refs (`useblocks/sphinx-needs#1750`, `useblocks/ubcode#3109`) are untouched, and
  so is `actions/setup-python#11`-shaped text.
- **The commit map**, committed as `packages/sphinx-mounts/design/import-commit-map.txt`.
  The rewrite destroys the old hashes, and links into the old repository name them — for
  example eclipse-score/score's `DR-008-infra.rst` points at a blob at `294526a0…`. The old
  repository will be archived, not deleted or transferred, so those permalinks keep serving;
  this file is what carries them across.

`git log packages/sphinx-mounts/<path>` works **with no `--follow`** — filter-repo rewrote
the paths, so from this repository's point of view those files were always there. That is
the opposite of the sphinx-needs files the 2026-09 restructure moved, which do need it.
`git blame` needs nothing either way, and `git log --first-parent master` shows the whole
import as one line.

### The wiring

The root lists the member and sources it from the workspace; `docs-mounts` re-exports its
`docs` extra; the shared `test` group gains `beautifulsoup4` and `sphinxcontrib-mermaid`
and states sphinx-mounts' `sphinxcontrib-plantuml>=0.30` floor; pytest collects the package
and knows the `bazel` marker; ruff gets its source roots and ty its `include`. `uv lock`
adds exactly three packages (ignore-python, sphinx-mounts, sphinxcontrib-mermaid), +34 KB.

The member's manifest is cut back to `[project]` and `[build-system]`. Its `tox.ini`,
`constraints/`, `.github/`, `.python-version`, `.pre-commit-config.yaml`, `.gitignore` and
`uv.lock` are deleted — the root owns every one of them — and so is
`tests/test_sphinx_version.py`, whose job `.github/scripts/check_sphinx_cell.py` already
does more strongly for every cell.

Bringing the package under the root's ruff and ty configuration cost 24 ruff findings (22
auto-fixed, 2 by hand) and turned up **one real bug**: `_MountAwareProject.__init__`
annotated `srcdir: str | Path`, but on the declared sphinx floor `Project.srcdir` is
`str | os.PathLike[str]`. mounts' own type check ran against whatever the default resolve
gave — the newest sphinx — and structurally could not see it. The `typing` group earned its
keep on day one.

### Two new fences, and the false greens they close

**1. `check_workspace.py` check (7): no member may carry a table the root owns**
(`[dependency-groups]`, `[tool.ruff]`, `[tool.pytest]`, `[tool.ty]`). The ruff case is the
one that matters: ruff resolves configuration per file by walking up to the nearest
`pyproject.toml` with a `[tool.ruff]` section, so a member table does not extend the root's
ruleset, it **replaces** it. Measured on this package before its table was deleted: 35
findings across six rules, four of them rules the root does not enable, and the root's own
`I001` gone entirely — with the lock unchanged and every command exiting 0.

**2. `compat-requirements.txt`, and three tests that assert instead of skipping.** The
compat cell in `release.yaml` is the only content-aware gate in the release pipeline: it
runs the member's suite against its dependencies as PUBLISHED. For sphinx-mounts it was
green from the first line with the only three sphinx-needs integration tests silently
skipped — `pytest.importorskip("sphinx_needs")`, and sphinx-needs is a test-only dependency
that reaches neither the wheel's `Requires-Dist` nor that environment. It cannot be fixed
in the root `test` group either: that group resolves through `[tool.uv.sources]`, so
`uv export --only-group test` emits `-e ./packages/sphinx-needs` and the cell would test the
checkout. So a plain per-package requirements file, installed before the wheel, and the
three `importorskip`s become assertions — which is what makes that file load-bearing rather
than decorative. The cell now reports 942 passed / 2 skipped with all three passing.

### CI

`tests-mounts` (ubuntu × sphinx-7/8/9 on the pinned interpreter, plus one `windows-latest`
cell on sphinx-9 — this package's subject is paths and symlinks) and a separate `bazel`
lane; both join `check`. Cost: about **+11 runner-minutes per push (+8.5 %)** against the
128.5 the last master run took, and **no change to wall time** — the longest job in the
graph is a 15-minute Windows sphinx-needs cell.

The renderer is the interesting part. mounts' uml tests assert on a PlantUML rather than
skipping, and no runner image has one; its own CI installed a package per cell (15–22 s
each, plus a Windows `plantumlc` shim). Instead `_require_renderer` now accepts a jar named
by `PLANTUML_JAR`, and the workflows point it at the jar `packages/sphinx-needs` already
vendors for its own suite — 0 s of install, one visible line each in `ci.yaml` and
`release.yaml`. That is the whole of the cross-package coupling, declared where it is
created: if sphinx-needs ever stops vendoring the jar, those lines fail loudly.

### The docs, and why RTD forces the rename

`conf.py` sat in `docs/` with its sources in `docs/source/` — a split only `sphinx-build -c`
can bridge. Read the Docs does not offer `-c`: it names a `conf.py` and runs sphinx with
that file's own directory as the source directory. Against the old layout that is
`Sphinx is unable to load the master document`, exit 2. And RTD is not optional here: this
repository's GitHub Pages site is already a branch deployment of the benchmark history, and
a repository gets exactly one. So the sources come up to meet `conf.py` (its own commit, so
the rename is reviewable), and `packages/sphinx-mounts/.readthedocs.yaml` is added for the
new project to point at. Every path in that file is relative to the repository root, not to
the file — RTD's documented rule for a monorepo.

### Deliberately not in this pull request

- the `end-of-file-fixer` / `trailing-whitespace` prek hooks mounts had and this repository
  does not: adding them is a whole-tree reformat and belongs in its own change;
- adopting mounts' ten extra ruff rule families (`S ARG FURB PERF PGH PIE PL PTH SLF T20`)
  repo-wide — the root's comment already parks that decision;
- `py.typed` for either package;
- the RTD project, custom domain and DNS move;
- the issue transfer and the archiving of the old repository;
- the `post_checkout` exit-183 job that would cancel an RTD build when nothing under the
  package changed: RTD's commit-status behaviour for a cancelled build is unverified;
- re-syncing `tests/fixtures/variant_condition_conformance.toml` with ubCode. Its header
  says "`useblocks/sphinx-mounts` carries this file at …", which this import makes
  inaccurate — but the file is byte-identical to ubCode's, which pins its digest, so the
  edit is owed there first;
- a commit-message convention in `AGENTS.md`. mounts' `✨ NEW:` / `🐛 FIX:` keyword form is
  not this repository's, so it is deleted rather than moved; nothing documents the
  emoji-first form master actually uses.

### After the merge

1. **Push the six tags** — `git push origin 'refs/tags/sphinx-mounts-v*'`. Deliberately
   after, not before: pushed early they would make the mounts commits reachable while the
   code is not on master, and `release_plan.py` would already report a second member.
2. **Read the Docs** (Chris), in this order — the order matters because RTD's automation
   rules run when a version is *created*, and both recommended actions *build* the version:
   a. the six tags are pushed first (step 1), so RTD's initial sync creates those six versions
      before any rule exists — a rule added first would activate and build all six, and all
      six fail, because their trees predate `packages/sphinx-mounts/.readthedocs.yaml`;
   b. create the mounts project against this repository; set *Admin → Settings → Build
      configuration file* to `packages/sphinx-mounts/.readthedocs.yaml`. Do not touch the
      sphinx-needs project's setting — it applies to every version that project builds;
   c. let the default `latest` build go green once;
   d. add the two automation rules on `^sphinx-mounts-v\d+\.\d+\.\d+$` — *Activate
      version* and *Set version as default*. They apply to versions created after they exist,
      so they are for the next release, not for the six. RTD derives `stable` with
      `packaging.version.Version(<tag>)` and drops what it cannot parse; a prefixed tag is not
      PEP 440, so without the second rule the project never gets a default;
   e. set the default version by hand this once, or leave it at `latest` until the next
      release lets the rule do it;
   f. if a historical version is ever activated (by hand or by a re-sync), expect a red build
      and deactivate it: those six are unbuildable by construction, not a regression.
3. **The domain** (Chris): add `sphinx-mounts.useblocks.com` as an RTD custom domain, then
   move the DNS CNAME off `useblocks.github.io`. Until that step the OLD repository's Pages
   deployment IS the live site, so there is no window with no site.
4. **The issues** (orchestrator): transfer the seven open issues, then label them
   `pkg: sphinx-mounts`. Transfer needs write access on both repositories, so it must happen
   before archiving. Transfer renumbers. (Correction after the transfer: GitHub rewrites every
   `#NN` in the bodies to `owner/repo#NN`, following transferred issues to their new numbers,
   so the hand pass found nothing to edit.) Note that the `pkg: sphinx-mounts` labeler entry
   on this branch is unproven until then: the label job runs as `pull_request_target` and reads
   `.github/labeler.yml` from master, so its green on this PR reflects labels applied by hand.
5. **The four open pull requests** (Marco): #46 (`example-variants`) filters cleanly out of
   the import and can be reopened here; #5 is three months stale on an unrelated branch;
   #41 and #42 are dependabot and this repository's own dependabot will re-raise anything
   still applicable.
6. **Archive the old repository** — never delete, never transfer. An archived repository
   keeps serving every blob permalink, every branch and every issue-URL redirect, and one
   external permalink to a pre-rewrite hash depends on exactly that.

Refs #1803


ubmarco (approved):
Approved, but I am concerned about this.
We need better CI action prefixes/names, to quickly identify where the job belongs to.
Look at
https://github.com/useblocks/sphinx-needs/pull/1850/checks

Image

Image

Can be done with this PR, or in a follow-up.

ubmarco:
Nice, good addition. Reaffirming my approval :)

ubmarco (approved):

PR #1861, opened 2026-09-05
🧪 Probe: what status does a Read the Docs build cancelled with exit 183 report?

chrisjsewell:
The sphinx-mounts Read the Docs configuration on this branch cancels every pull-request build unconditionally through a `post_checkout` job exiting 183. The only purpose is to observe the `docs/readthedocs.org:sphinx-mounts` commit status such a cancelled build reports (success, failure, pending forever, or nothing), which decides whether the real conditional job (cancel when nothing under `packages/sphinx-mounts/` changed) can be used on a project whose status is required. Not for merging; closed once recorded.

### DIFF
--- a/packages/sphinx-mounts/.readthedocs.yaml
+++ b/packages/sphinx-mounts/.readthedocs.yaml
@@ -17,6 +17,17 @@ build:
   os: ubuntu-24.04
   tools:
     python: "3.13"
+  jobs:
+    post_checkout:
+      # PROBE ONLY -- never to be merged. Cancels EVERY pull-request build of this project
+      # unconditionally, to observe what commit status Read the Docs reports for a build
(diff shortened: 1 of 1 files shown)


chrisjsewell:
Answered. The mounts build for this pull request (34411602) was cancelled by the `exit 183` in `post_checkout` after 10 seconds, Read the Docs recorded it as `cancelled`, and the commit status it posted is `docs/readthedocs.org:sphinx-mounts` = **success**, description "Read the Docs build skipped." So a conditional cancel job leaves a green, non-blocking status, which means it is safe even on a project whose status is a required check. The real job (cancel a pull-request build when nothing under `packages/sphinx-mounts/` changed) follows as a normal pull request. Closing without merging, as intended.

PR #1868, opened 2026-09-06
📚 Document nested ports in needflow

### LINKED ISSUE #980: Docu: Add example on how to realize port display in needflow
When using `style="portin"`, "portout" or "port" for a needs_type  and using nested needs to put the ports under another element (e.g. a block) a needflow covering those elements will display them nicely.
Additionally a needs_diagram_template for the port to like this
```
needs_diagram_template = """
{%- if (type_name == "Publisher Port") -%}
<size:12>{{type}}</size>
...
```
to only to display the title of the port is very helpful to clean up clutter in the graph.

Maybe a scenario like this could be added to the documentation, since the usage of ports is quite common (e.g. in SysML)

roli-lpci:
Closes #980.

This adds a copyable configuration and RST example for displaying `portin` and
`portout` needs inside a parent block. It also shows how to keep port labels
compact with `needs_diagram_template` and explicitly selects the PlantUML
needflow engine required by those styles.

Since this documents existing behavior, the already released 8.5.0 changelog
is unchanged.

Validation: `uv run poe docs-needs`

This contribution was produced by agents through [Hermes Labs](https://hermes-labs.ai)’ engineering infrastructure. [Rolando Bosch](https://github.com/roli-lpci) is the responsible human contributor and authorized publication from his personal GitHub account.


chrisjsewell (approved):
Thanks — this is a clear, copyable answer to #980, and I verified it rather than reading it:

- Built the example as written (the `conf.py` and the RST, with `-W`) against the PlantUML jar this repository pins. Zero warnings. The emitted PlantUML nests the two ports inside the `CONTROLLER` node block, and the rendered diagram shows `Sensor input` and `Actuator command` as compact ports on the block, with the block itself keeping the default three-line label — exactly what the text says.
- `type` is in the template's render context (`filter_context()`), so `{%- if type in ["inport", "outport"] -%}` is the right key.
- `:engine: plantuml` is belt-and-braces (the `needs_flow_engine` default is already `plantuml`), but it is the correct thing to show given the note that the port styles are PlantUML-only.
- Heading level and placement under `needs_diagram_template` are consistent with the rest of the page; the RTD build is green.

No changelog entry is needed for documentation of existing behaviour. Merging once the CI re-run on the updated branch is green.

PR #1870, opened 2026-09-06
🔧 One pinned PlantUML for the whole workspace, vendored at vendor/plantuml/

chrisjsewell:
> **Note on the merge commit.** This landed as `951fa8ed` while its description was still the one written for the earlier, fetched design, so the squash commit's message carries that title and text. This page is the record of what actually merged: the jar is **committed**, not fetched.

## Why

This repository vendored **two** PlantUML jars, at **two** versions, and downloaded a third:

| where | version | bytes |
|---|---|---|
| `packages/sphinx-needs/tests/doc_test/utils/plantuml.jar` | 1.2022.5 | 10,071,904 |
| `packages/sphinx-needs/docs/utils/plantuml-1.2022.14.jar` | 1.2022.14 | 11,349,479 |
| `docker/Dockerfile` | `releases/latest`, unpinned | — |

* **The sdist.** `[tool.flit.sdist]` ships `tests/` and `docs/`, so both jars went to PyPI with
  every release: the published sdist was ten times its own wheel. Moving the jar to a
  repository-root directory is what takes it out of the tarball — flit's `include` patterns
  cannot escape the package directory — and that is true whether or not the jar is committed.
* **One version, said out loud.** The test jar was four years and ~46 releases old and nothing
  in the tree said so: its filename carried no version at all. The version now lives in
  `vendor/plantuml/pin.toml` and in the filename, and `uv run poe verify-plantuml` — which
  CI's Lint job runs — fails when the two disagree.
* **Zero network.** A checkout renders. Nothing has to reach a host to draw a diagram: not the
  22 of a CI run's 25 jobs that render (measured on a run of this branch; the three that do
  not are `Lint`, the smoke test and the `check` aggregator), not a Read the Docs build, not an offline machine, and not a sandboxed
  agent session whose allowlist is set on the environment rather than in this repository.
* **The path was retyped eight times**, across two workflows and two `AGENTS.md` files, plus a
  ninth in `performance/project/conf.template` that pointed at no file in any tree. And it did
  not survive a second package: sphinx-mounts reaches PlantUML through `PLANTUML_JAR`, which
  CI pointed at sphinx-needs' jar — a cross-package coupling this workspace has none of
  elsewhere.

### The design history, honestly

**This pull request built and reviewed a *fetched* design first, and dropped it.** In that
version `pin.toml` was committed, the jar was gitignored, and every consumer downloaded it on
demand — a fetch script, a `deps` on every rendering task, a `post_install` job on Read the
Docs, and a fetch step in seven CI jobs. It worked, and it was reviewed green. It was dropped
because of where the renderer then has to be reachable from: the release URL redirects to
`release-assets.githubusercontent.com` (measured), and that host has to be reachable by those
22 jobs, by every Read the Docs build, by every offline machine, and by every sandboxed
session — whose network allowlist lives on the environment and cannot be set from this
repository (this repository's own `CLAUDE.md` records `api.github.com` having to be added by
hand). A committed jar needs none of it, for a file that does not change between bumps.

**The cost is stated rather than hidden:** one ~30 MB blob enters the history at each bump.
The repository already carried six jar blobs, 49,933,998 bytes, from the two it used to
vendor; this makes seven and 79,805,495. That is accepted because bumps are rare — the jar
this replaces sat four years — but PlantUML's jar has tripled in that time (10.1 MB → 29.9 MB)
and **GitHub warns above 50 MB and refuses above 100 MB**, so `vendor/plantuml/README.md` names
that as the thing to watch rather than leaving it to be rediscovered.

## What this does

**One jar, one version, committed and fenced.** `vendor/plantuml/pin.toml` is the only place
the workspace writes a PlantUML version:

```toml
version = "1.2026.8"
sha256 = "5e1ecfa8ecd32c90b03bbf3b1eb6f020943f98ab0fcf4032be31a0002ee2c462"
url = "https://github.com/plantuml/plantuml/releases/download/v{version}/plantuml-{version}.jar"
```

`vendor/plantuml/plantuml-1.2026.8.jar` beside it is exactly those bytes.
`tools/src/sn_tools/fetch_plantuml.py` is stdlib-only — `urllib`, `hashlib`, `tomllib` — and
has two jobs: `--verify` (**never** touches the network) checks the committed jar against the
pin and prints its path, and the default mode downloads the jar the pin names, which is what a
bump does.

### The resolution order, everywhere

1. **`PLANTUML_JAR`** — an explicit choice wins, and a value naming no file is an error, never
   a silent fall-through.
2. **`vendor/plantuml/plantuml-<pinned version>.jar`** — what a checkout has.
3. **`plantuml` on `PATH`** (`plantumlc` first on Windows, whose chocolatey `plantuml` shim is
   a non-blocking `javaw` launcher).

Failing all three the error names the routes that tree actually has: in a checkout the poe task
leads; in an unpacked sdist, which ships neither the jar nor the pin, `PLANTUML_JAR` and the
executable lead and the task is mentioned last.

### Per site

| site | change |
|---|---|
| `vendor/plantuml/` | `pin.toml`, the committed `plantuml-1.2026.8.jar`, and a `README.md` that explains the bargain and the bump |
| `.gitignore` / `.gitattributes` | no `*.jar` ignore any more (the `.part` an interrupted download leaves is still ignored); the jar is marked `binary`, so `text=auto` is not asked to classify a 30 MB zip |
| `tools/src/sn_tools/fetch_plantuml.py` | the bump tool plus `--verify`, the fence |
| `pyproject.toml` | `poe verify-plantuml` (the fence, declared by `poe lint`) and `poe fetch-plantuml` (the bump step; on a checkout, a 0.03 s hash check). Every rendering task keeps `deps = ["fetch-plantuml"]`; the sphinx-mounts tasks keep `uses = { PLANTUML_JAR = "fetch-plantuml" }`, because that suite reads the variable and never a path |
| `tests/conftest.py`, `docs/conf.py`, `performance/performance_test.py` | the same three-step resolution, written out once each (a `conf.py` can import neither `tests/` nor `tools/`) |
| `tests/test_plantuml_command.py` | 16 cases pinning that order, including the sdist shape |
| `directives/needgantt.py` | a float completion is rounded, the way a float duration already was — see **One rendering change the bump exposed** |
| `performance/` | the dead template path is fixed by moving the resolution out of the template — see below |
| sphinx-mounts | messages and prose only; its `PLANTUML_JAR`-or-`PATH` resolution is untouched |
| `ci.yaml`, `docs.yaml`, `benchmark.yaml`, `release.yaml`, `test-package.yaml` | seven **"Point PLANTUML_JAR at the vendored jar"** steps: `--verify`, then `PLANTUML_JAR` from what it printed. Two job-level `env:` blocks became steps; `tests-core` (14 cells), `tests-js`, `tests-no-mpl`, the linkcheck and the benchmark job were all getting a renderer *by accident* from a jar that happened to be in the tree, and now say so. The Lint job runs `--verify` itself, beside the workspace-manifest check |
| `.readthedocs.yml` | nothing to install: the checkout carries the jar. The exit-183 skip filter keeps `vendor/` (`docs/conf.py` reads the pin) and drops `tools/` |
| `docker/Dockerfile` | `ARG PLANTUML_VERSION=1.2026.8` and the versioned URL, replacing `releases/latest` — the one place the renderer could change without a commit. The image builds from a git ref with no checkout as its context, so it still downloads its own copy; a Dockerfile cannot read TOML, and the duplication is declared in a comment |

**`performance/project/conf.template` was dead, and fixing the filename would not have helped.**
`performance_test.start()` copies `project/` into a throwaway tempdir and renders the template
*there*, so the generated `conf.py`'s `__file__` is in that tempdir and
`../../docs/utils/plantuml-….jar` never resolved to a file in any tree. `performance/` is not
run by CI, so nothing caught it. The resolution moves to `performance_test.py`, whose own
`__file__` is in the checkout.

## What a bump looks like

1. Download the asset and `shasum -a 256` it.
2. Edit `version` and `sha256` in `vendor/plantuml/pin.toml`.
3. `uv run poe fetch-plantuml`, then `git rm` the old jar and `git add` the new one.
4. Edit `ARG PLANTUML_VERSION` in `docker/Dockerfile`.
5. `uv run poe verify-plantuml` and re-run the renderer-heavy suites.

Forget step 3 and **Lint goes red naming both hashes** — that is the whole point of the fence,
because the jar is the half of a bump a reviewer cannot see in a diff. One thing worth knowing:
pushing a 30 MB blob over an HTTPS remote can answer `error: RPC failed; HTTP 400`; it is
git's chunked-transfer switch, not a refusal, and `git -c http.postBuffer=629145600 push` sends
the same objects. The README says so.

## What it costs, and what it buys

**The sdist:**

```
before   28.0 MB sdist  |  2.8 MB wheel   (10.1 : 1)
after    ≈8.1 MB sdist  |  2.8 MB wheel   ( 2.9 : 1)     -71 %
$ tar tzvf sphinx_needs-8.5.0.tar.gz | grep -ci jar   →  0
```

(Approximate on purpose: the tarball stores mtimes, so gzip lands a few hundred bytes apart on
different machines — always the same 1379 entries and always zero jars.)

**An sdist consumer** — a distribution packager, who repacks an embedded pre-built jar anyway —
gets a tarball with no jar and two supported routes: `PLANTUML_JAR=/usr/share/plantuml/plantuml.jar`,
or a `plantuml` executable on `PATH`. Both suites honour them, and both **assert** rather than
skip when neither is there, so a packager gets a message rather than a quieter test run.

**A clone costs ~30 MB more and needs nothing else.** No fetch step, no cache, no network
allowlist entry, no `post_install` job — and no new failure mode in any of the four places a
download would have had to work.

## The version bump: 1.2022.5 → 1.2026.8

```
$ pytest tests/test_plantuml.py tests/test_plantuml_incdir.py tests/test_needuml.py tests/test_needflow.py
143 passed, 1258 warnings in 197.92s (0:03:17)      # "7 snapshots passed"
```

Measured on these exact bytes earlier in this pull request, and re-run on the committed jar for
`tests/test_plantuml.py` since. Per-invocation cost is unchanged: 1.0 s (1.2022.14) vs 1.1 s
(1.2026.8) for a trivial diagram, three runs each.

## The gates

All of these on the merged head, with `origin/master` merged in first:

```
$ poe lint                    8/8 hooks pass, with `verify-plantuml` as its dep
$ poe typecheck               All checks passed!
$ poe check-workspace         10 OK lines
$ pytest tools/tests -q       263 passed          # 6 new, all with the network made to raise
$ poe verify-plantuml         0.162 s, no network
$ poe fetch-plantuml          0.157 s -- a hash check of the committed jar
$ poe test-needs tests/test_plantuml.py     3 passed, 104 warnings in 7.63s
$ poe test-mounts             943 passed, 2 deselected in 24.95s
$ poe docs-needs (cold)       build succeeded, 0 warnings, 1m01s
$ poe smoke-needs             pass -- 1379 sdist entries, 0 jar entries, 0 vendor entries
$ actionlint .github/workflows/*.yaml       exit 0, no findings
```

and, because the committed jar is the thing that has to survive a checkout:

```
$ git clone --depth 1 --branch tooling/fetch-plantuml … && shasum -a 256 vendor/plantuml/*.jar
5e1ecfa8ecd32c90b03bbf3b1eb6f020943f98ab0fcf4032be31a0002ee2c462     # the pin's hash
$ env -i … https_proxy=http://127.0.0.1:9 python tools/src/sn_tools/fetch_plantuml.py --verify
…/vendor/plantuml/plantuml-1.2026.8.jar                              # exit 0, offline
```

Flip one hex digit of the pin's `sha256` and `poe lint` and the Lint job both stop before
anything else runs, naming both hashes; `git rm` the jar and `--verify` names the path and the
fix. The four plantuml-heavy test files (143 passed, seven syrupy snapshots unchanged) were
measured on these exact bytes earlier in this pull request, on two machines.

## One rendering change the bump exposed

One commit is a 🐛, and it is here because the bump found a real bug rather than because the
bump caused one. On 1.2026.8 the docs built with exactly one warning:

```
docs/directives/needgantt.rst:296: WARNING: error while running plantuml
b'…ERROR\n13\nSyntax Error? (Assumed diagram type: gantt)\n' [plantuml]
```

Captured from the real build (`-D plantuml=<a wrapper that saves every diagram>`), the failing
statement is `[ACT_CREATE_EX_C] is 90.0% completed`. A/B on that exact file, and on a copy with
the percentages changed to integers:

| the generated gantt | 1.2022.14 (the old vendored jar) | 1.2026.8 |
|---|---|---|
| as generated — `is 90.0% completed` | exit 0 | **exit 200, `Syntax Error? (Assumed diagram type: gantt)`** |
| `is 90% completed` | exit 0 | exit 0 |

**PlantUML's gantt grammar takes an integer percentage; 1.2022.14 tolerated a fractional one
and 1.2026.8 enforces it.** The `.0` is ours: `needgantt.py` rounds a float *duration* ("float
durations not supported by plantuml") and did not round the *completion*, and a completion
field typed `number` — which is what this project's own docs use, `amount` in
`docs/ubproject.toml`, pointed at by `:completion_option: amount` — delivers a float. Any user
with such a field had the same latent bug; the four-year-old renderer simply tolerated it.

The fix mirrors the duration handling exactly, and the new test in `tests/test_needgantt.py`
asserts on the **generated PlantUML source** (the shape that file already uses) rather than on
a rendered image. It is red without the fix, with the docs build's own message.
`uv run poe docs-needs` on 1.2026.8 is `build succeeded.` with **0 warnings**, and Read the
Docs' `fail_on_warning: true` is satisfied.

## Changelog bullets (draft — the release engineer stamps them)

```rst
- 🐛 ``needgantt``: a completion value that is a float (a ``number``-typed field) is now
  rounded to the integer percentage PlantUML's gantt grammar takes; newer PlantUML releases
  reject ``90.0%`` outright (:pr:`XXXX`).

- 🔧 One PlantUML for the whole workspace, at ``vendor/plantuml/``, pinned and checksummed by
  ``vendor/plantuml/pin.toml`` and bumped from 1.2022.5 (tests) / 1.2022.14 (docs) to 1.2026.8
  (:pr:`XXXX`). **The sdist shrinks from 28.0 MB to ≈8.1 MB** — the two jars it used to ship
  were 71 % of it — and no longer contains a renderer. Building or testing from the sdist
  therefore needs a PlantUML of your own: set ``PLANTUML_JAR`` to a jar (with ``java`` on
  ``PATH``) or install a ``plantuml`` executable; both are honoured ahead of anything in the
  tree. A checkout is unaffected — it carries the jar.
```

## Reviewed

A second pair of eyes reproduced every measurement here on another machine, against the fetched
design, and found one blocking defect that is fixed and still fixed under this one:
**`ci.yaml`'s `tests-core` — 14 cells, the whole sphinx-needs suite — called the reusable unit
without `needs-plantuml`**, whose default is `false`, and the unit runs bare `pytest` rather
than `poe`, so nothing pointed those cells at a jar. That is not only the uml tests: `test_app`
takes the `plantuml_command` fixture, which raises rather than skipping, so **478 tests errored**
in that shape. One line in the caller.

Four smaller findings from the same review are in: the Read the Docs skip filter, `| tr -d '\r'`
on every captured path with `shell: bash` so the pipe keeps the step fail-fast
(`bash -eo pipefail`), a `PLANTUML_JAR` naming no file refused by the script instead of
"ignored" and then killing the run at the first consumer, a 404 reported as a pin error rather
than a network one, and the sdist's own message no longer leading with a task that does not
exist in a tarball.

## Follow-ups this leaves open, deliberately

* Delete `copy_test_utils` and its two tests: `tests/doc_test/utils/` no longer exists, so the
  call is a guarded no-op. Kept here so the fixture cannot raise on a directory nothing
  guarantees.
* A `check_workspace.py` fence asserting `docker/Dockerfile`'s `ARG PLANTUML_VERSION` equals
  `pin.toml`'s `version`, the way it already fences the `__version__` literal — and, in the same
  place, that every step running `fetch_plantuml.py` in a workflow carries `shell: bash` and
  the `tr -d '\r'`.


chrisjsewell:
The red `Image: sphinxneeds` check is pre-existing and expected, not this branch's doing. The docker workflow only runs on pull requests that touch `docker/`, and it has failed the same way since the workspace move (#1839): the image installs sphinx-needs from the `8.5.0` tag with `#subdirectory=packages/sphinx-needs`, and that tag predates the directory. `docker.yaml` says so in its own comment: the path stays red until the first monorepo release, when `NEEDS_VERSION` becomes `sphinx-needs-v<version>`. The step this pull request changed, the pinned PlantUML download, succeeded before the failing one. It is not a required check.

PR #1907, opened 2026-09-07
🧪 Adopt the shared test layer in sphinx-mounts and sphinx-codelinks

chrisjsewell:
Fourth and last slice of the shared-test-layer arc (#1899, #1900, #1902, #1904): the member
`packages/sphinx-needs-testkit` exists and sphinx-needs' suite uses it, and this retires the copies of
the same ideas in the other two suites. **Tests and prose only** — no published package's `src/` is
touched, no `[project]` dependencies change, counts hold: **sphinx-needs 1781/12, sphinx-mounts 943,
sphinx-codelinks 359**.

## sphinx-mounts — the warnings

`count_warnings` counted occurrences of the substring `WARNING:`; `count_mount_warnings` counted
occurrences of `[mounts.` — the warning *family*, not a type. Both are gone; the 82 assertion sites go
through `warning_count`, which counts warning **records** and, given a type, only those carrying exactly
that `[type]` token. The 25 family sites land as **21 typed `== 1` + 2 family + 2 dropped** (a typed
count of zero passes for any type string, and `warning_count(app) == 0` was already beside them). Each
type was read off an assertion the same test already made on the raw stream where one existed — 18 of
the 21 — and otherwise off the source that emits it. An exact-type count says nothing about the *other*
warnings, so at the six sites where the family count was the only "how many" claim the swap would have
been a step down: all six builds are clean apart from the warning they are about, so each gained
`warning_count(app) == 1`. The two survivors say why the total is unavailable to them.

## sphinx-mounts — the renderer

`_plantuml_jar_command` / `_plantuml_extra_conf` were a copy of the resolution chain, and the short one:
`PLANTUML_JAR`, then `plantuml` on `PATH`, with **no step for the jar this repository commits** — the
whole reason the jar had to be handed to this suite in an environment variable. Both are now
`plantuml_conf` rendered as `conf.py` lines, so a checkout needs nothing exported: `poe test-mounts`,
and a bare `pytest`, find `vendor/plantuml/plantuml-<pinned>.jar` themselves. The lazy half earns its
keep on the one parametrised test whose nine cases include a single uml one — only that parameter
resolves a renderer. So `test-mounts` and its three matrix tasks take `deps = ["fetch-plantuml"]` like
every other rendering task; **`test-mounts-bazel` keeps `uses`**, its build running in a Bazel sandbox
that cannot see `vendor/` and reads the variable through `--action_env`. CI is unaffected either way: it
runs `pytest` directly with the variable set by the composite action — route 1 of the same chain.
`test_example.py`'s guard was another copy whose message named the wrong task; it is now one condition
naming `test-mounts-bazel`. Two copies remain on purpose — `docs/conf.py` and `performance_test.py`,
which must not import a test-only member — and the root `AGENTS.md` now says which is which.

## sphinx-codelinks, and one tooling fix

Three of its assertions were `app.warning.getvalue() == ""`, which fails with a bare `assert '' == '<the
whole stream>'`; `assert_no_warnings` says how many were emitted and prints each. A fourth grepped the
raw stream and would have passed on any number of warnings. Nothing else there duplicates the layer: it
renders no diagrams, and its `copytree` calls copy fixture trees into pytest's `tmpdir`. Separately,
`check_workspace.py` check (6) reports a contradiction when a member the root *publishes* carries a
`Private ::` classifier; it read the route as falsy rather than as the empty string, so a member
declared in neither place — already condemned by check (1) — was told "the root depends on it in
`[project]` dependencies". One test; 309.

## Mutations

| mutation | result |
|---|---|
| `PLANTUML_JAR=/nonexistent/plantuml.jar` | red with the **same message from the same line** in sphinx-mounts and sphinx-needs alike, and a third time from `fetch_plantuml.py` through the task |
| the source emits `docname_conflict` under another topic | the exact-type assertion red (`assert 0 == 1`); the family count it replaced, restored on the same tree, green |
| mounts loses the testkit from `pytest_plugins` | 3 failed + 940 passed = the same 943, `fixture 'plantuml_command' not found` — nothing silently collects less |
| `test_example.py` reached with no jar and no `plantuml` | the skip now names `test-mounts-bazel`, the task that sets the variable; with it set, the guard passes and the test proceeds into Bazel |

The plugin **order** is fenced by nothing here, and the conftest says so: reversing it is 943 passed,
because the only fixture both define is `sphinx_test_tempdir`, which this suite never resolves. Gates:
the three suites `-rs`, `pytest tools/tests`, `lint`, `typecheck`, `check-workspace`, `release-plan`.

### DIFF
--- a/.github/workflows/test-extensions.yaml
+++ b/.github/workflows/test-extensions.yaml
@@ -75,8 +75,8 @@ jobs:
           # sphinx-mounts' path directives render with `dot`; sphinx-codelinks renders
           # nothing at all, and pays for the install here because the two share a cell
           needs-graphviz: true
-          # sphinx-mounts' uml tests render for real and `_require_renderer` asserts
-          # rather than skips
+          # sphinx-mounts' uml tests render for real, and the shared resolution RAISES
+          # rather than skipping when it finds no renderer
           needs-plantuml: true
       - name: Check the libclang engine this cell asked for is really there
         # `extra-groups: codelinks-libclang` above is the ONLY executable place that wheel
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -121,10 +121,13 @@ against the pin (one sha256 of 30 MB, well under a second including `uv` and `po
 is what CI's Lint job runs; `uv run poe fetch-plantuml` downloads the jar the pin names, which
 is a **bump** step and otherwise the same hash check. **You will rarely run either by hand**:
 every task that renders declares `fetch-plantuml` through `deps` — the sphinx-needs suites,
-`docs-needs*`, `benchmark-needs` and the sphinx-mounts suites, all of which resolve the jar
-for themselves through the shared test layer; `test-mounts-bazel` is the one that still needs
-`uses = { PLANTUML_JAR = "fetch-plantuml" }`, because its build runs in a Bazel sandbox that
-cannot see `vendor/` — and `lint` declares `verify-plantuml`. `smoke-needs`, `test-needs-js` and the
+`docs-needs*`, `benchmark-needs` and the sphinx-mounts suites. The two SUITES resolve the jar
+for themselves through the shared test layer; `docs-needs*` and `benchmark-needs` keep their
+own hand-written copies of the same three-route chain in `packages/sphinx-needs/docs/conf.py`
+and `packages/sphinx-needs/performance/performance_test.py`, and have to: a docs build and a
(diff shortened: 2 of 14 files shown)


chrisjsewell:
The one red cell (`Needs py3.12 sphinx-7 ubuntu`) is `test_parallel_execution.py::test_doc_build_html` losing its merge-order race again — not this diff. #1908 fixes the test properly (the HTML assertions now follow whichever story kept the id); merge that first, then this branch gets updated and re-run.

PR #1912, opened 2026-09-07
🧪 fix: make sphinx-codelinks fixtures xdist-safe

### LINKED ISSUE #1909: The sphinx-codelinks suite is not xdist-safe: temporary_gitignore races on a checked-in fixture file
`uv run poe test-codelinks -n 4` gives `359 passed, 3 errors`, exit 1, identically on `master` (`292a6c25`) and on #1907's branch. The errors are `FileNotFoundError: …/tests/data/dcdc/.gitignore` at the teardown of the session-scoped autouse fixture `temporary_gitignore`: three xdist workers each write and then remove the same file inside a **checked-in** fixture tree, and the second and third to reach teardown find it gone.

Serial runs are unaffected, which is why CI (which runs this suite without `-n`) and `poe test-codelinks` are green; the sphinx-needs suite runs on four workers since #1899, so this is the one suite that cannot follow.

The fix is either a per-worker copy of `tests/data/dcdc` (or of the whole `tests/data`) into `tmp_path_factory.getbasetemp()`, so nothing under the checkout is written during a run, or a fixture that only the first worker writes and that tolerates a missing file at teardown. The first is the honest one: a test run should never write into checked-in fixtures at all.

Found during the review of #1907 (probe G7 in the review record).

mikemikimike:
## Summary

Fix xdist races in the sphinx-codelinks test fixtures.

The `temporary_gitignore` fixture previously wrote and removed `tests/data/dcdc/.gitignore`, a checked-in fixture path shared by xdist workers. Parallel workers could remove each other's file and fail during teardown.

The fixture now copies `tests/data/dcdc` into each worker's temporary directory, initializes a temporary Git root so `.gitignore` discovery remains covered, and updates the discovery tests to use the isolated fixture.

## Testing

- `uv run poe test-codelinks -n 4` — 360 passed
- `uv run poe lint`
- `uv run poe typecheck`
- `uv run poe docs-codelinks`

closes #1909

### DIFF
--- a/packages/sphinx-codelinks/tests/conftest.py
+++ b/packages/sphinx-codelinks/tests/conftest.py
@@ -1,4 +1,6 @@
 import json
+import shutil
+import subprocess
 from pathlib import Path
 
 import pytest
@@ -45,9 +47,12 @@ ONELINE_COMMENT_STYLE_DEFAULT = OneLineCommentStyle()
 
 
 @pytest.fixture(scope="session")
-def source_directory() -> Path:
+def source_directory(tmp_path_factory: pytest.TempPathFactory) -> Path:
     tests_dir = Path(__file__).parent
-    source_directory = tests_dir / "data" / "dcdc"
+    source_fixture = tests_dir / "data" / "dcdc"
+    source_directory = tmp_path_factory.getbasetemp() / "dcdc"
+    shutil.copytree(source_fixture, source_directory)
+    subprocess.run(["git", "init", "--quiet"], cwd=source_directory, check=True)
     return source_directory
 
 
--- a/packages/sphinx-codelinks/tests/test_cmd.py
+++ b/packages/sphinx-codelinks/tests/test_cmd.py
@@ -142,19 +142,15 @@ def test_analyse_logs_per_project_summary_and_gates_detail(tmp_path: Path) -> No
 
 
 @pytest.mark.parametrize(
-    ("options", "stdout"),
+    ("gitignore", "stdout"),
     [
-        (
-            ["discover", str(TEST_DIR / "data" / "dcdc"), "--no-gitignore"],
-            "4 files discovered",
-        ),
-        (
-            ["discover", str(TEST_DIR / "data" / "dcdc"), "--gitignore"],
-            "3 files discovered",
-        ),
+        (False, "4 files discovered"),
+        (True, "3 files discovered"),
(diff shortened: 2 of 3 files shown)


chrisjsewell (approved):
Thanks — reproduced rather than read, on a worktree at `f037a5c3`:

- **The issue is real and this closes it.** `poe test-codelinks -n 4` on `master`: 359 passed, 3 errors (the `temporary_gitignore` teardown race). On this head: 360 passed in parallel *and* serially, and `git status` is clean after both runs — nothing under `tests/data/` is written any more, which is the property that matters.
- **The `git init` is necessary, not decoration.** The walker comes from the `ignore` crate, whose git-related rules only apply *inside* a repository by default. Measured with the binding directly: a `.gitignore` next to two files is ignored without a `.git` (both files walked) and honoured with one. Without that line the `--gitignore` case would report 4 files. Worth a one-line comment in the fixture saying so, because it reads like scaffolding.
- **Nothing else depended on the checked-in copy having a `.gitignore`.** The Sphinx-route fixture (`tests/data/sphinx/src_trace.toml`, `gitignore = true`, `src_dir = "../dcdc"`) still points at the checkout, so I built it with and without a `demo_1.cpp` entry in `tests/data/dcdc/.gitignore`: identical output both times, 12 needs including the two from `demo_1.cpp` — and the `master` snapshot already contains them. So this PR changes nothing for those tests. (It does mean a `.gitignore` in `dcdc/` was never honoured when the directive walks `./charge`; the CLI on `dcdc/` itself honours it. Pre-existing and separate from this fix — I'll raise it on its own.)

Small things, none blocking:

1. `temporary_gitignore` can now fold into `source_directory`: write the `.gitignore` right after the copy and drop the autouse fixture with its `unlink` teardown — the directory is temporary, so there is nothing to clean up, and one fixture is easier to read than two that depend on each other.
2. `tmp_path_factory.mktemp("dcdc")` is the idiomatic spelling and cannot collide with anything else that might use `getbasetemp()`.
3. The fence test would say more as `assert source_directory.is_relative_to(tmp_path_factory.getbasetemp())` than as `!= checkout_fixture`, which any other path satisfies.
4. Repository conventions: titles carry an emoji (this is 🧪), and a `Closes #1909` line in the body lets the merge close the issue.

Happy for this to merge as is, or with the four folded in.

PR #1914, opened 2026-09-07
👌 register common CSS files deterministically

### LINKED ISSUE #1913: Make common CSS file registration deterministic
## Summary

The order of the common Sphinx-Needs CSS files in generated HTML can vary between otherwise identical builds.

This makes the HTML output non-reproducible and can also change the CSS cascade when the files contain selectors with overlapping specificity.

## Abstract example

Assume the extension copies several CSS files into one directory and registers them during the build:

```text
common_css/
  first.css
  second.css
  third.css
```

Depending on the filesystem or packaging/runfiles implementation, the same build can produce either order in the generated HTML:

```html
<link rel="stylesheet" href="_static/sphinx-needs/common_css/first.css">
<link rel="stylesheet" href="_static/sphinx-needs/common_css/second.css">
<link rel="stylesheet" href="_static/sphinx-needs/common_css/third.css">
```

or:

```html
<link rel="stylesheet" href="_static/sphinx-needs/common_css/second.css">
<link rel="stylesheet" href="_static/sphinx-needs/common_css/third.css">
<link rel="stylesheet" href="_static/sphinx-needs/common_css/first.css">
```

The input files and dependency versions are unchanged; only their directory iteration order differs. This is especially visible in byte-for-byte HTML/golden-output tests and when building locally versus in CI.

## Cause

In `install_styles_static_files`, the common CSS files are discovered and registered with an unsorted `Path.glob()` result:

- [sphinx_needs/environment.py#L54-L61](https://github.com/useblocks/sphinx-needs/blob/8.3.1/sphinx_needs/environment.py#L54-L61)

```python
for common_path in dest_dir.joinpath("common_css").glob("*.css"):
    _add_css_file(app, common_path.relative_to(statics_dir))
```

Sphinx documents that stylesheets with the same priority are emitted in registration order. Therefore, Sphinx preserves the nondeterministic order supplied by Sphinx-Needs rather than introducing a different order:

- [Sphinx `add_css_file` API](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_css_file)

## Suggested fix

At minimum, register the paths in a deterministic order, for example by sorting the glob result by filename:

```python
for common_path in sorted(
    dest_dir.joinpath("common_css").glob("*.css"),
    key=lambda path: path.name,
):
    _add_css_file(app, common_path.relative_to(statics_dir))
```

If the CSS cascade depends on a deliberate semantic order, an explicit ordered list would be safer than alphabetical sorting.

## Environment

Observed with:

- Sphinx-Needs 8.3.1
- Sphinx 9.1.0
- Python 3.12
- a local Bazel build compared with GitHub Actions

AlexanderLanin:
## Why

Common CSS files are discovered with `Path.glob()`, whose iteration order is not guaranteed. Because Sphinx emits same-priority stylesheets in registration order, this can make generated HTML and the CSS cascade vary between otherwise identical builds.

## What changed

- Sort common CSS paths by filename before registering them.
- Add regression coverage asserting that registered common CSS files are ordered deterministically.

Fixes #1913

### DIFF
--- a/packages/sphinx-needs/src/sphinx_needs/environment.py
+++ b/packages/sphinx-needs/src/sphinx_needs/environment.py
@@ -57,7 +57,12 @@ def install_styles_static_files(app: Sphinx, env: BuildEnvironment) -> None:
         str(dest_dir.joinpath("common_css")),
         lambda path: not path.endswith(".css"),
     )
-    for common_path in dest_dir.joinpath("common_css").glob("*.css"):
+    # Sphinx preserves registration order for stylesheets with the same priority.
+    # Sort by filename to keep generated HTML and the CSS cascade deterministic.
+    common_css_files = sorted(
+        dest_dir.joinpath("common_css").glob("*.css"), key=lambda path: path.name
+    )
+    for common_path in common_css_files:
         _add_css_file(app, common_path.relative_to(statics_dir))
 
     # Add theme css file
--- a/packages/sphinx-needs/tests/test_styles/test_style_css_js_registration.py
+++ b/packages/sphinx-needs/tests/test_styles/test_style_css_js_registration.py
@@ -45,6 +45,14 @@ def test_file_registration(tmp_path: Path, make_app: type[SphinxTestApp]):
     css_run_1 = len(css_files)
     script_run_1 = len(script_files)
 
+    common_css_files = [
+        x.filename.rsplit("/", maxsplit=1)[-1]
+        for x in css_files
+        if "/common_css/" in x.filename
+    ]
+    # Sphinx emits stylesheets with the same priority in registration order.
+    assert common_css_files == sorted(common_css_files)
+
     # Check for duplicates
     assert css_run_1 == len(set(css_files))
     assert script_run_1 == len(set(script_files))


chrisjsewell (approved):
Thanks, this is a real one and the fix is the right shape.

**Reproduced.** With the `environment.py` change reverted, a build on macOS (APFS, which iterates directories in hash order like ext4) registers the common stylesheets as `need_style, need_core, needstable, need_links, need_toggle`, and the new assertion in `test_file_registration` fails on it. So the test is a genuine fence, not just a restatement of the fix.

**Sorting is safe for the cascade.** The five files under `css/common/` hold 110 selectors between them and no selector appears in more than one file, so their relative order has no effect on what a page looks like. Alphabetical is also the order NTFS builds were already producing, so Windows users see no change at all.

**Nothing else has the pattern.** The other two `glob` calls in the package (the theme lookup in the same function, and `clear_debug_dir`) are a membership test and a delete loop, so registration order is only decided here.

For the next changelog: the common stylesheets are now registered in filename order, so identical builds produce identical HTML heads.

chrisjsewell (approved):

PR #1917, opened 2026-09-08
🧪 Every test in the sphinx-needs suite builds in process

chrisjsewell:
> *"Ugly workaround to get the sphinx build output. I have no glue how to get it from an
> `app.build()`, because stdout redirecting does not work."*

That comment has sat above `test_doc_github_44` since 2018, and 27 further subprocess builds
grew from it. All it could not reach, the in-process application exposes; all 28 sites are
converted, in eight modules. **No test in sphinx-needs' suite spawns a Sphinx build any more** —
two tree tests in `test_testkit_subprocess.py` keep it so, one against the bare word and one
against `sphinx_build_command(`. sphinx-mounts' bazel tests are untouched.

| the subprocess proved | the in-process assertion |
|---|---|
| warnings on stderr | `build_warnings(app)` compared to the exact list |
| `not out.stderr` | `assert_no_warnings(app)`, which names what it found |
| exit 1 + a class name in a traceback | `pytest.raises(<the class>, match=re.escape(<the full message>))` |
| exit 0 | the build not raising, plus the exact (usually empty) warning list |
| `-W` / `--keep-going` exit codes | dropped: Sphinx's contract, changed in 8.1. What they stood for is the warning list, now asserted directly |
| status text on stdout | `app._status.getvalue()` |
| a second builder over the same source | `make_app(buildername=…, srcdir=app.srcdir, builddir=…)` |
| a rendered diagram | `"plantuml": True` in the parameter dict; `-a -E` becomes nothing at all |

No site loses what it asserted about sphinx-needs; one stdout check narrowed to Sphinx's status
stream, the only channel a Sphinx warning could leak into in process; a stray `print` in
`src/` is no longer seen, and no test in this suite claims it is. Where the in-process form
asserts more, the ledger says so — a class asserted as a type rather than matched in a
traceback; an exact three-record list where a `"WARNING: "` count said two (it read the
*fourth* build into one output directory, where Sphinx had nothing left to re-pickle).

| | before | after |
|---|---|---|
| the eight modules, best of 2 | 48 passed in **43.20 s** | 48 passed in **21.94 s** |
| `git grep -c subprocess tests/` | 51 hits in 11 files | 9 in 4, none a Sphinx build |
| `needuml.py` lines covered by `test_needarch.py` (a subprocess contributes none) | 223 / 317 | 225 / 317 |

**Isolation.** A subprocess was hermetic by accident, so `sphinx_needs`' module-level state was censused: 35
containers, 12 caches, 0 `global`s. `config._NEEDS_CONFIG`, the one large mutable singleton,
is cleared at the end of every `setup(app)`, and nothing else a converted test writes outlives
its application — the eight sites that abort a build half-way included, since `pytest.raises`
keeps the exception in the test body and `cleanup()` still runs.

| proof | result |
|---|---|
| full suite, default / reversed / random seed 1911 | 1787 passed in each at `141ac75f`; 1788 at the head, which adds one test |
| each exception site immediately before `test_basic_doc.py` | 8 × 9 passed |
| all eight as one battery, then three clean modules | 31 passed |

The random order did find one order-dependence, and it is **not** this PR's — reproduced
identically on `master`. `test_measure_time` sets `debug.EXECUTE_TIME_MEASUREMENTS` and never
resets it, so a later build measures a filter function whose test directory has been deleted
and dies in `inspect.getsourcelines`. A commit here closes it test-side; the production half
(`debug.process_timing` should reset it, which matters for esbonio) is a follow-up.

**Also retired:** `plantuml_subprocess_args` (the testkit fixture that pointed a spawned build
at the pinned renderer; no consumer left), the two imports in the eight modules, and the stale
prose. Tests only; nothing shipped changes. No changelog entry.

### DIFF
--- a/packages/sphinx-needs-testkit/src/sphinx_needs_testkit/_plantuml.py
+++ b/packages/sphinx-needs-testkit/src/sphinx_needs_testkit/_plantuml.py
@@ -181,10 +181,9 @@ def make_plantuml_inert(app: SphinxTestApp) -> None:
       starts, and a JVM start is 2.04 s of every 2.13 s render, measured.
     * THIS APP'S OWN ``PlantumlBuilder`` has its two render entry points replaced with one
       that raises. That is the assertion that nothing renders, and it is made on the app
-      rather than on the output directory because 23 test functions in the sphinx-needs
-      suite run a real ``sphinx-build`` SUBPROCESS, four of them into ``app.outdir``
-      itself: a rendered file found in that directory cannot be attributed to the
-      fixture's app, while a call reaching this object can only have come from it.
+      rather than on the output directory because a rendered file found in a directory
+      cannot be attributed to any particular application, while a call reaching this
+      object can only have come from this one.
     * the ``plantuml`` configuration is pointed at :data:`_INERT_PLANTUML_COMMAND`. This one
       is a SENTINEL rather than a fence, and the difference is worth stating: its whole
       protection is that the token cannot be ``exec``-ed, and when sphinxcontrib does try it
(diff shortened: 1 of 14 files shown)


chrisjsewell:
The production half of the order-dependence this found is #1918.

Issue #1425, opened 2025-03-19
needtable layout options


christopheseyler:
In many cases, I had to insert need table with different kind of requirements:
- high level requirement (typically doesn't have parent link and have and 'covered by' link)
- low level requirement (have a parent requirement link and an 'implemented by' link)

I handle that with the columns (and the title feature), i also specify the colwidth and the style (table in my case). And that, each time i need to insert a table.

I would like to specific a table_layout in the conf.py to group all the needed options, something like:

```
needs_table_layouts = {
    'hl_req': {
        'style': 'table', 
        'columns': 'id;title as "Label";content as "Description"; incoming as "Covered by"',
        'colwidths': '10,8,72,10'
     }, 

 'll_req': {
        'style': 'table', 
        'columns': 'id;title as "Label";content as "Description"; outgoing as "Uplink(s)"; incoming as "Implemented by"',
        'colwidths': '10,8,62,10,10'
     }, 

}
```

And, in my documentation, i simply specify:

.. needtable::
   :filter: 'hl_req' in tags and 'my_section' in tags
   :layout: hl_req

It will be, therefore, far easier to update the table content (title, ...)

Issue #1144, opened 2024-03-14
Support sorting needtable with order of apperance in the document


chenlijun99:
Hi! First, thanks for Sphinx-needs. It's very cool!
Maybe I missed something in the documentation, but is it possible to sort the items in the needtable in the order they appear in the source files? 
One challenge I see is how to deal with needs coming from multiple files, but maybe something as simple as sort by filename could work?

chenlijun99:
I see that each need in `needs.json` already has the `docname` and `doctype` properties. Would it be possible to introduce a property like `docsrcline`? And then support a list of keys instead of a single key for the `sort` field of `needtable`.

danwos:
The idea is good, but I'm not sure if sourceline is a good argument.
In a lot of projects, the need objects are stored on different files, get imported by `needs.json` files and are downloaded from services. Therefore not all of them can be assigned to a source line.

May I ask for your use case, in which sorting by sourceline is a benefit?
This helps us to understand all different ways of using Sphinx-Needs much better :)

However, the idea of supporting multiple parameters for `:sort:` is good and makes totally sense.

chenlijun99:
Thanks for the reply!
My use case is that there is a certain logic order in which I write requirements in a document, where e.g. I start from high-level requirements about the whole system followed by more specific requirements about each part of the system.
Finally, for each document I want to include at the top a need table that provides a concise view of all the requirements in the document, with their incoming and outgoing links. I would like the order of the requirements in the table to match the order in which they appear in the document.

PhilipPartsch:
If you have local written need elements, you can use `lineno`.
Here is an example:
```
.. needtable:: List of software requirements
   :types: sw_req
   :style: table
   :columns: id; title; docname as "document"; lineno as "line no"
   :sort: lineno
```

The `lineno` is more correctly set only with sphinx-needs 2.1.0 with support of embedded needs #1150

Issue #771, opened 2022-11-03
Datatable not displaying properly with Read the Docs theme


vasib:
I'm using Read the Docs theme 1.0.0/1.1.0 and this is how a `needstable` with `datatables` style looks like:

![image](https://user-images.githubusercontent.com/90906984/199627043-30b59a7d-77b0-4593-a066-2bd291732dcd.png)

If I remove the `overflow: auto` css property from `.wy-table-responsive` it fixes the issue, but I'm not sure if that is the right way to do this. Since this is very popular theme I was assuming it would work out of the box.
The regular `table` style displays fine.

Is this a *needs* issue or a *rtd* one?

twodrops:
@vasib This is a known issue and there is a PR created at RTD by @danwos before several months for this
https://github.com/readthedocs/sphinx_rtd_theme/pull/1218
The RTD guys are too slow on their PRs.

See issues
https://github.com/readthedocs/sphinx_rtd_theme/issues/1179
https://github.com/readthedocs/sphinx_rtd_theme/issues/767

vasib:
Thanks for the info. That 1 line PR has been sitting there for more than a year 😐.

twodrops:
@vasib 
That's right. We, at our project, are seriously thinking of moving to immaterial theme https://github.com/jbms/sphinx-immaterial

That number doesn't speak well for RTD Theme anymore
![image](https://user-images.githubusercontent.com/103930/200197697-8422428e-22ad-4977-bf8b-28257ccfdb31.png)

chrisjsewell:
#1920 (merged, in the next release) removes half of this collision: DataTables and its `div.dataTables_wrapper` are gone, and the interactive table now sits in a scroll frame of its own. Under sphinx_rtd_theme the theme's own `div.wy-table-responsive` wrapper — the one whose `overflow: auto` this issue is about — ends up nested *inside* that frame, and in our docs build (which renders against sphinx_rtd_theme as one of five themes) a 900 px table scrolls correctly inside it, with the widget's controls staying put.

What we have not verified is a plain project on that theme without our docs' own small override stylesheet, so this is "probably fixed" rather than "fixed". If you still have the project from the screenshot above, could you try it against the next release (or `pip install git+https://github.com/useblocks/sphinx-needs@master#subdirectory=packages/sphinx-needs`) and say whether the table still escapes its container? If it does, the remaining rule belongs in the package rather than in our docs, and that is a small change.

PR #1920, opened 2026-09-08
♻️ Replace DataTables with an in-place, dependency-free needtable enhancer

### LINKED ISSUE #462: Add sphinx-needs specific .css and .js only on html pages where necessary
Currently, the sphinx-needs specific .css and .js files (datatables.min.css/js, common.css, layouts.css , etc.) are added to every .html file globally, disregarding whether needed or not.
For performance reasons (especially the big datatables.min.js), this should be changed to a selective additon only to the necessary htmls.

This can be realized by using the [html-page-context ](https://www.sphinx-doc.org/en/master/extdev/appapi.html#event-html-page-context) event, which is called by the builder after each build of a html page. The css/js can be added via context['css_files'].append("_static/sphinx-needs/....css"),  context['script_files'].append("_static/sphinx-needs/....js") each time.

@danwos I already tested this quickly and think it would work.

This is also loosly related to #455

chrisjsewell:
A `needtable` with the default `:style: datatables` is still interactive — sort a column, search the whole table, page through it, choose how many rows a page holds, switch columns off, copy it or download it as CSV — but it is now Sphinx-Needs' own script instead of a vendored copy of DataTables 1.10.16 from 2018, and it only loads on pages that actually have a table. The enhancement happens **in place**: every `<tr>` and `<td>` Sphinx-Needs rendered is still the same element afterwards, so `:style_row:` classes, the `needs_<column>` cell classes, the links inside cells and the `<colgroup>` all survive sorting and searching, and a reader with JavaScript off still gets the whole table. The script is deferred, so it never blocks the parser.

**4.72 MB of vendored JavaScript in 34 files becomes 58.5 KB in three; a page with a table loads 55.9 KB (14.8 KB gzipped) instead of 2.29 MB (997 KB gzipped), and `search.html`, `genindex.html` and every page without a table now load none of it** (closes #462 for the table assets).

### Why not Grid.js, and why not DataTables 3

Grid.js — what #1464 proposed — reads `td.innerHTML` into a `string[][]`, hides the source table and offers one class string for *all* rows, so `:style_row:` cannot survive it by construction; it has had no release since March 2024. DataTables 3 is jQuery-free, in-place and MIT, and was the honest fallback at about 67 KB gzipped; it was not chosen because it brings its own control DOM and CSS with fourteen `!important` rules that every host theme has to fight, it is a six-week-old major from a one-person project, and any third-party bundle needs a build toolchain and a manifest in this repository that a hand-written file does not. The full comparison — simple-datatables (LGPL-3.0), Tabulator, List.js, tablesort, TanStack — is the decision record at the top of `packages/sphinx-needs/design/needstable-contract.md`.

The claims in this description were re-measured in review on two independent corpora: 41.1 ms and 40.2 ms to initialise ten thousand rows against our 41.5 ms, and 10 000 of 10 000 rows, 60 000 of 60 000 cells and 6 of 6 `<col>` elements still the same DOM objects after five sort, filter and destroy cycles.

### The markup is now a contract

`<table>` keeps `NEEDS_DATATABLES` as its hook and gains `data-needstable-*` options; every `<th>` gains `scope="col"`, a `data-col` column key and, where Sphinx-Needs knows it, a `data-type`; every `<tr>` gains `data-need-id`, and a part row `data-parent`. `:show_filters:` now writes its paragraph *after* the table instead of between `</tbody>` and `</table>`, which was invalid HTML. The design document specifies all of it — the markup, the script's API and behaviours, the CSS custom properties a host answers, and the vendoring rule — so that any producer of needtable markup can emit against it. ubCode, useblocks' Rust-based tooling for needs projects, renders needtables from the same sources and will ship this asset unchanged against the same markup contract.

The script and its structural stylesheet, `src/sphinx_needs/libs/html/needstable.{js,css}`, are plain ES2020 JavaScript type-checked through JSDoc (`// @ts-check`; the one-command `tsc` recipe is in the design document) and host-agnostic CSS that asks for every colour through a `--needstable-*` custom property. Sphinx-Needs' own theme sheets answer those properties; a consumer without them gets `currentColor` fallbacks.

### Two visible behaviour changes

The count under the table reads *Showing 1–10 of 42* and counts **needs**, not rows: a need and its parts are one entry, so a page never splits a need from its parts, and a table with `:show_parts:` reports a smaller total than DataTables did. And a cell with no value now sorts **last** whichever way the column points, where DataTables sorted blanks first.

### Removed, and what replaces them

- **Excel and PDF export** → the CSV download (UTF-8 with a BOM, so Excel opens it directly) and the browser's own print command. They were about 900 KB of the payload for two buttons.
- **Column reordering** → nothing. It was switched on, undocumented, and no issue ever mentioned it.
- **`.dataTables_*` / `.dt-button` CSS hooks** → the `needstable-*` classes. A project that styled the old chrome has to re-point at the new one; a project that styled `.NEEDS_DATATABLES`, `.need`, `.need_part` or `needs_<column>` is unaffected.
- **The `--sn-color-datatable-*` theme variables** → `--sn-color-table-{label,btn-border}`.

Upgrading needs nothing else: this release changes four of the extension's own asset files, and they are now re-copied over a build directory that still holds the old ones. Sphinx would otherwise leave the stale stylesheet in place with a `copy_overwrite` warning — and fail a `-W` build — on the first incremental build after the upgrade.

### Gained

- a page-size control that includes *All* (a review finding on #1464), configurable through `needs_table_page_size`, `needs_table_page_sizes` and a `:page_size:` option;
- typed sorting that returns to the server's `:sort:` order on the third click, so the documented "sorting only works with `:style: table`" caveat is gone;
- part rows that travel with their need through sorting and filtering;
- `aria-sort` and keyboard-operable headers, an `aria-live` result count, focus that stays inside the pager when you page with the keyboard;
- a scroll frame of the widget's own, so a wide table scrolls inside it instead of escaping the page, and a table that fits fills its column;
- the columns control is a plain button in every theme we render, rather than whatever the theme does to a `<summary>`;
- twenty-one browser tests and a markup-contract test module where there were none.

jQuery is still a dependency, because `sphinx_needs_collapse.js` (the need meta box, not the table) is still jQuery — porting it is the follow-up that lets `sphinxcontrib-jquery` go.

### Render pass

The docs were built with furo, sphinx_rtd_theme, pydata-sphinx-theme, sphinx-immaterial and alabaster, in light and dark where the theme has one, before and after, with a new `poe docs-needs-shots` instrument (its own PR follows). The Read the Docs preview of this PR shows the furo build live.

### Changelog draft

```rst
* :ref:`✨ Improvement <needtable>`: The ``datatables`` table style is now Sphinx-Needs' own,
  dependency-free enhancer instead of the bundled `DataTables <https://datatables.net>`__
  package. It sorts (typed, and back to ``:sort:`` order on the third click), searches, pages
  with a 10/25/50/All control, hides columns, copies and downloads CSV -- and it enhances the
  rendered table in place, so ``:style_row:``, the ``needs_<column>`` cell classes and the
  links inside cells survive every interaction. Need parts stay with their need. New:
  :ref:`needs_table_page_size`, :ref:`needs_table_page_sizes` and
  :ref:`page_size <needtable_page_size>`. (:pr:`1920`)
* :ref:`✨ Improvement <needtable>`: The table's script and stylesheet are added only to pages
  that contain a needtable; ``search.html`` and ``genindex.html`` no longer carry them.
  4.72 MB of vendored JavaScript becomes 58.5 KB. (:issue:`462`, :pr:`1920`)
* :ref:`👌 Improvement <needtable>`: The interactive table's count reads *Showing 1-10 of
  42* and counts **needs**: a need and its :ref:`parts <needtable_show_parts>` are one
  entry, so a page never splits a need from its parts. A table with ``:show_parts:``
  therefore reports a smaller total than the DataTables style did, which counted rows.
  Cells with no value now sort last whichever way the column points, where DataTables
  sorted them first. (:pr:`1920`)
* 💥 Removed: Excel and PDF export, and column reordering, from the ``datatables`` style. The
  CSV download opens in Excel and browsers print; column reordering was undocumented.
  Projects that styled ``.dataTables_*`` or ``.dt-button`` must re-point at the
  ``needstable-*`` classes, and ``--sn-color-datatable-*`` is now ``--sn-color-table-*``.
  (:pr:`1920`)
```

Closes #462.

### Related issues, and what this changes for each

- **#1464** — superseded by this PR; the size measurement, the deletion of the dark-theme DataTables block and the 10/25/50/All requirement were kept from it.
- **#408** (per-table DataTables configuration) — the requested shape, a dict of raw DataTables options, no longer applies. Per-table settings now travel as `data-needstable-*` attributes emitted from directive options and config (page size and page sizes ship here); more knobs follow the same route. Note that the side extension posted there in August reconfigures DataTables and stops working with this change, and PDF export is replaced by CSV plus the browser's print dialog.
- **#1425** (named table layouts) — orthogonal server-side presets; nothing in the contract stands in their way. Follow-up.
- **#771** (Read the Docs theme) — the collision was between DataTables' wrapper and the theme's own table wrapper; the DataTables wrapper is gone and the widget's scroll frame is the container, with the theme's wrapper nesting inside it. Measured correct on the docs build; a plain project on that theme still wants confirmation, because these docs carry a small RTD override of their own.
- **#1144** (order of appearance) — `:sort: lineno` already answers it, but only worked with `:style: table` because DataTables re-sorted at load. The interactive table now keeps the server's `:sort:` order as its initial and "unsorted" state, so it works in the default style too. Multiple `:sort:` keys remain a follow-up.

### In short: issues and follow-ups

| | after this PR |
|---|---|
| #462 | closed for the table assets; the shared host stylesheet is still global — follow-up |
| #1464 | superseded; close with the measurement that Grid.js re-renders from cell text |
| #1144 | `:sort: lineno` now works in the default style; close after release, multi-key `:sort:` stays open |
| #771 | DataTables' wrapper gone, frame + theme wrapper nest cleanly on the docs build; confirm on a plain RTD project, then close |
| #408 | answered by `data-needstable-*` attributes rather than a DataTables option dict; the August side extension there stops working — needs a heads-up |
| #1425 | orthogonal server-side presets; unaffected, follow-up |

Follow-up work, none blocking: port `sphinx_needs_collapse.js` off jQuery and drop `sphinxcontrib-jquery`; gate the shared host stylesheet per page; type-check the shared asset in CI; a sticky header now that the frame is an explicit box; a caption-derived CSV file name; logical instead of physical padding for the sort glyph under RTL; and, on request, column reorder, multi-column sort, responsive collapse, URL state or translation catalogues inside the script.


ubmarco (commented):
Did you test the new solution with sphinx simplepdf?

chrisjsewell:
> Did you test the new solution with sphinx simplepdf?

Not before you asked — now yes, and it needs no change on either side.

**What I measured.** sphinx-simplepdf's own `demo/` (which has a `:style: datatables` needtable with `:colwidths:` next to a plain one), built with `-b simplepdf` twice into scratch environments: sphinx-needs 8.5.0 from PyPI, and this branch. Same sphinx 9.1.0, same WeasyPrint 67.0, same simplepdf checkout at `847e1f9`.

- Both builds exit 0 with the same four warnings, all the demo's own (a citation, a footnote, an import key).
- Both PDFs are 80 pages. The extracted text is identical apart from the demo's debug page, which prints the build path and dependency versions.
- The two pages that hold the needtables (53 and 54) rendered at 110 dpi are **pixel-identical: 0 differing pixels of 1,171,170 on each page**.

**Why nothing changes.** WeasyPrint runs no JavaScript, so DataTables never ran in a simplepdf build and the new script does not either — the PDF gets the server-rendered table in both cases. The things that reach WeasyPrint are:

- the `<table class="NEEDS_DATATABLES …">` markup — the hook class is unchanged, so simplepdf's `_needs.scss` rule (`table.NEEDS_DATATABLES { table-layout: fixed; … }`) keeps applying, which is why `:colwidths:` still lays out the second demo table;
- the new `needstable.css` in place of `datatables.min.css` — every selector in it is scoped to elements the script creates (`div.needstable`, `div.needstable-scroll`, …), so without the script it is inert, exactly as the DataTables sheet was (its `table.dataTable` rules needed a class the script added);
- a `<script defer>` tag WeasyPrint ignores, and the same `sphinx_needs_collapse.js` as before.

The only markup differences are additive attributes (`scope`, `data-col`, `data-need-id`, `data-needstable-*`), and `:show_filters:` now writing its paragraph after the table rather than inside it — the demo does not use that option, and in a PDF it renders as a paragraph after the table either way.

**If simplepdf ever wants to adapt** — it does not have to — the one honest improvement would be to key its needtable rule on `table.NEEDS_TABLE, table.NEEDS_DATATABLES` rather than the DataTables name alone, since both styles now share the same static shape; and a project that used simplepdf's `ssp-landscape` container for wide tables keeps working unchanged. The two things a *theme* might have styled that no longer exist are the `.dataTables_*` / `.dt-button` chrome classes and the `--sn-color-datatable-*` variables, both listed under "Removed" above; simplepdf styles neither.

ubmarco (approved):

PR #1944, opened 2026-09-16
🔧 Dependabot: back to the default strategy, with the review rule written down

chrisjsewell:
Reverts #1942's `versioning-strategy: increase-if-necessary` and records, in two places, what was measured and what a reviewer does instead.

### What went wrong with #1942

After the merge, every dependabot job for this repository errored with "Dependabot can't resolve your Python dependency files":

```
No solution found when resolving dependencies for split (… included: sphinx-needs-workspace[group:typing] …):
  Because sphinx>=7.4.0,<=7.4.7 depends on docutils>=0.20,<0.22 and
  sphinx-needs-workspace:typing depends on docutils==0.23, …
```

The mechanism, from dependabot-core's python update checker (which uv reuses): every strategy except `lockfile-only` *unlocks* a requirement before asking uv for the newest resolvable version. For the `typing` group's `docutils~=0.20.0` that asks uv to resolve docutils 0.23 beside `sphinx~=7.4`, which forbids anything above 0.21. Under the default strategy that probe degrades to "no update for docutils" (which is why the 2026-09 batch, #1940, came through with three updates and no docutils); under `increase-if-necessary` it is fatal to the whole job. Both refreshes of #1940 after the merge failed that way, and the run triggered by the config change proposed nothing. Net effect: no dependabot updates of any kind.

`lockfile-only` was the other candidate. It never unlocks, so it cannot hit this, and it never edits a manifest, but it also never proposes a release outside a range. 44 of the 97 external specifiers in these manifests carry a cap or a series, and the widening pull request is the one signal that a new major of one of them exists. Not worth the silence.

### What this does instead

- `.github/dependabot.yml`: the option is removed, and the comment beside the uv block records both measurements and why the default stays.
- `AGENTS.md`: one paragraph under *Pull request requirements* telling a reviewer of a dependabot uv pull request to read the `pyproject.toml` hunks, not only the lock; what the default strategy does to floors, caps and `~=` series, with #1940's three cases as the examples; and the procedure: keep the lock update, restore the specifier lines by hand on the dependabot branch, relock, and give a cap or series that really should move its own pull request (#1943 is the shape).

### After the merge

Dependabot's run triggered by this config change should complete green on the Insights page, which is the proof the jobs are back. #1940 then gets its specifier lines restored by hand and is merged as the lock update it should have been; #1943 drops the click cap on its own reasoning.

### Checked

`uv run poe lint` 8/8 (yamlfmt, taplo, check-readme included). No lock change.


chrisjsewell:
Correction, retracting one claim in this pull request's description and in the comment it added to `.github/dependabot.yml`: the dependabot job that ran after this merged, under the default strategy, errored exactly the way the `increase-if-necessary` runs did. The strategy was never the cause. The failing probe is docutils, made a direct dependency by sphinx-codelinks' bare `docutils` line in #1871, which dependabot tries to pin to 0.23 on every requirement line including the `typing` group's series pin. #1946 fences it with an ignore rule and corrects the wording here and in AGENTS.md.

PR #1948, opened 2026-09-16
🔧 Dependabot: ignore the six releases the workspace cannot lock, found by probing rather than waiting

chrisjsewell:
Six more `ignore` entries in `.github/dependabot.yml`, one per release the workspace cannot lock today, found by reproducing dependabot's failure locally instead of waiting for its next run.

### The probe

For every direct external dependency, `uv lock --dry-run --upgrade-package <name>==<newest on PyPI>` is exactly dependabot's universal pin, and it reproduces "Dependabot can't resolve your Python dependency files" byte for byte, in about a second per package. A negative control (`prek==0.5.3`) resolves cleanly.

Swept over all 67 direct dependencies from a clean worktree at master:

| dependency | newest | why it cannot be locked |
|---|---|---|
| myst-parser | 5.1.0 | needs sphinx>=8; the root `test` group requires it and the sphinx-7 cell installs that group. **This is the failure dependabot is reporting right now.** |
| sphinxcontrib-typer | 0.10.0 | needs sphinx>=8 (sphinx-codelinks' `docs` extra) |
| comment-parser | 1.2.5 | needs Python>=3.13 (sphinx-codelinks' runtime dependency) |
| jsonschema-rs | 0.56.0 | outside sphinx-needs' deliberate `<0.54` cap |
| pydata-sphinx-theme | 0.21.0 | outside sphinx-needs' deliberate `<0.20` cap |
| tree-sitter | 0.26.0 | outside sphinx-codelinks' `~=0.25.1` series |

plus docutils, sphinx and types-docutils, already ignored. The other 58 lock fine.

The last three are out of range, but dependabot's default strategy attempts widenings too (#1940 widened the click cap that way), and a group *refresh* dies on the first unlockable candidate it meets; creation jobs skip it, which is why batches keep arriving and then cannot be rebased.

### Why not a structural fix

Measured: declaring `docs-codelinks` as conflicting with the `sphinx-7` and `typing` groups in `[tool.uv] conflicts` relocks fine, but the probe still fails for both docs tools, because `sphinx-codelinks[docs]` is still reached through the member's own extra, and myst-parser is required by the root `test` group in the sphinx-7 cell regardless. uv can hold a package at two versions across splits; dependabot cannot express that, whatever the manifests say.

### The entries

Each `versions` range is the exact set of releases that cannot be locked today, so patch releases inside the current series still flow. Every entry dissolves when sphinx 7.4 leaves the matrix, and comment-parser's when Python 3.12 does. The comment says to re-run the sweep before adding or removing one.

### Follow-up

The probe becomes a `tools/` script with a poe task and a monthly job that runs it the day before dependabot's, so the list is maintained by measurement. That is a separate pull request; the script as run today is kept for it.

### Checked

`uv run poe lint` 8/8 (yamlfmt included). No lock change.


chrisjsewell:
Closing unmerged: Chris ruled that working around dependabot's one-version-per-package limit is the wrong direction. The uv ecosystem leaves dependabot instead: a scheduled `uv lock --upgrade` job, modelled on `prek-update.yaml`, is being built now. It resolves every conflict split correctly (measured: `docutils v0.20.1, v0.21.2 -> v0.20.1, v0.22.4`), never edits a manifest, and carries a held-back list in its pull request body for the releases the manifests' caps exclude. This branch stays until that lands, in case it is needed as the fallback.

